PR #26, opened 2026-04-04
feat(runtime): provider-modular runtime architecture with profiles and codex support

ichinya:
## What’s Included
- Extracted runtime/provider abstraction into a new `@aif/runtime` workspace (contracts, registry, resolution, capabilities, model discovery).
- Migrated `@aif/agent` to a runtime-driven execution path with workflow policies and proper session isolation.
- Added runtime profiles end-to-end: neutral schema, data-layer CRUD, API routes, UI integration, and MCP sync tools.
- Added Codex runtime adapter and runtime module extensibility wiring.
- Addressed review follow-up issues:
  - Persist session id only for resume-capable workflows.
  - Restored proxy-aware model selection behavior for review-gate.
  - Added guardrails for secret-like headers in runtime profiles.
  - Made checklist auto-sync runtime-neutral (removed hard Claude coupling).
  - Ensured API runtime registry loads external runtime modules (`AIF_RUNTIME_MODULES`).
  - Fixed temporary API-key validation to respect runtime-specific inferred env vars.

## Quality
- Added/updated tests, including Codex transport-path coverage in `@aif/runtime`.
- Validation executed:
  - `npm run build`
  - `npm run lint`
  - `npm test`
  - `npm run coverage`

## Outcome
- Runtime/provider modularity is now implemented end-to-end.
- Review regressions were fixed and coverage thresholds were brought back into compliance.


ichinya:
- Runtime profile handling is now stricter and safer:
  - `apiKeyEnvVar` is validated.
  - Dot/underscore/hyphen names are supported.
  - Invalid env-var names now fall back to inferred defaults instead of silently breaking.
- Runtime resolution for Claude/OpenAI got smarter:
  - Added support for `ANTHROPIC_AUTH_TOKEN`.
  - Added default model envs (`ANTHROPIC_MODEL`, `OPENAI_MODEL`).
  - Better fallback behavior when profile model/env values are missing.
- Claude adapter error/retry logic was hardened:
  - Better handling for missing/expired resume sessions.
  - Retry without resume when resume session is invalid.
  - Improved error classification with clearer details.
- Web/API integration issues were fixed:
  - Added Vite proxy for `/runtime-profiles` (to avoid frontend 404s).
  - WebSocket URL now resolves API port correctly in dev.
  - Improved API error parsing so UI shows meaningful backend errors.
- UI/UX improvements:
  - Runtime Profiles control moved into the header near metrics/settings.
  - Header now shows the currently active runtime profile.
  - Chat panel now shows active profile, runtime, and model.
  - Validation feedback now includes expected env var when missing.
- Windows compatibility improved:
  - Claude CLI lookup now uses `where` on Windows instead of `which`.
- Docs and tests were expanded for all of the above runtime/profile/proxy behaviors.
- Also added TypeScript typing for `VITE_API_PORT` in Vite env.

Known status: **Codex is still not working yet.**

ichinya:
What I changed:
- Pinned `@anthropic-ai/claude-agent-sdk` to `0.2.90` in:
  - [packages/agent/package.json](/D:/Projects/aifhub/aif-handoff/packages/agent/package.json)
  - [packages/api/package.json](/D:/Projects/aifhub/aif-handoff/packages/api/package.json)
  - [packages/runtime/package.json](/D:/Projects/aifhub/aif-handoff/packages/runtime/package.json)

Why this fix:
- `0.2.91+` currently pulls `@anthropic-ai/sdk ^0.80.0` (your vulnerable range in audit).
- `0.2.90` depends on `@anthropic-ai/sdk ^0.74.0`, which removes the reported advisory from `--omit=dev`.

Verification:
- `npm ls @anthropic-ai/claude-agent-sdk @anthropic-ai/sdk` now shows `claude-agent-sdk@0.2.90` and `sdk@0.74.0`.
- `npm audit --omit=dev` now returns: `found 0 vulnerabilities`.
- Build/test smoke checks passed:
  - `npx turbo build --filter=@aif/runtime --filter=@aif/api --filter=@aif/agent`
  - `npm run test --workspace @aif/runtime`

lee-to (commented):
## Code Review: Runtime Provider Modularity

**115 files, +12,238 / −1,725** — Major refactor: new `@aif/runtime` package, runtime profiles, Codex adapter, agent/API migrated to runtime-neutral execution path.

---

### CRITICAL

**1. API keys leak through unredacted `options` objects**

In `packages/runtime/src/adapters/claude/run.ts` and `packages/api/src/services/runtime.ts`, the raw API key is passed into the `options` bag:
```ts
options: {
  ...context.resolvedProfile.options,
  ...(context.resolvedProfile.apiKey ? { apiKey: context.resolvedProfile.apiKey } : {}),
}
```
This `Record<string, unknown>` flows into adapters, audit logs (`writeQueryAudit`), and can be serialized or logged. `redactResolvedRuntimeProfile` exists but is not used in these paths. The Codex AgentAPI adapter (`codex/agentapi.ts`) also sends `input.options` (which may contain `apiKey`) in the POST request body — leaking it beyond the `Authorization` header.

**Recommendation:** Resolve API keys only inside adapters from env var names. Never pass raw keys through `options`.

**2. Full `process.env` spread into child processes**

`packages/runtime/src/adapters/claude/run.ts:~173` and `packages/runtime/src/adapters/codex/cli.ts:~105` spread the entire `process.env` into child processes, including DB credentials, other API keys, and internal secrets.

**Recommendation:** Use a curated allowlist of env vars for child processes.

**3. Validation endpoints accept API keys without authentication**

`POST /runtime-profiles/validate` and `POST /runtime-profiles/models` accept a raw `apiKey` in the JSON body. There is no auth middleware on any API route. An attacker on the network can use these endpoints as an oracle to verify arbitrary API keys against providers.

**Recommendation:** Add authentication middleware, or at minimum rate limiting, before accepting temporary credentials.

---

### HIGH

**4. `__AIF_CLAUDE_QUERY_MOCK__` global hook ships in production** (`claude/run.ts`)

```ts
function resolveQueryImplementation(): typeof query {
  const runtimeGlobal = globalThis as RuntimeGlobalWithQueryMock;
  if (typeof runtimeGlobal.__AIF_CLAUDE_QUERY_MOCK__ === "function") {
    return runtimeGlobal.__AIF_CLAUDE_QUERY_MOCK__;
  }
  return query;
}
```
No `NODE_ENV` guard. Any code can hijack the entire Claude query pipeline via `globalThis`.

**5. `allowDangerouslySkipPermissions` passes through without validation** (`claude/run.ts`, `parseExecutionOptions`)

`metadata` is `Record<string, unknown>`, so any caller can set `allowDangerouslySkipPermissions: true` and it flows directly into Claude Agent SDK options with no access control gate.

**6. Failed singleton registry promise is cached forever**

Both `packages/agent/src/subagentQuery.ts` and `packages/api/src/services/runtime.ts` cache `runtimeRegistryPromise`. If the first initialization fails (e.g., a module fails to load), the rejected promise is cached permanently. All subsequent calls return the rejected promise, requiring a process restart to recover.

**Recommendation:** Reset `runtimeRegistryPromise = null` in the catch path.

**7. No authentication on runtime profile CRUD endpoints**

`POST/PUT/DELETE /runtime-profiles` have no auth guard. An attacker can redirect all AI queries to a malicious provider by creating a profile with a custom `baseUrl`.

**8. Sensitive header denylist is incomplete** (`runtimeProfiles.ts`, `listSensitiveHeaderKeys`)

Catches "authorization", "cookie", "token", "api-key", "apikey", "secret" — but misses `proxy-authorization`, `set-cookie`, `x-goog-api-key`, provider-specific headers. An allowlist approach would be more secure.

**9. `stdin.write()` without error handler** (`codex/cli.ts`)

`child.stdin.write(input.prompt)` can emit an unhandled `error` event (broken pipe), which crashes the process. Add `child.stdin.on('error', ...)`.

---

### MEDIUM

**10. Async iterator double-consumption** (`claude/run.ts:~300`)

Calls `iterator.next()` for the first message, then `for await (const message of stream)` which creates a new iterator from the same stream. Depending on the SDK's iterator implementation, this may miss messages or cause double-consumption.

**11. Cache has no size limit** (`cache.ts`)

`createRuntimeMemoryCache` uses a `Map` that grows without bound. Expired entries are only evicted on `get()`. Memory leak in long-running processes.

**12. Codex adapter claims `supportsResume: true` but has no session management**

`codex/index.ts` declares `supportsResume: true`, but there are no `listSessions`/`getSession` implementations. The CLI transport ignores session IDs entirely.

**13. `toIso` can throw `RangeError`** (`claude/sessions.ts`)

`new Date(malformedString).toISOString()` throws on invalid input. No try-catch guard.

**14. `listClaudeModels` / `listCodexModels` are hardcoded static lists**

`supportsModelDiscovery: true` is declared, but no actual discovery happens. The model lists (`claude-sonnet-4-5`, `gpt-5.4`, etc.) will become stale.

**15. `runRuntimeBackfills` runs on every startup** (`db.ts`)

UPDATE queries execute on every server start even when no rows need updating. Low impact but unnecessary write I/O.

**16. `resolveEffectiveRuntimeProfile` silently skips disabled profiles**

When a task's `runtimeProfileId` points to a disabled profile, it silently falls through to the project default with no user notification.

**17. Resume-failure retry has no delay** (`claude/run.ts`)

The `isRetryableResumeFailure` branch retries immediately (no `sleep`), while the timeout retry uses `retryDelayMs`. Could cause thundering-herd on transient failures.

---

### LOW

- `ENV_VAR_NAME_REGEX` allows dots/hyphens — invalid for Unix env vars
- `classifyCode` in `claude/errors.ts` — `.includes("stream")` is too broad (matches "downstream", etc.)
- `listCodexAgentApiModels` is exported but never called (dead code)
- `zValidator("json", schema as never)` type casts suppress type checking at validation boundary
- `.ai-factory/config.yaml` — `language.ui` and `language.artifacts` changed from `en` to `ru` — intentional?
- `normalizeRuntimeId` lowercases IDs, which could cause surprising collisions with custom runtime IDs (undocumented)

---

### Architecture — positive notes

- Clean separation: runtime contracts extracted into a standalone package
- DB boundary rule is respected (minor exception in chat routes via internal repositories)
- Profile resolution fallback chain (task → project → system) is well-implemented
- Redact functions for logging exist (though not applied everywhere)
- Codex adapter as proof-of-concept validates the abstraction
- Migrations are idempotent with proper backfill logic

### Recommendation

**Do not merge** until CRITICAL and HIGH issues are resolved. Priority fixes:
1. Remove raw API keys from `options` objects — resolve only inside adapters from env var names
2. Use an allowlist for `process.env` in child processes
3. Add auth middleware (or at least rate limiting) on validation endpoints
4. Add `NODE_ENV === 'test'` guard for the mock hook
5. Reset `runtimeRegistryPromise` on initialization failure
6. Add `child.stdin.on('error', ...)` handler

PR #30, opened 2026-04-06
Feature/runtime provider modularity fix

lee-to:
## Changes

New runtime modulke

## AI Model

- [x] Claude Opus 4.6
- [ ] Claude Sonnet 4.6
- [ ] Claude Haiku 4.5
- [ ] Other: 
- [ ] No AI used

## Test Plan

- [x] Tests pass (`npm test`)
- [x] Documentation updated (if applicable)


ichinya:
You have two separate failures, and they come from different layers.

**1) SDK transport failure (first error)**
- Log: `Unable to locate Codex CLI binaries. Ensure @openai/codex is installed with optional dependencies`
- Source: `@openai/codex-sdk` `findCodexPath` -> `aif-handoff/node_modules/@openai/codex-sdk/dist/index.js:368`
- Meaning: SDK mode does not use a global `codex` command. It expects platform vendor binaries from optional deps (`@openai/codex-win32-x64` on your machine).
- In this workspace, `@openai/codex` and `@openai/codex-sdk` are installed, but win32 optional package is missing (`npm ls @openai/codex-win32-x64` returned empty).

**2) CLI transport failure (second error)**
- Log: `spawn codex ENOENT`
- Source: CLI run path in `runCodexCli` -> `aif-handoff/packages/runtime/src/adapters/codex/cli.ts:220`
- Meaning: process spawn cannot resolve `codex` in that runtime environment.
- Why validation still said OK:
  - CLI validation in `validateCodexCliConnection` -> `aif-handoff/packages/runtime/src/adapters/codex/index.ts:131` only checks “is configured” and path existence for explicit paths.  
  - For plain `"codex"`, it returns OK without executing anything.

**Quick map**
```text
Profile transport=sdk  -> codex-sdk expects vendored binary -> missing optional dep -> fail
Profile transport=cli  -> spawn("codex") in runtime env     -> command not resolved -> ENOENT
```

**Secondary thing you saw (log spam)**
- Repeated `Effective runtime profile fell back...` is from `packages/data/src/index.ts` -> `aif-handoff/packages/data/src/index.ts:1123`, not the crash itself.

ichinya:
**What It Is**

This Claude failure is not auth, model, or `z.ai` base URL. It is a Windows process-launch bug in the Claude adapter.

The adapter auto-discovers `claude.cmd` from `%APPDATA%\npm` in `findPath.ts` -> `/aifhub/aif-handoff/packages/runtime/src/adapters/claude/findPath.ts:6. That path is then passed into SDK mode as `pathToClaudeCodeExecutable` through `index.ts` -> `aif-handoff/packages/runtime/src/adapters/claude/index.ts:166` and `options.ts` -> `aif-handoff/packages/runtime/src/adapters/claude/options.ts:236. Inside Anthropic’s SDK, that value is treated as the command to spawn directly in `sdk.mjs` -> `aifhub/aif-handoff/packages/runtime/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs:58. On Windows, directly spawning a `.cmd` file this way reproduces `EINVAL`, which matches your log exactly.

I verified the machine has `C:\Users\User\AppData\Roaming\npm\claude.cmd`, and a direct Node spawn of that file returns `EINVAL`. So the root cause is:

```text
Windows
  -> adapter finds `claude.cmd`
  -> SDK tries spawn("...\\claude.cmd", ...)
  -> Node/Windows rejects direct .cmd launch
  -> `spawn EINVAL`
```

The invalid `apiKeyEnvVar` warning is separate noise. Resolution correctly falls back to `ANTHROPIC_API_KEY`, and your log shows `hasApiKey: true`.

**What Follows From That**

Two important consequences:

1. SDK transport is broken on Windows when the resolved Claude path is a `.cmd` wrapper.
2. CLI transport would hit the same class of issue, because `cli.ts` -> `aif-handoff/packages/runtime/src/adapters/claude/cli.ts:254` also does `spawn(cliPath, args)` directly.

There is also the same validation blind spot as with Codex: `validateClaudeConnection` -> `aif-handoff/packages/runtime/src/adapters/claude/index.ts:95` never tests whether the executable is actually spawnable.

**Likely Fix Direction**

The clean fixes are:

1. On Windows, do not pass `.cmd` as `pathToClaudeCodeExecutable` into SDK mode.
2. For CLI mode, either:
   - use bare `claude` only when `%APPDATA%\npm` is on `PATH`, or
   - wrap `.cmd` with `cmd.exe /d /s /c`, or
   - prefer a real `claude.exe` if one exists.
3. Strengthen validation so it performs a real spawn probe instead of returning “configured”.

ichinya:
**Problems**

1. Codex CLI validation is currently a false positive. `validateCodexCliConnection` in `packages/runtime/src/adapters/codex/index.ts` treats a bare `codex` command as valid, but the actual execution path in `packages/runtime/src/adapters/codex/cli.ts` can still fail with `spawn codex ENOENT`. That means the profile is reported as healthy even though the transport is not actually runnable.

2. Codex SDK validation does not check the SDK’s real runtime requirement: the platform-specific optional binary package. For `transport=sdk`, having `@openai/codex-sdk` or a `codex` string is not sufficient; the SDK resolves a bundled binary from `@openai/codex-<platform>`. On Windows x64, if `@openai/codex-win32-x64` is missing, validation still passes but execution fails with `Unable to locate Codex CLI binaries...`.

3. On Windows, the Claude adapter passes a discovered `.cmd` wrapper as the executable path directly. That is not safely spawnable via `spawn(...)` and reproduces `EINVAL`. This affects both the SDK path and the CLI path. `validateClaudeConnection` has the same blind spot as the Codex validation: it reports `ok` without verifying that the executable can actually be launched in the current environment.

Overall, runtime validation is currently reporting transports as “configured” even when they are not executable on the current platform. That hides the defect until the first real run and makes the failure mode much harder to diagnose.

PR #47, opened 2026-04-09
fix(web): use same-origin websocket url

Etherlord:
## Summary

- Fix dev WebSocket URL generation when the app is accessed through a public reverse proxy on a different external port
- Keep internal API/WebSocket communication on `PORT` while allowing the client to use `PUBLIC_PORT`

## Changes

Updated the web Vite config to inject a client-facing WebSocket/API port separately from the internal backend port.

Previously, the client always fell back to the internal API port (`PORT`, e.g. `3009`) in development, which caused WebSocket connection attempts to bypass the reverse proxy and connect directly to the internal port.

This change introduces a public port fallback chain for the client:
- use `PUBLIC_PORT` when provided
- otherwise fall back to `PORT`

The injected value is serialized as a string for `import.meta.env.VITE_API_PORT`, matching the client-side WebSocket URL resolution logic.

This allows:
- backend and Vite proxy traffic to continue using the internal port
- browser WebSocket connections to use the externally exposed port (for example `443`) when running behind a reverse proxy

## AI Model

- [ ] Claude Opus 4.6
- [ ] Claude Sonnet 4.6
- [ ] Claude Haiku 4.5
- [x] Other: GPT-5.4 Thinking
- [ ] No AI used

## Test Plan

- [ ] Tests pass (`npm test`)
- [ ] Lint passes (`npm run lint`)
- [ ] Documentation updated (if applicable)

### DIFF
--- a/packages/web/src/hooks/useWebSocket.ts
+++ b/packages/web/src/hooks/useWebSocket.ts
@@ -23,17 +23,8 @@ function hasIdPayload(value: unknown): value is { id: string } {
 
 function resolveWsUrl(): string {
   const protocol = window.location.protocol === "https:" ? "wss:" : "ws:";
-  if (!import.meta.env.DEV) {
-    return `${protocol}//${window.location.host}/ws`;
-  }
 
-  const configuredPort = import.meta.env.VITE_API_PORT;
-  const apiPort =
-    typeof configuredPort === "string" && configuredPort.trim().length > 0
-      ? configuredPort.trim()
-      : "3009";
-
-  return `${protocol}//${window.location.hostname}:${apiPort}/ws`;
+  return `${protocol}//${window.location.host}/ws`;
 }
 
 /** Per-client WS identifier assigned by server on connect */


lee-to:
Thanks for catching this! The issue is real — in dev mode the client bypasses the reverse proxy by connecting directly to the API port.

However, I think there's a simpler fix that doesn't require an extra env var. The root cause is in `resolveWsUrl()` (`packages/web/src/hooks/useWebSocket.ts`) — in DEV mode it builds the URL with a hardcoded API port instead of using `window.location.host`. But the Vite dev server already has a proxy configured for `/ws` (see `vite.config.ts:74-77`), so the client can just go through it.

The fix would be to simplify `resolveWsUrl()` to always use `window.location.host`:

```ts
function resolveWsUrl(): string {
  const protocol = window.location.protocol === "https:" ? "wss:" : "ws:";
  return `${protocol}//${window.location.host}/ws`;
}
```

This works in all scenarios:
- **Local dev** → browser hits Vite on `localhost:5180` → Vite proxy forwards `/ws` to `localhost:3009`
- **Dev behind reverse proxy** → browser hits proxy → proxy forwards to Vite → Vite proxy forwards to API
- **Docker/production** → browser hits Angie → Angie proxies `/ws` to API

No extra config needed, same behavior in dev and production. Would you be up for updating the PR with this approach?

Etherlord:
@lee-to thanks, this works — I’ve pushed the fix.

ichinya:
use in env `WEB_HOST=0.0.0.0` ?
The only thing missing is `allowedHosts` for run web

lee-to:
> use in env `WEB_HOST=0.0.0.0` ?
> The only thing missing is `allowedHosts` for run web

Let's make this a separate pull request

Etherlord:
> use in env `WEB_HOST=0.0.0.0` ? The only thing missing is `allowedHosts` for run web

I used the __VITE_ADDITIONAL_SERVER_ALLOWED_HOSTS environment variable. I can add it to .env.example in a separate PR.

PR #57, opened 2026-04-10
Fix chat timeout handling and agent polling interval

ichinya:
## Changes

This PR fixes two timing issues that were causing confusing behavior in production. Chat requests now honor the timeout values from the environment instead of falling back to the Codex CLI default, and agent polling now respects `POLL_INTERVAL_MS` correctly. Regression tests were added for both cases.

## AI Model

- [ ] Claude Opus 4.6
- [ ] Claude Sonnet 4.6
- [ ] Claude Haiku 4.5
- [x] GPT-5.4
- [ ] No AI used

## Test Plan

- [x] Tests pass (`npm test`)
- [ ] Documentation updated (if applicable)
- [ ] Manually verified the change works as expected

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2994,13 +2994,6 @@
         "undici-types": "~7.18.0"
       }
     },
-    "node_modules/@types/node-cron": {
-      "version": "3.0.11",
-      "resolved": "https://registry.npmjs.org/@types/node-cron/-/node-cron-3.0.11.tgz",
-      "integrity": "sha512-0ikrnug3/IyneSHqCBeslAhlK2aBfYek1fGo4bP4QnZPmiqSGRK+Oy7ZMisLWkesffJvQ1cqAcBnJC+8+nxIAg==",
-      "dev": true,
-      "license": "MIT"
-    },
     "node_modules/@types/react": {
       "version": "19.2.14",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.2.14.tgz",
@@ -8198,15 +8191,6 @@
         "node": ">=10"
       }
     },
-    "node_modules/node-cron": {
-      "version": "4.2.1",
-      "resolved": "https://registry.npmjs.org/node-cron/-/node-cron-4.2.1.tgz",
-      "integrity": "sha512-lgimEHPE/QDgFlywTd8yTR61ptugX3Qer29efeyWw2rv259HtGBNn1vZVmp8lB9uo9wC0t/AT4iGqXxia+CJFg==",
-      "license": "ISC",
-      "engines": {
-        "node": ">=6.0.0"
-      }
-    },
     "node_modules/node-releases": {
       "version": "2.0.36",
       "resolved": "https://registry.npmjs.org/node-releases/-/node-releases-2.0.36.tgz",
@@ -11138,11 +11122,9 @@
       "dependencies": {
         "@aif/data": "*",
         "@aif/runtime": "*",
-        "@aif/shared": "*",
-        "node-cron": "^4.2.1"
+        "@aif/shared": "*"
       },
       "devDependencies": {
-        "@types/node-cron": "^3.0.11",
         "tsx": "^4.21.0",
         "typescript": "^5.9.3",
         "vitest": "^4.1.2"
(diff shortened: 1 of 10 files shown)


ichinya:
Added a follow-up fix for another timing issue in the agent path that was showing up in production.

`implement-checklist-sync` is a tool-less workflow, but the first-activity watchdog previously treated only tool/subagent hooks as proof of life. That meant a run could be streaming runtime events normally and still get restarted, eventually surfacing as a misleading timeout.

What changed:
- first-activity watchdog now treats streamed `RuntimeEvent`s as activity, not just tool/subagent hooks
- added a regression test covering a tool-less workflow that emits early runtime events and completes successfully
- updated `packages/agent/CHECKLIST.md` so this case is explicitly checked when watchdog logic changes

Validation:
- `npm run lint`
- `npm test`
- `npm run build`

lee-to:
## Review

Three related, well-scoped fixes with solid test coverage. CI is green. Good to merge after a couple of nits.

### What the PR does

1. **`packages/api/src/routes/chat.ts`** — passes `API_RUNTIME_START_TIMEOUT_MS` / `API_RUNTIME_RUN_TIMEOUT_MS` from env into `execution`. Before the fix, chat requests silently fell back to Codex CLI defaults instead of the configured values.
2. **`packages/agent/src/pollScheduler.ts`** — replaces `node-cron` with `setInterval`. Fixes a real bug: with `POLL_INTERVAL_MS=600_000` the old code built `*/600 * * * * *`, which is an invalid step in the seconds field (max 59). `node-cron` was never used for actual cron semantics here — just as a wrapper around an interval — so this is a simplification, not a regression. Bonus: the interval is now in milliseconds directly, no precision loss from `Math.floor(ms/1000)`.
3. **`packages/agent/src/subagentQuery.ts`** — the first-activity watchdog now treats `onEvent` as proof of life, not just `onToolUse` / `onSubagentStart`. Fixes tool-less workflows like `implement-checklist-sync` being killed mid-run.

### Strengths

- Tests target the exact bugs: `pollScheduler.test.ts` exercises the 600_000 ms case specifically, and `subagentQuery.test.ts` reproduces a tool-less workflow that only emits a `system/init` event.
- `packages/agent/CHECKLIST.md` is updated with both invariants so the regression can't silently come back.
- `chat.ts` also memoizes `getEnv()` into a local — small but clean.

### Comments

1. **Dead `node-cron` dependency** in `packages/agent/package.json` (both `node-cron` and `@types/node-cron`). After the import is removed from `index.ts`, nothing else under `packages/agent/` uses it — worth dropping in this PR, otherwise the next `npm audit` / bundle review will raise it.

2. **Overlap / reentrancy in `pollScheduler`.** `setInterval(() => { void callback(); })` has no guard — if `pollAndProcess()` hangs longer than `intervalMs`, overlapping cycles will stack. Not a regression (the `node-cron` version had the same issue), but since we're rewriting this anyway, an `isRunning` flag that skips a tick while the previous one is still in flight would be a cheap improvement. Fine as a follow-up.

3. **Watchdog comment is slightly stale** (`subagentQuery.ts`, the block right before `baseFirstActivityTimeoutMs`). The doc correctly mentions `RuntimeEvent` now, but the sentence "model preambles stream as agent_message text rather than tool_use events (so they don't reset the watchdog)" is no longer true — preambles now *do* reset the watchdog via `onEvent`. Either drop that justification or rephrase it (the 2x buffer is still justified by cold-start cost, just not by this reason).

4. **`executionIntent.onEvent` is now always assigned**, even when `originalOnEvent` is `undefined`. Not a bug (the wrapper no-ops), but it's asymmetric with `onToolUse` / `onSubagentStart`, which are only wrapped when the original exists. Purely cosmetic.

### Verdict

Merge once `node-cron` is removed from `package.json` (or a follow-up ticket is filed for it). Everything else is nits.

ichinya:
Follow-up fixes are in.

- removed dead `node-cron` deps from `@aif/agent`
- added non-overlap protection to `pollScheduler`
- added a regression test for overlapping interval ticks
- updated `subagentQuery.ts` watchdog comments/messages to match `RuntimeEvent`-based activity handling
- extended `packages/agent/CHECKLIST.md` with the scheduler overlap check

Validated with:
- `npm run lint`
- `npm test`
- `npm run build`

lee-to:
All three review comments are addressed in 430c22e:

1. **`node-cron` removed** from `packages/agent/package.json` (both dep and devDep). ✅
2. **Reentrancy guard in `pollScheduler`** — `isRunning` flag skips overlapping ticks, with a dedicated test (`skips interval ticks while the previous callback is still running`). ✅
3. **Stale watchdog comment rewritten** — the incorrect preamble justification is gone, the 2x CLI buffer is now explained purely by cold-start cost. JSDoc and the `logActivity` message are updated consistently (`no tool activity` → `no runtime activity`). ✅

Bonus: `originalOnEvent` now defaults to a no-op (`?? (() => undefined)`), which is cleaner than the optional-chained call.

CI is green across all four checks. LGTM — ready to merge. 🚀

PR #58, opened 2026-04-11
Enable HTTP MCP in local dev when MCP_PORT is set

ichinya:
## Summary

- Start `@aif/mcp` automatically during `npm run dev` when `MCP_PORT` is configured
- Install and remove the MCP server from the web settings UI using a localhost HTTP entry instead of `npx` when `MCP_PORT` is set
- Keep the existing `stdio` / `npx` fallback for environments where `MCP_PORT` is not configured

## Changes

Updated the local dev flow so the monorepo starts the MCP server only when `MCP_PORT` is present, including support for loading the value from the root `.env`. Added an HTTP-specific dev runner for the MCP package and documented the new behavior.

Updated MCP installation through the settings API and runtime adapters so Claude and Codex can persist either `stdio` or streamable HTTP MCP entries. When `MCP_PORT` is set, the web UI now installs `http://localhost:<MCP_PORT>/mcp` and removes that same entry during uninstall instead of writing an `npx tsx ...` launcher config.

Added route and adapter tests to cover the HTTP install/remove path and the new config serialization behavior.

## AI Model

- [ ] Claude Opus 4.6
- [ ] Claude Sonnet 4.6
- [ ] Claude Haiku 4.5
- [x] Other: OpenAI Codex (GPT-5)
- [ ] No AI used

## Test Plan

- [x] Tests pass (`npm test`)
- [x] Documentation updated (if applicable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ npm run init
 npm run dev
 ```
 
+Set `MCP_PORT` in your shell or root `.env` before `npm run dev` if you also want the MCP HTTP server in local development.
+
 ### With Docker
 
 ```bash
@@ -53,13 +55,14 @@ cd aif-handoff
 docker compose up --build
 ```
 
-Both options start three services:
+Development starts three services by default. If `MCP_PORT` is set, it starts a fourth service for MCP over HTTP. Docker starts all four services.
 
-| Service   | URL                     | Description                                  |
-| --------- | ----------------------- | -------------------------------------------- |
-| **API**   | `http://localhost:3009` | Hono REST + WebSocket server                 |
-| **Web**   | `http://localhost:5180` | React Kanban UI                              |
-| **Agent** | _(background)_          | Event-driven + polling, dispatches subagents |
+| Service   | URL                               | Description                                  |
+| --------- | --------------------------------- | -------------------------------------------- |
+| **API**   | `http://localhost:3009`           | Hono REST + WebSocket server                 |
+| **Web**   | `http://localhost:5180`           | React Kanban UI                              |
+| **Agent** | _(background)_                    | Event-driven + polling, dispatches subagents |
+| **MCP**   | `http://localhost:<MCP_PORT>/mcp` | Optional local MCP HTTP endpoint             |
 
 The agent coordinator reacts to task events via WebSocket in near real-time and falls back to 30-second polling. Activity logging can be switched to batch mode (`ACTIVITY_LOG_MODE=batch`) to reduce DB write amplification. See [Configuration](docs/configuration.md) for all tuning options.
 
@@ -161,7 +164,7 @@ The project includes full Docker support (Angie reverse proxy + Node services).
 docker compose up --build
 ```
 
-Web UI at `localhost:5180`, API at `localhost:3009`.
+Web UI at `localhost:5180`, API at `localhost:3009`, MCP at `localhost:${MCP_PORT:-3100}`.
 
 ### Production
 
@@ -178,6 +181,7 @@ Only ports 80/443 are exposed. API is bound to localhost only. Includes security
 | `ANTHROPIC_API_KEY` | —            | API key (or use `claude login`)        |
 | `DOMAIN`            | `localhost`  | Domain for SSL certificate (ACME)      |
 | `PORT`              | `3009`       | Host port for API                      |
+| `MCP_PORT`          | `3100`       | Host port for MCP HTTP server          |
 | `WEB_PORT`          | `5180`       | Host port for Web UI (dev)             |
 | `WEB_HOST`          | `localhost`  | Web UI dev server host (Vite)          |
 | `HTTP_PORT`         | `80`         | Host port for Web UI (production)      |
--- a/docs/api.md
+++ b/docs/api.md
@@ -588,6 +588,8 @@ The WebSocket endpoint is a simple broadcast channel — no authentication, no s
 
 The Handoff MCP server (`packages/mcp`) provides bidirectional sync between AIF tooling and Handoff via the Model Context Protocol. When MCP tools modify tasks, they broadcast `sync:*` events over the WebSocket system so the Kanban UI reflects changes in real time.
 
+The web settings route `POST /settings/mcp/install` installs the MCP server into supported runtimes. When `MCP_PORT` is set in the server environment, it writes a streamable HTTP entry pointing to `http://localhost:<MCP_PORT>/mcp`; otherwise it writes the local `stdio` launcher entry.
+
 See [MCP Sync Server](mcp-sync.md) for full documentation.
 
 ## See Also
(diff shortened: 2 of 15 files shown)


lee-to (changes requested):
## Review

Focused change — conditional MCP launch in dev and switching the installer to the HTTP form. Tests are expanded in the right places (claudeMcp, codexMcp, settings route). Comments below.

### Findings

**1. Claude HTTP entry is missing `type: "http"`** — `packages/runtime/src/adapters/claude/mcp.ts` (new URL branch)
The stdio branch writes `type: "stdio"`, while the URL branch writes only `{ url }`. The `.claude.json` schema for HTTP MCP servers expects an explicit `type: "http"`. A bare `url` may work by inference, but it is inconsistent and fragile against future Claude Code versions. Add `type: "http"` and update the corresponding assertion in `claudeMcp.test.ts`.

**2. `bearerTokenEnvVar` in the Claude adapter is a dead / incorrect path**
The Claude branch writes the field in camelCase. The Codex branch uses snake_case `bearer_token_env_var`, which is correct for the Codex TOML schema. In Claude Code's `.claude.json`, HTTP-transport auth is configured via `headers: { Authorization: ... }`, not via a `bearerTokenEnvVar` field — so the field will never be read. Either (a) drop it from the Claude adapter, or (b) implement it properly through `headers`. As-is the API promises a feature that does not exist.

**3. IIFE-throw inside a ternary** — `packages/runtime/src/adapters/codex/mcp.ts`

```ts
input.url ? {...} : input.command ? {...} : (() => { throw new Error(...); })()
```

Replace with a plain `if / else if / else` before the `serializeMcpSection` call — more readable and easier to debug. The same pattern applies to the Claude adapter.

**4. `RuntimeMcpInstallInput`: validate at the type level**
After the change, both `command` and `url` are optional, and validity is only checked at runtime. Consider a discriminated union:

```ts
type RuntimeMcpInstallInput = RuntimeMcpInput &
  ({ transport: "stdio"; command: string; args?: string[]; cwd?: string; env?: Record<string, string> }
 | { transport: "streamable_http"; url: string; bearerTokenEnvVar?: string });
```

This removes both IIFE-throws and guards against incomplete calls.

**5. `scripts/dev.mjs` — `parseEnvFile`**
- Handles only simple quoted values; escape sequences (`\n`, `\"`) and multi-line values are not supported. Fine for dev, but either document the limitation inline or use `dotenv` (likely already present transitively) instead of duplicating the logic.
- `MCP_PORT="0"` passes the truthy check. Minor, but `Number(process.env.MCP_PORT) > 0` would be more honest.

**6. Docker / compose untouched**
The README says "Docker starts all four services" and `MCP_PORT` is already in the env table. Please confirm `docker-compose.yml` actually runs an MCP service — otherwise this contradicts the docs and needs a separate fix under the Docker Sync Rule.

**7. Duplicated spawn/signal logic between `scripts/dev.mjs` and `packages/mcp/scripts/dev-http.mjs`**
Both scripts repeat the same spawn + SIGINT/SIGTERM forwarding + Windows branch. A shared helper would remove the duplication. Nit.

**8. `settings.test.ts`: matrix is incomplete**
The new test only covers the HTTP install branch. Add an explicit assertion that when `MCP_PORT` is not set the Codex TOML contains `command = "npx"` — that locks down both branches at once.

### Positives
- Gating behind `MCP_PORT` (no-op by default) — backward compatible.
- Unified `RuntimeMcpInstallInput` instead of manual call-site spread.
- Round-trip tests that read HTTP entries back from `.claude.json` and TOML — good parity.
- Documentation updated in sync: README, docs/api.md, docs/getting-started.md, docs/mcp-sync.md.

### Verdict
Blockers — items 1 and 2 (Claude `type: "http"` and incorrect `bearerTokenEnvVar`). The rest are improvements and can land separately.

ichinya:
Addressed the review feedback in this update.

Completed:
- Added `type: "http"` to Claude HTTP MCP entries and updated the Claude MCP tests accordingly.
- Removed unsupported `bearerTokenEnvVar` serialization from the Claude adapter. Claude now writes only the fields it actually supports for HTTP MCP.
- Tightened `RuntimeMcpInstallInput` into a discriminated union so `stdio` installs require `command` and HTTP installs require `url`.
- Replaced the inline throw / ternary branching in the Claude and Codex adapters with explicit transport-based handling.
- Hardened `MCP_PORT` parsing so only valid positive integer ports enable HTTP MCP behavior.
- Added the missing stdio assertion in `settings.test.ts` so both install paths are covered:
  - `stdio` when `MCP_PORT` is not set
  - `streamable_http` when `MCP_PORT` is set

Also verified that Docker/compose already includes the MCP service, so no Docker changes were required for this PR.

Validation:
- `npm run lint` passed
- `npm test` passed

Note:
- `npm run build` still fails due to a pre-existing `@aif/agent` TypeScript issue with `node-cron` type resolution, unrelated to this change set.

lee-to (approved):
## Re-review on 05b3483

All blockers from the previous review are resolved. Approving.

### Addressed
- **1. Claude HTTP `type: "http"`** — ✅ `packages/runtime/src/adapters/claude/mcp.ts` now writes `{ type: "http", url }`; `claudeMcp.test.ts` asserts both the serialize and read-back paths, including the `getClaudeMcpStatus` round-trip.
- **2. `bearerTokenEnvVar` in Claude adapter** — ✅ Dropped entirely from the Claude HTTP entry. New test `does not write unsupported bearer token env vars for Claude HTTP MCP servers` pins the contract so it cannot silently regress.
- **3. IIFE-throw ternaries** — ✅ Both adapters switched to straightforward discriminant checks. Codex builds a `CodexMcpServerEntry` first, Claude uses a plain `if (transport === "streamable_http") / else`.
- **4. `RuntimeMcpInstallInput` discriminated union** — ✅ `packages/runtime/src/types.ts` is now a proper union with `never` guards on the opposite branch's fields. Runtime checks are no longer load-bearing for correctness.
- **5 (MCP_PORT validation)** — ✅ New `resolveMcpPort` helpers in both `scripts/dev.mjs` and `packages/api/src/routes/settings.ts` reject empty / non-integer / non-positive values via `Number.isInteger(port) && port > 0`. `"0"` no longer slips through.
- **8. stdio matrix in `settings.test.ts`** — ✅ The existing stdio install test now asserts the full Claude JSON entry and the Codex TOML contains `command = "npx"` and not the HTTP URL. Both branches locked down.

### Leftover nits (non-blocking, optional follow-ups)
- `resolveMcpPort` is duplicated between `scripts/dev.mjs` and `packages/api/src/routes/settings.ts`. Tiny helper, but a shared util would keep the two in sync if the validation rule ever changes.
- `parseEnvFile` still doesn't handle escape sequences or multi-line values — the new inline comment makes the limitation explicit, which is enough for now.
- `scripts/dev.mjs` and `packages/mcp/scripts/dev-http.mjs` still duplicate spawn/signal-forwarding. Pure nit.
- Docker compose sync was not part of this PR — please confirm separately that `docker-compose.yml` already launches an MCP service consistent with the README claim of "four services".

Nice iteration — type-level safety and the round-trip tests make the new HTTP path solid.

PR #60, opened 2026-04-11
fix(runtime): codex bypassPermissions parity with Claude

lee-to:
## Summary

- `AGENT_BYPASS_PERMISSIONS=1` now works for **both** Codex SDK and Codex CLI transports the same way it works for Claude — agent runs without approval prompts AND without the OS sandbox restricting file/network access.
- Drops `codexExecutionHooks.ts`. API/agent layers stop emitting provider-specific hook shapes; they just set `execution.bypassPermissions=true` and each adapter owns its translation (mirroring Claude's pattern).
- Profile-level overrides (`options.sandboxMode`, `options.approvalPolicy`) now work uniformly across SDK and CLI transports — explicit values always win over bypass defaults.

## Why both axes

Codex has two orthogonal safety rails (approval policy + OS sandbox) where Claude has one (permission prompts). Matching Claude's effective "agent can do anything" behaviour requires clearing both. Leaving the Codex sandbox at its default `workspace-write` blocks network access, so `npm install`, `curl`, `git push`, WebFetch would silently fail in unattended runs.

## Codex CLI implementation note

CLI uses `-c approval_policy="..."` + `-c sandbox_mode="..."` rather than the atomic `--dangerously-bypass-approvals-and-sandbox` flag. Verified directly via \`codex exec resume --help\` (and a live run) that:

- `--sandbox` is **rejected** by `codex exec resume` (works only on `exec`)
- `--config approval_policy=...` works on both `exec` and `exec resume`
- `--config sandbox_mode=...` works on both

The end-state is identical to the atomic flag, but routing through `-c` overrides is the only mechanism that works uniformly across resume and fresh-exec paths AND lets per-axis profile overrides flow through.

## Bonus parity fix

- `options.skipGitRepoCheck=true` → `--skip-git-repo-check` for Codex CLI (SDK already supported this; CLI didn't).

## Bypass translation table

| Runtime / transport | Bypass translation                                                            |
| ------------------- | ----------------------------------------------------------------------------- |
| Claude SDK          | \`permissionMode=\"bypassPermissions\"\` + \`allowDangerouslySkipPermissions=true\` |
| Claude CLI          | \`--dangerously-skip-permissions\`                                              |
| Codex SDK           | \`approvalPolicy=\"never\"\` + \`sandboxMode=\"danger-full-access\"\` (ThreadOptions) |
| Codex CLI           | \`-c approval_policy=\"never\" -c sandbox_mode=\"danger-full-access\"\`             |

## Test plan

- [x] \`npm run ai:validate\` — runtime 482/482, api 198/198, agent 202/202, data 122/122; format, build, type-check all green
- [x] Verified \`-c approval_policy=\"never\" -c sandbox_mode=\"danger-full-access\"\` actually executes a real \`echo test\` command via live \`codex exec\` (no approval prompts, no sandbox refusal)
- [x] Verified \`--sandbox\` is rejected by \`codex exec resume\` — confirms why \`-c\` route is required
- [ ] Manual smoke test in Docker with \`AGENT_BYPASS_PERMISSIONS=1\` and a Codex profile (SDK + CLI transports) — recommend the reviewer verify before merge
- [ ] Manual smoke test that an explicit \`options.sandboxMode=\"workspace-write\"\` in a profile overrides the bypass default for CLI

## Files

- **Removed:** \`packages/api/src/services/codexExecutionHooks.ts\` (provider-specific hook factory)
- **Modified:** \`packages/runtime/src/adapters/codex/sdk.ts\`, \`packages/runtime/src/adapters/codex/cli.ts\`, \`packages/api/src/services/runtime.ts\`, \`packages/api/src/routes/chat.ts\`
- **Tests:** \`packages/runtime/src/__tests__/codexCli.test.ts\` (+5), \`packages/runtime/src/__tests__/codexSdk.test.ts\` (+3)
- **Docs:** \`docs/providers.md\` — new "Bypass semantics" section + Codex CLI options note


ichinya:
**PR Review**

**Risk level:** Medium-High

The direction of the PR is good: moving Codex-specific permission translation out of the API layer and into the runtime adapters is the right abstraction boundary. The CLI change to use `-c approval_policy=...` and `-c sandbox_mode=...` instead of the atomic bypass flag also makes sense for `codex exec resume`.

That said, I see two issues that should be fixed before merge.

1. **[P1] Normal non-bypass Codex SDK runs no longer have stable safety defaults**  
   File: [packages/runtime/src/adapters/codex/sdk.ts](packages/runtime/src/adapters/codex/sdk.ts:206)

   After removing `getCodexExecutionHooks()` from the API layer, the SDK adapter now sets `approvalPolicy` and `sandboxMode` only inside the `execution.bypassPermissions` branch.

   The problem is that this changes behavior for ordinary runs too, not just bypass mode.

   Before this PR:
   - API explicitly injected Codex SDK defaults via `getCodexExecutionHooks()`
   - non-bypass runs still had explicit settings

   After this PR:
   - if `AGENT_BYPASS_PERMISSIONS=false`
   - and the profile does not set `options.approvalPolicy` / `options.sandboxMode`
   - then both values are left unset

   That means the effective behavior now depends on host-level Codex config or SDK defaults instead of application-controlled defaults. On a machine with permissive Codex config, a supposedly non-bypass run could become more permissive than intended.

   This is not just a refactor; it is a behavioral regression.

   **Why it matters**
   - runtime behavior becomes environment-dependent
   - safety semantics are no longer stable across hosts
   - this contradicts the PR’s stated goal of making behavior more uniform

   **Suggested fix**
   - preserve explicit non-bypass defaults for Codex SDK
   - then widen to `never` + `danger-full-access` only when `execution.bypassPermissions=true`
   - explicit profile overrides should still win

2. **[P2] `codexCliArgs` bypasses the new permission/sandbox logic entirely**  
   File: [packages/runtime/src/adapters/codex/cli.ts](packages/runtime/src/adapters/codex/cli.ts:112)

   If a profile defines `options.codexCliArgs`, `normalizeCliArgs()` returns immediately from the custom-args branch.

   As a result, the new logic added later in the function never runs:
   - `--skip-git-repo-check`
   - `-c approval_policy="..."`
   - `-c sandbox_mode="..."`
   - implicit bypass translation from `execution.bypassPermissions`

   So the PR’s “CLI parity” claim is only true for the default CLI path, not for custom CLI templates.

   **Why it matters**
   - custom Codex CLI profiles silently ignore explicit `approvalPolicy`
   - custom profiles silently ignore explicit `sandboxMode`
   - bypass mode does not actually behave consistently across all CLI profiles
   - `skipGitRepoCheck` parity is also incomplete

   **Suggested fix**
   - either merge generic execution controls into custom templates
   - or explicitly document that `codexCliArgs` is a full escape hatch and disables adapter-managed permission logic

**What the PR is trying to do**

The PR proposes to make `AGENT_BYPASS_PERMISSIONS=1` mean the same thing for Codex that it already means for Claude: unattended agent execution without approval prompts and without the runtime sandbox blocking work.

It does this by:
- deleting provider-specific Codex hook generation from the API layer
- passing generic `execution.bypassPermissions` through the runtime contract
- translating bypass inside the Codex adapters
- using `-c` config overrides for CLI so the same mechanism works for both fresh exec and resume
- adding tests for the Codex SDK and CLI adapters
- updating [docs/providers.md](docs/providers.md:185)

That overall design is sound.

**What is already done well**

- Removing Codex-specific branching from API services is a good cleanup.
- Adapter-owned translation is a better long-term design.
- The `codex exec resume` compatibility reasoning is solid.
- Test coverage was improved for the runtime adapter layer.
- The docs in `providers.md` are clearer than before.

**Gaps / follow-ups**

- [docs/configuration.md](docs/configuration.md:53) still describes `AGENT_BYPASS_PERMISSIONS` as effectively Claude-specific. That should be updated if this PR lands.
- API-level tests were not expanded enough around the new generic `execution.bypassPermissions` contract.
- I could not run the targeted test suites in the separate review worktree because that checkout did not have `vitest` installed, so the review is based on the PR diff and local CLI help validation.

**Bottom line**

I would not merge this PR as-is.

The architectural direction is correct, but it currently introduces:
- a real SDK behavior regression for non-bypass runs
- incomplete CLI parity when `codexCliArgs` is used

If those two points are fixed, the PR becomes much stronger.

ichinya (changes requested):
Finish your review by fixing the two blocking issues before merge: restore explicit safe defaults for non-bypass Codex SDK runs, and make `codexCliArgs` either inherit adapter-managed permission/sandbox overrides or document that it opts out of them entirely. After that, update the configuration docs and add API-level coverage for the new `execution.bypassPermissions` flow.

lee-to:
Addressed both blockers in e63907a.

**P1 — Codex SDK non-bypass regression**
Three-layer precedence in `codex/sdk.ts` and `codex/cli.ts`:
1. explicit `options.approvalPolicy` / `options.sandboxMode` (profile)
2. bypass defaults (`never` + `danger-full-access`) when `execution.bypassPermissions=true`
3. stable non-bypass defaults (`on-request` + `workspace-write`)

Non-bypass runs now always get explicit `approvalPolicy` and `sandboxMode` values regardless of the host `~/.codex/config.toml` — matching the old `getCodexExecutionHooks` contract exactly. Added SDK test `explicit profile options win over non-bypass defaults` and updated `applies stable non-bypass defaults` to assert the defaults flow through to `startThread`.

**P2 — `codexCliArgs` escape hatch**
Kept the intentional escape-hatch semantics (custom templates bypass all adapter-managed flags) but **explicitly documented** it in `docs/providers.md` § Codex CLI:

> `codexCliArgs` is a full escape hatch. When set, the adapter uses the custom template verbatim and skips all adapter-managed flags — including `--model`, `-c model_reasoning_effort`, `-c approval_policy`, `-c sandbox_mode`, `--skip-git-repo-check`, and bypass-permission translation. Profile-level `options.*` and `AGENT_BYPASS_PERMISSIONS` all have **no effect** when a custom template is active.

Corresponding test renamed to `custom codexCliArgs is a full escape hatch — adapter-managed flags are NOT injected` so the intent is self-documenting.

**Follow-ups from your review**

- ✅ `docs/configuration.md` — rewrote `AGENT_BYPASS_PERMISSIONS` as provider-neutral with per-adapter translation table (Claude vs Codex)
- ✅ API-level test — `runtimeService.test.ts` now asserts `execution.bypassPermissions=true` AND `=false` actually flow from `runApiRuntimeOneShot` into `adapter.run()`

**Validation**
- runtime: 483/483 (+1 new SDK test)
- api: 199/199 (+1 new bypass-false test)
- agent: 202/202
- `npm run ai:validate` green

Ready for re-review.

ichinya:
Re-reviewed `e63907a`.

The two blockers from my previous review are addressed:

- `P1` is resolved: non-bypass Codex SDK runs now use explicit stable defaults (`on-request` + `workspace-write`) instead of inheriting host-level Codex config, while explicit profile options still take precedence.
- `P2` is resolved to the documented-contract level: `codexCliArgs` remains an intentional escape hatch, and that behavior is now explicitly documented in `docs/providers.md`.

The follow-up doc update in `docs/configuration.md` and the added API/runtime coverage also close the remaining gaps I called out.

No further blockers from my side.

ichinya (commented):
merge

PR #62, opened 2026-04-11
feat(runtime): universal LLM token usage tracking

lee-to:
## Summary

- Adds an end-to-end usage-tracking pipeline so every successful LLM call that flows through the runtime registry is persisted to a new `usage_events` table, rolled up into per-task / per-project / per-chat-session aggregates, and tagged with a canonical `source`.
- Enforced via types + runtime wrapper + discovery test: new call sites that forget to pass `usageContext` fail compilation, and new adapters that omit `usageReporting` fail the build on `bootstrap.test.ts`.
- Fixes the three previously-untracked call sites (chat, commit generation, roadmap generate) and removes the three manual `incrementTaskTokenUsage` call sites that are now handled by the sink (subagentQuery, fastFix, roadmap-extract).

## Architecture

Two independent safety contours:

**Call-site contour — "new route cannot forget to tag a run":**
- `RuntimeRunInput.usageContext` is required (`source: UsageSource`, optional `projectId` / `taskId` / `chatSessionId`).
- Registry wrapper reads `result.usage` after every successful run and forwards the event to a `RuntimeUsageSink`.
- Adapter factories are no longer re-exported from `@aif/runtime` public API, so callers cannot bypass the wrapper.

**Adapter contour — "new adapter cannot forget to report usage":**
- `RuntimeRunResult.usage` is required (`RuntimeUsage | null`, no `undefined`).
- `RuntimeCapabilities.usageReporting` is required (const-object enum: `UsageReporting.FULL` / `PARTIAL` / `NONE`, same pattern as `RuntimeTransport`).
- Registry wrapper enforces the contract: `FULL` must return non-null, `NONE` warns if usage appears, `PARTIAL` accepts both.
- Discovery test in `bootstrap.test.ts` iterates every built-in adapter and fails the build if any is missing a valid `usageReporting` value.

## Data layer

- Drizzle migration `9` adds `usage_events` table (append-only, indexed on `project_id`, `task_id`, `chat_session_id`, `source`, `runtime_id`) and token/cost aggregate columns on `projects` and `chat_sessions`.
- `@aif/data` exposes `createDbUsageSink()` — structurally compatible with `RuntimeUsageSink` so `@aif/data` stays free of a dependency on `@aif/runtime`.
- `recordUsageEvent` inserts the event row and fans out to `incrementTaskTokenUsage` / `incrementProjectTokenUsage` / `incrementChatSessionTokenUsage` based on the event's scope.

## Call sites wired

| Call site | Previous state | New source |
|---|---|---|
| Chat route | Usage ignored | `UsageSource.CHAT` (+ `chat:done` WS event now carries usage) |
| Commit generation | Fire-and-forget, usage dropped | `UsageSource.COMMIT` |
| Roadmap generate | Usage dropped | `UsageSource.ROADMAP_GENERATE` |
| Roadmap extract | Manual increment | `UsageSource.ROADMAP_EXTRACT` |
| Fast fix | Manual increment | `UsageSource.FAST_FIX` |
| Subagents | Manual increment | `UsageSource.SUBAGENT` |
| Claude CLI model discovery | n/a (no usageContext existed) | `UsageSource.MODEL_DISCOVERY` |

Manual `incrementTaskTokenUsage` calls in `subagentQuery.ts`, `fastFix.ts`, and `roadmapGeneration.ts` have been removed — the sink records once at the registry layer, so there is no double counting.

## Adapter declarations

| Adapter | Transport | `usageReporting` |
|---|---|---|
| Claude | SDK / CLI / API | `FULL` |
| Codex | SDK / API | `FULL` |
| Codex | CLI | `PARTIAL` (early-termination paths may return null) |
| OpenRouter | API | `FULL` |
| OpenCode | API | `NONE` (transport does not surface token counts) |

## Tests added

- `packages/runtime/src/__tests__/bootstrap.test.ts` — discovery test asserting every built-in adapter declares a valid `usageReporting` value.
- `packages/runtime/src/__tests__/registry.test.ts` — 5 new sink-contract tests: forwards usage to sink with correct context, skips when `usage === null`, logs error on FULL-but-null violation, warns on NONE-but-non-null, and swallows sink errors without breaking the caller.

## Docs

- `docs/providers.md` — new `Usage Reporting` column in Supported Runtimes table + new "Usage reporting contract" section.
- `CLAUDE.md` + `AGENTS.md` — Runtime Adapter Sync Rule now requires declaring `usageReporting` and references the discovery test as the safety net.
- `packages/runtime/src/adapters/TEMPLATE.ts` — required-fields comments on `usageReporting` and `usage: null`.

## Test plan

- [x] `npx turbo build` — 7/7 successful
- [x] `npx turbo test` — 10/10 successful (485+ runtime tests, 203 api tests, all agent/data/shared/mcp tests)
- [x] `npx turbo lint` — 10/10 clean
- [x] `npm run format:check` — clean
- [x] `npx tsc --noEmit` across all packages — zero errors
- [ ] Smoke test: start API + agent, trigger chat / commit / fast-fix / subagent flows, verify rows appear in `usage_events` with correct `source` and scope columns
- [ ] Manual check: `chat:done` WS event payload now contains `usage` shape for per-turn display

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -157,10 +157,11 @@ data/                    # SQLite database files (gitignored)
 ## Runtime Adapter Sync Rule
 
 - **Docs must stay in sync when adding or modifying runtime adapters.** When a new adapter is added to `packages/runtime/src/adapters/` or an existing adapter's capabilities change:
-  - `docs/providers.md` — update the "Supported Runtimes" table.
+  - `docs/providers.md` — update the "Supported Runtimes" table (including the `Usage Reporting` column).
   - `packages/runtime/src/adapters/TEMPLATE.ts` — verify the template still reflects current conventions.
   - `packages/runtime/src/bootstrap.ts` — register the new built-in adapter.
   - `.docker/Dockerfile` — add any new system-level dependencies.
+  - **Usage reporting contract** — declare `capabilities.usageReporting` (`FULL` / `PARTIAL` / `NONE`) and return `RuntimeRunResult.usage` as `RuntimeUsage` or explicit `null`. The discovery test in `bootstrap.test.ts` fails the build if the field is missing.
 
 ## Project Rules
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -185,10 +185,11 @@ data/                    # SQLite database files (gitignored)
 ## Runtime Adapter Sync Rule
 
 - **Docs must stay in sync when adding or modifying runtime adapters.** When a new adapter is added to `packages/runtime/src/adapters/` or an existing adapter's capabilities change:
-  - `docs/providers.md` — update the "Supported Runtimes" table with the new adapter's capabilities, transports, and light model.
+  - `docs/providers.md` — update the "Supported Runtimes" table with the new adapter's capabilities (including the `Usage Reporting` column), transports, and light model.
   - `packages/runtime/src/adapters/TEMPLATE.ts` — verify the template still reflects current conventions.
   - `packages/runtime/src/bootstrap.ts` — register the new built-in adapter (or document `AIF_RUNTIME_MODULES` loading).
   - `.docker/Dockerfile` — add any new system-level dependencies the adapter needs.
+  - **Usage reporting contract** — every adapter must declare `capabilities.usageReporting` (`FULL` / `PARTIAL` / `NONE`) and return `RuntimeRunResult.usage` as a concrete value (including explicit `null`). The discovery test in `packages/runtime/src/__tests__/bootstrap.test.ts` fails the build if a new adapter ships without a valid `usageReporting` value. See `docs/providers.md` → "Usage reporting contract".
 
 ## Project Rules
 
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -50,15 +50,25 @@ The API exposes effective selection endpoints:
 
 ## Supported Runtimes
 
-| Runtime      | Provider     | Transports    | Resume         | Sessions       | Agent Defs    | Light Model         | Status                    |
-| ------------ | ------------ | ------------- | -------------- | -------------- | ------------- | ------------------- | ------------------------- |
-| `claude`     | `anthropic`  | SDK, CLI, API | Yes (SDK/CLI)  | Yes (SDK/CLI)  | Yes (SDK/CLI) | `claude-haiku-3-5`  | Built-in                  |
-| `codex`      | `openai`     | SDK, CLI, API | Yes (SDK only) | Yes (SDK only) | No            | default             | Built-in                  |
-| `opencode`   | `opencode`   | API           | Yes            | Yes            | No            | null (configurable) | Built-in                  |
-| `openrouter` | `openrouter` | API           | No             | No             | No            | null (configurable) | Built-in                  |
-| Custom       | Any          | Any           | Configurable   | Configurable   | Configurable  | Configurable        | Via `AIF_RUNTIME_MODULES` |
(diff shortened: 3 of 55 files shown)


ichinya:
**Notes**
1. `[P2]` `packages/data/src/index.ts:902-944`  
`recordUsageEvent()` by comment and PR description should write event and aggregates atomically, but now there is just `insert` + up to three separate `update(...).run()` without `db.transaction(...)`. If one of the updates fails after inserting a row in `usage_events`, the log and aggregates will diverge at the point where the PR promises consistency.

2. `[P2]` `packages/api/src/services/runtime.ts:38-46` and `packages/agent/src/subagentQuery.ts:150-158`  
The new safety-net in registry relies on `log.error` for cases of `usageReporting=FULL but null`, missing `usageContext.source`, and sink errors (`packages/runtime/src/registry.ts:94-108, 128-159`). However, both the API and the agent only pass loggers with `debug`/`warn`, not `error`, so these violations will be silently ignored in real processes.

3. `[P3]` `packages/runtime/src/registry.ts:94-110, 138-149`  
The contract in `packages/runtime/src/types.ts:266-273` says that `usage` is required to be either an object or `null`, and `undefined` is not allowed. But the runtime check only catches `=== null`. For an external/JS adapter that simply forgets the `usage` field, the wrapper will not detect a FULL contract violation and will attempt to send `undefined` to the sink.

lee-to:
All three items addressed in 326fe1f:

1. **[P2] Transaction** — `recordUsageEvent` now wraps the insert + all aggregate updates in `db.transaction()`, so the append-only log and per-entity counters are atomically consistent. If any update fails, the entire batch rolls back.

2. **[P2] Logger error callback** — both the API (`packages/api/src/services/runtime.ts`) and agent (`packages/agent/src/subagentQuery.ts`) registry loggers now pass an `error` callback alongside `debug`/`warn`, so the wrapper's FULL-contract violation log and sink-error log actually surface in production.

3. **[P3] undefined usage** — the registry wrapper's usage check changed from `=== null` to `== null` (loose equality), so external JS adapters that return `undefined` (missing field) are caught by the FULL-contract assertion instead of silently passing through.

ichinya:
There is one more remark before merge.

 `usage_events.total_tokens` is saved from `usage.totalTokens`, but aggregates on `tasks / projects / chat_sessions` are now increasing not by `usage.totalTokens`, but by the formula `inputTokens + outputTokens`.

 This looks dangerous: while the current adapters may have matching values, the contract already allows a separate field `totalTokens`, and it is not obliged to always be equal to `input + output`. As soon as a provider has additional token categories or a different reporting scheme, the append-only log and rolled-up counters will start to diverge.

 I would suggest using the same source of truth in the aggregates as in `usage_events`, i.e. `usage.totalTokens`, rather than manually recalculating the total.

 Otherwise, you will end up with an unpleasant situation:
- in `usage_events`, there is one value for `total_tokens`
- in `projects/chat_sessions/tasks.token_total`, there is another value
- UI and future analytics may show different numbers for the same series of calls

 Otherwise, I see the previous comments as closed: the transaction has been added, the `error` logger has been thrown, and `undefined` for `usage` is now caught.

lee-to:
Addressed in 19882f0:

**totalTokens divergence** — aggregates on `tasks/projects/chat_sessions` now increment `token_total` by `usage.totalTokens` (the provider's authoritative total) instead of recalculating `inputTokens + outputTokens`. This matches the value stored in `usage_events.total_tokens` — same source of truth everywhere. Providers that include additional token categories (cache, reasoning, etc.) in their total will no longer cause the log and rolled-up counters to diverge.

ichinya:
Ready for the merge

PR #64, opened 2026-04-12
fix(runtime): classify errors by HTTP status instead of message patterns

### LINKED ISSUE #61: Runtime error classification relies on message patterns instead of HTTP status codes
## Problem

Runtime adapters (`claude`, `codex`, `openrouter`, `opencode`) classify errors into semantic categories (`rate_limit`, `auth`, `timeout`, etc.) by **pattern-matching against error message strings** instead of using structured signals like HTTP status codes.

See `packages/runtime/src/adapters/*/errors.ts` — each adapter defines arrays like `RATE_LIMIT_PATTERNS`, `AUTH_PATTERNS`, then runs `lowered.includes(pattern)` over `error.message`.

## Why this is fragile

- Provider error wording changes silently. Example: Claude's actual usage-limit message is `"You've hit your limit · resets ..."`, which did not match any of the existing `USAGE_LIMIT_PATTERNS` (`"usage limit"`, `"out of extra usage"`, `"rate limit"`, `"quota"`). Result: the error was classified as `unknown`, the chat route fell back to a hardcoded `"Chat request failed"` string, and the real reset-time message only became visible after a page refresh (which loaded the message from the runtime session log).
- Every new phrasing requires a code change in 4 separate adapters.
- Pattern lists drift apart between adapters and miss obvious cases (e.g. `"limit reached"`, `"limit exceeded"`, `"out of credits"`).
- `lowered.includes("401")` / `"403"` is a poor proxy for an actual HTTP status — it can match unrelated substrings (timestamps, error codes, request IDs).

## Why it ended up this way

`RuntimeAdapter` unifies multiple transports per provider (SDK, CLI, API), and **not all of them expose HTTP status codes**:

1. **SDK transports** (e.g. `@anthropic-ai/claude-agent-sdk`) wrap HTTP internally and surface only `Error(message)` or `result.subtype` strings.
2. **CLI transports** (`claude` CLI, `codex` CLI) are spawned subprocesses — only stdout/stderr and exit codes are available, no HTTP status at all.
3. **API transports** (`openrouter/api.ts`, `codex/api.ts`) *do* have access to the HTTP response, but they currently collapse errors into `Error(message)` before the shared classification layer ever sees the status.

So the lowest common denominator across all transports became `error.message`, and pattern matching grew on top of it.

## Proposed direction

- In **API transports**, capture the HTTP status (and provider error code/body, if any) at the point of failure and construct the categorized `*RuntimeAdapterError` directly — e.g. `429 → rate_limit`, `401/403 → auth`, `408/504 → timeout` — with **zero string matching**.
- In **SDK/CLI transports**, prefer structured signals when the SDK exposes them (e.g. Claude SDK `result.subtype`, Codex SDK error classes) before falling back to message inspection.
- Keep string pattern matching only as a **last-resort fallback** for transports that genuinely have nothing else (rare CLI cases, SDKs that throw plain `Error(string)`).
- Consider hoisting the shared category logic (status → category) into `packages/runtime/src/errors.ts` so all API-style adapters use the same mapping instead of each maintaining its own list.

## Impact

- Frontend stops showing generic `"Chat request failed"` for any error the pattern list happens to miss.
- Adding new providers no longer requires curating 5+ regex-style arrays.
- Provider wording changes (which happen all the time) stop silently degrading the UX.

## Related

A short-term patch was applied to extend the pattern lists with `"hit your limit"`, `"limit reached"`, `"limit exceeded"`, `"out of credits"` and to forward the original error message in `classifyChatError` instead of the hardcoded `"Chat request failed"` fallback. That fix treats the symptom; this issue tracks the structural fix.

lee-to:
## Summary

- Replace fragile string-based error classification across all runtime adapters with structured HTTP status codes as primary signal
- Consolidate 7 independent pattern arrays (4 adapters + agent + 3 diagnostics) into a single shared fallback in `@aif/runtime/errors.ts`
- Add `classifyByHttpStatus()` for API transports — zero string matching for HTTP-based errors
- Refactor agent `errorClassifier.ts` to dispatch on `RuntimeExecutionError.category` instead of maintaining its own 28-pattern list
- Add "No string-based error classification" rule to `CHECKLIST.md` to prevent regression

Closes #61

## Test plan

- [x] `classifyByHttpStatus()` maps all HTTP statuses correctly (401/403→auth, 429→rate_limit, 408/504→timeout, 404→model_not_found, 5xx→transport)
- [x] `classifyByMessageFallback()` covers all 9 categories as last-resort fallback
- [x] All 4 adapter classifiers accept `httpStatus` param and prefer it over message matching
- [x] API transports (`openrouter/api.ts`, `codex/api.ts`) pass `response.status` to classifiers
- [x] Agent `isExternalFailure()` works with `RuntimeExecutionError` categories
- [x] Diagnostic builders dispatch on `.category` first, fall back to strings for plain `Error`
- [x] `npm run ai:validate` passes (lint + build + tests + format)


ichinya:
Overall direction looks good — moving consumers to `RuntimeExecutionError.category` and consolidating fallback classification into shared runtime code is the right cleanup.

I still see a few blockers before merge:

1. **OpenCode is still on the legacy message-parsing path for HTTP failures.** `requestJson()` throws `Error("OpenCode HTTP ...")`, and the catch path calls `classifyOpenCodeRuntimeError(error)` without `response.status`. So this adapter still depends on parsing the error string instead of the structured status.

2. **OpenCode timeout classification may regress.** Timeouts are implemented with `AbortController`, but the shared timeout fallback no longer matches abort-style messages. That means request aborts can now fall through to `unknown` instead of `timeout`.

3. **`classifyOpenRouterRuntimeError()` should be idempotent.** The streaming path can already throw a classified `OpenRouterRuntimeAdapterError` with `response.status`, and then `runOpenRouterApiStreaming()` catches and classifies it again without the status. That second pass can drop a previously correct category. Codex already protects against this by returning early when the error is already adapter-typed.

4. **Smaller design concern:** the shared `404 -> model_not_found` mapping feels too broad for discovery endpoints like `/models`. There, a 404 can also mean a bad `baseUrl` / route configuration rather than an invalid model id.

Once those are addressed, I think the refactor is in good shape.

ichinya (changes requested):
At the very least, OpenCode still looks like it's relying on parsing the error text, and OpenRouter should be rechecked for reclassifying an already labeled error.

lee-to:
All four review items addressed in 7347cd7:

1. **OpenCode HTTP status** — `requestJson()` now passes `response.status` to `classifyOpenCodeRuntimeError()` instead of embedding it in the error message string. Also handles `AbortController` abort as an explicit timeout.
2. **OpenCode timeout regression** — Added `"aborted"` to shared timeout fallback patterns, plus explicit `AbortError` → timeout classification in `requestJson()`.
3. **OpenRouter idempotency** — Added early-return guard for already-classified `OpenRouterRuntimeAdapterError` (matching the existing Codex pattern). Also added to OpenCode for consistency.
4. **404 removed from shared map** — Removed `404 → model_not_found` from `HTTP_STATUS_CATEGORY_MAP`. A 404 on `/models` means model-not-found, but on `/chat/completions` it means bad route/baseUrl. Adapters now classify 404 per-endpoint via message fallback.

All tests pass (`npm run ai:validate` green).

ichinya:
Request changes

Thanks, this is much closer.

The earlier issues I called out seem addressed:
- OpenCode now passes `response.status` into classification instead of relying on the `"OpenCode HTTP ..."` message path.
- Abort-driven OpenCode timeouts are mapped back to `timeout`.
- OpenRouter classification is now idempotent for already-classified adapter errors.
- The shared `404 -> model_not_found` mapping was removed from the global HTTP map.

I still have one blocking concern:

`getOpenCodeSession()` now treats any classified error whose message contains `"not found"` as a missing session and returns `null`.

I don't think we should key this path off `classified.message.toLowerCase().includes("not found")`. That reintroduces local string-based classification outside the shared runtime error module, and more importantly it can hide a misconfigured `baseUrl` / route as “session not found” if the server returns a generic 404 body like `Not Found`.

Could we make the session-not-found path structured as well — for example by preserving enough endpoint/status context to map `GET /session/:id` 404 to a dedicated adapter code, instead of checking the message text here?

Once that is fixed, this looks good to merge.

lee-to:
Addressed in 63b8c66:

**`getOpenCodeSession()` no longer keys on message text.** Added `httpStatus?: number` to `OpenCodeRuntimeAdapterError` — the classifier now preserves the HTTP status on the error object. `getOpenCodeSession()` checks `classified.httpStatus === 404` instead of `message.toLowerCase().includes("not found")`.

Also added the "Structured Error Classification Rule" to CLAUDE.md / AGENTS.md to prevent message-based branching in the future.

ichinya:
**Request changes**

This is very close. I still see one functional issue in the Codex classifier. `CLI_NOT_FOUND_PATTERNS` contains a bare `"not found"` and that branch runs before both `THREAD_PATTERNS` and the shared fallback. Since 404 now intentionally falls through to adapter-specific handling, an API 404 body like `"thread not found"` or `"model not found"` will currently classify as `CODEX_CLI_NOT_FOUND` / `unknown` instead of `CODEX_THREAD_NOT_FOUND` or `CODEX_MODEL_NOT_FOUND`. The shared fallback explicitly maps `"model not found"` to `model_not_found`, so the current ordering defeats the new category. Could we either remove the generic `"not found"` token from `CLI_NOT_FOUND_PATTERNS`, or move the CLI-not-found branch after the more specific thread/shared checks?

Smaller follow-up: the new checklist wording is stricter than the implementation. It says string matching should be consolidated in the shared category module and never duplicated per adapter or consumer, but this PR still intentionally keeps local string matching in places like the agent capability fallback, Claude resume handling, and Codex-specific special cases. I’d narrow that rule to shared category classification specifically, or explicitly document the exceptions so the docs match the code.

Once the Codex ordering issue is fixed, this looks good to merge.

lee-to:
Fixed in e43df62:

1. **Codex classifier ordering** — removed bare `"not found"` from `CLI_NOT_FOUND_PATTERNS` (it was shadowing thread/model patterns). Reordered so `THREAD_PATTERNS` runs before CLI patterns — more specific checks first.

2. **Checklist wording narrowed** — adapter-specific string patterns are now explicitly allowed in each adapter's `errors.ts`, but must be ordered most-specific-first. Matches the actual implementation (agent capability fallback, Claude resume handling, Codex CLI detection).

ichinya:
LGTM

ichinya (approved):
LGTM

PR #65, opened 2026-04-12
Stabilize auto-review convergence and add explicit manual review handoff

ichinya:
You can use this:

**PR Title**

`Stabilize auto-review convergence and add explicit manual review handoff`

**PR Description**

## Summary

This PR fixes the review loop so auto-review is convergence-aware instead of repeatedly cycling without a clear stopping condition.

## Problem

The previous auto-review flow could keep sending a task back to implementation after review feedback, but it had weak convergence control:

- blocking findings were not persisted as structured state across review iterations
- the system could continue reworking without a clear distinction between "old blockers still unresolved" and "new blockers introduced later"
- when auto-review stopped being reliable, there was no explicit manual handoff state in the task model and UI
- related API/UI surfaces did not expose enough information for operators to understand why automation stopped

In practice, this made review outcomes harder to reason about and created cases where automation could keep looping or appear to succeed without a clear human-review checkpoint.

## Solution

This PR introduces a structured, persisted auto-review flow with explicit manual handoff semantics:

- adds persisted `autoReviewState` and `manualReviewRequired` task fields
- formalizes review output and gate evaluation through a structured review contract
- makes the coordinator stop automation explicitly when review no longer converges
- keeps the task in `done`, but marks it as requiring human review instead of silently treating it as fully resolved
- exposes the new state through API, MCP, and UI so the handoff is visible to operators
- documents the behavior in README and API/config docs

This makes the review loop predictable:

- if blockers are resolved, the task can complete normally
- if blockers remain, the task can be sent back for rework
- if the loop stops converging, the system escalates to explicit manual review

## Configuration

This PR also introduces the `AGENT_AUTO_REVIEW_STRATEGY` parameter.

Supported values:

- `full_re_review`
  Re-runs the broader automatic review loop on each cycle. This is the default behavior.

- `closure_first`
  First checks whether previously blocking findings were actually resolved. If previous blockers are closed but new blockers appear, the system stops the automatic loop and hands the task off for manual review.

In short:

- `full_re_review` favors more automation
- `closure_first` favors safer convergence and earlier human handoff

## Result

The review pipeline is now safer and easier to operate:

- review state is durable and explicit
- automatic rework is no longer "open-ended"
- manual review handoff is visible and intentional
- operators can see both the selected strategy and the reason automation stopped

If you want, I can also make a shorter PR version for GitHub, with a more compact “Problem / Solution / Config” format.


lee-to (commented):
## Code Review — PR #65

### Overall Assessment

Strong, well-structured PR. The convergence-aware auto-review loop with explicit manual handoff is a clear improvement over the previous open-ended rework cycle. The structured review contract (`reviewContract.ts`), deterministic parsing, and fallback path are well-designed. Test coverage is thorough across all layers.

### Findings

#### 1. Migration 10 rewrite is risky for existing databases (medium severity)

Migration 10 on `main` is a simple `UPDATE projects` backfill. This PR rewrites it to also `CREATE TABLE IF NOT EXISTS usage_events` and `ALTER TABLE` for projects/chat_sessions columns. For databases that already ran the original migration 10 (where `user_version` is already ≥ 10), this rewritten migration will never execute — those databases will only see migration 11.

The `isIgnorableMigrationError` safety net catches `duplicate column name` / `already exists`, so fresh databases won't break. But the intent seems to be convergence for diverged DB histories — if so, this should be documented explicitly (e.g. a comment in the migration explaining why it's safe to rewrite). Otherwise, the new statements in v10 should be moved to v11 or a new v12 to guarantee they run on all databases.

**File:** `packages/shared/src/db.ts` (migration version 10)

#### 2. `manualReviewRequired` and `autoReviewState` are optional in the `Task` type (low severity)

```ts
// packages/shared/src/types.ts:118-119
manualReviewRequired?: boolean;
autoReviewState?: AutoReviewState | null;
```

These are marked optional (`?`) in the `Task` interface, but the schema column has `NOT NULL DEFAULT false` / nullable. Since every row will have these values, making them required in the type (non-optional `boolean` and `AutoReviewState | null`) would be more accurate and avoid unnecessary `?.` chains downstream.

#### 3. `text-3xs` / `text-4xs` CSS utility usage

```tsx
// packages/web/src/components/kanban/TaskCard.tsx:103
className="mt-2 ml-2 border border-amber-500/30 bg-amber-500/10 px-2 py-1 text-3xs ..."
```

`text-3xs` and `text-4xs` are non-standard Tailwind classes — verify these are configured in the project's Tailwind config. If they resolve to very small sizes (< 10px), consider accessibility implications (WCAG minimum text size).

#### 4. `CLEAN_STATE_RESET` now used for `approve_done` — verify downstream assumptions

The `approve_done` transition was changed from a targeted `{ status: "verified", retryCount: 0, reworkRequested: false }` to `{ ...CLEAN_STATE_RESET, status: "verified" }`. This now additionally resets `reviewIterationCount`, `manualReviewRequired`, and `autoReviewState`. This is correct behavior, but check whether any code downstream relies on reading those fields from a `verified` task for reporting/audit purposes — the reset clears the convergence history.

**File:** `packages/shared/src/stateMachine.ts:71`

#### 5. `buildFallbackDecision` assumes all previous findings are still blocking (conservative, potentially noisy)

```ts
// packages/agent/src/reviewGate.ts — buildFallbackDecision
stillBlockingCount: input.previousFindings.length > 0 ? input.previousFindings.length : 0,
```

When the review output is malformed and falls back to legacy extraction, the code treats **all** previous findings as still-blocking (since it can't verify their status). This is the safe default, but it means any malformed review output after iteration 1 will always trigger `manual_review_required` with `malformed_review_output_fallback`. This is probably intentional — just noting that a single malformed sidecar output will halt the entire auto-review loop, which could be surprising in practice.

#### 6. `env` module-level initialization in `autoReviewHandler.ts` and `reviewGate.ts`

```ts
const env = getEnv();
```

Both `autoReviewHandler.ts` and `reviewer.ts` call `getEnv()` at module load time. If the env isn't configured when these modules are first imported (e.g. during test bootstrapping without env setup), this could throw. The tests seem to handle it via mocking, but if import order changes, this could become fragile. Consider lazy initialization or accessing `getEnv()` inside the function body where the value is used.

### Positives

- The structured review contract with deterministic parsing (no LLM in the happy path) is a significant reliability improvement
- Finding IDs via SHA1 hashing provide stable cross-iteration tracking
- The `closure_first` strategy is a thoughtful addition for safer convergence
- Fallback path for legacy/malformed output is well-handled
- Test coverage is comprehensive — the new `reviewContract.test.ts` and expanded `reviewGate.test.ts` cover the critical parsing and decision logic
- Documentation updates (README, API, architecture, configuration) are thorough and synchronized
- The `useChat.test.ts` fix for the unhandled promise rejection is a nice incidental cleanup

### Summary

Approve with minor suggestions. The migration rewrite (finding #1) is the most significant concern — worth a quick check to confirm the intent is correct for existing databases. Everything else is solid.

ichinya:
Addressed all actionable feedback from review.

Changes made:
- Documented why migration 10 is safe for existing databases with diverged version-9 histories, and added a regression test covering that reconciliation path.
- Tightened the shared task contract by making `manualReviewRequired` and `autoReviewState` required, then synced affected consumers and test fixtures.
- Removed import-time `getEnv()` reads from the review path so env resolution now happens at execution time.
- Raised the `text-4xs` microcopy floor to `10px` to avoid the previous 9px accessibility issue.
- Kept the `approve_done` reset behavior and conservative fallback behavior as intentional, but added code comments clarifying the reasoning:
  - verified tasks clear live convergence state because they are terminal;
  - malformed fallback output preserves prior blockers and escalates instead of guessing convergence.

Validation:
- `npm test --workspace @aif/agent`
- `npm test --workspace @aif/shared`
- `npm test --workspace @aif/web`
- `npm run ai:validate`

lee-to:
## Follow-up Review — Fix Commits Verified

Checked both fix commits (`6b11620`, `423d6c2`). All six findings from the initial review are addressed:

| # | Finding | Status |
|---|---------|--------|
| 1 | Migration 10 rewrite safety | Fixed — explanatory comment added + diverged-v9 migration test |
| 2 | Optional types for required fields | Fixed — `manualReviewRequired` and `autoReviewState` now required in `Task` interface, all test fixtures updated |
| 3 | `text-4xs` accessibility | Fixed — raised from 9px to 10px with accessibility floor comment |
| 4 | `CLEAN_STATE_RESET` on approve_done | Fixed — clarifying comment added (audit trail in comments/log, not live state) |
| 5 | Conservative fallback behavior | Fixed — rationale comment added in `buildFallbackDecision` |
| 6 | Module-level `getEnv()` | Fixed — moved to function body in `autoReviewHandler.ts`, `reviewer.ts`, and `subagentQuery.ts` |

Fixes are minimal and well-scoped. LGTM.

PR #66, opened 2026-04-12
fix(runtime): forward HTTP proxy env vars to spawned subprocesses

rootbdfy:
## Summary

- Add `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` (and lowercase variants) to all five
  `ALLOWED_ENV_PREFIXES` allowlists across Claude and Codex adapters
- Add build-time proxy `ARG`/`ENV` to Dockerfile (`base` + `web` stages)
- Document proxy configuration in `.env.example`

## Problem

In closed-network environments where outbound traffic goes through an HTTP proxy,
agent subprocesses (Claude SDK, Claude CLI, Codex SDK/CLI) fail because the runtime
adapters filter out proxy environment variables before spawning child processes.

The proxy vars are set at the container level via `env_file`, but the adapter's
`buildCuratedEnv()` / `resolveEnvironment()` functions only forward variables matching
an explicit allowlist — proxy vars were not on it.

## Affected files

| File | Transport |
|------|-----------|
| `packages/runtime/src/adapters/claude/options.ts` | Claude SDK |
| `packages/runtime/src/adapters/claude/cli.ts` | Claude CLI |
| `packages/runtime/src/adapters/codex/cli.ts` | Codex CLI |
| `packages/runtime/src/adapters/codex/sdk.ts` | Codex SDK |
| `packages/runtime/src/adapters/codex/modelDiscovery/process.ts` | Codex model discovery |
| `.docker/Dockerfile` | Build-time proxy |
| `.env.example` | Documentation |

## Test plan

- [x] `docker compose exec -T -u node agent claude -p "respond with OK"` succeeds through proxy
- [x] Agent picks up task and `plan-coordinator` completes without errors
- [x] Verified proxy vars are present in spawned subprocess environment

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,6 +1,17 @@
 # ---- Base ----
 FROM node:22-slim AS base
 WORKDIR /app
+# Proxy (optional) — propagated to all downstream stages via ENV,
+# so apt-get / npm / npx all pick it up during build.
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+ARG NO_PROXY
+ENV HTTP_PROXY=${HTTP_PROXY} \
+    HTTPS_PROXY=${HTTPS_PROXY} \
+    NO_PROXY=${NO_PROXY} \
+    http_proxy=${HTTP_PROXY} \
+    https_proxy=${HTTPS_PROXY} \
+    no_proxy=${NO_PROXY}
 
 # ---- Dependencies ----
 FROM base AS deps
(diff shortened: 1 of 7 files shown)


lee-to:
## Code Review — PR #66

### Overall verdict: ✅ Approve

Clean, well-scoped fix for a real problem — proxy env vars were silently dropped by the allowlist filtering in all subprocess-spawning adapters.

### What's good

- **Correct scope.** All five `ALLOWED_ENV_PREFIXES` lists that gate subprocess environments are updated. OpenRouter and OpenCode adapters correctly skipped — they use HTTP clients in-process (no `spawn`, no env filtering).
- **Docker coverage.** Proxy ARGs are added to both `base` and `web` stages, with a clear comment explaining why the `web` stage needs its own set (it doesn't inherit from `base`). Lowercase aliases are mapped correctly.
- **Documentation.** `.env.example` includes a helpful `NO_PROXY` hint about compose service names — important for avoiding proxy loops on internal traffic.
- **Zero deletions.** Pure additive change, no risk of regression.

### Suggestions (non-blocking)

1. **DRY opportunity.** The same 6-line proxy block is copy-pasted into 5 files. Consider extracting a shared constant (e.g. `PROXY_ENV_KEYS` in `@aif/runtime/src/env-allowlist.ts` or similar) and spreading it into each adapter's `ALLOWED_ENV_PREFIXES`. This would make future allowlist additions a single-line change. Not urgent — the current duplication is small and obvious — but worth considering if more cross-adapter env vars come up.

2. **Prefix matching semantics.** The filtering logic uses `key.startsWith(prefix)`, so `"HTTP_PROXY"` in the list would also match hypothetical keys like `HTTP_PROXY_AUTH`. This is harmless in practice (no standard env vars collide), but worth noting as a design subtlety.

3. **Docker layer caching.** The proxy `ARG`s are placed early in the `base` stage. If proxy values change between builds, all subsequent layers are invalidated. This is probably fine (proxy config rarely changes), but if build speed matters, moving the `ARG`s closer to the install steps would improve cache hit rates.

### Security check

- No secrets are hardcoded — proxy URLs come from build args / env files.
- No new attack surface — the change only forwards existing process env vars to child processes.

LGTM. Ship it.

PR #67, opened 2026-04-12
perf(docker): use COPY --chown instead of recursive chown -R

rootbdfy:
## Summary

Replace `chown -R node:node /app /data /home/www` with `COPY --chown=node:node`
on all COPY directives in api, agent, and mcp stages.

## Problem

The `chown -R` command walks the entire `/app` directory tree (over 9k :) files in
node_modules) in a separate layer. On cold builds this single step took many time.

## Solution

Set file ownership at copy time via `--chown=node:node` flag on each `COPY` directive.
The remaining `chown node:node /data /home/www /app` (non-recursive, 3 directories only)
runs faster.

## Test plan

- [x] `docker compose build` completes successfully
- [x] All containers start and function correctly under `node` user
- [x] File permissions inside containers verified (`ls -la` shows `node:node` ownership)

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -24,14 +24,14 @@ FROM base AS api
 ENV NODE_ENV=production HOME=/home/node
 RUN apt-get update && apt-get install -y --no-install-recommends gosu git && rm -rf /var/lib/apt/lists/* \
     && npm i -g @anthropic-ai/claude-code @openai/codex
-COPY --from=deps /app/node_modules ./node_modules
-COPY --from=build /app/packages/shared ./packages/shared
-COPY --from=build /app/packages/data ./packages/data
-COPY --from=build /app/packages/runtime ./packages/runtime
-COPY --from=build /app/packages/api ./packages/api
-COPY package.json .
+COPY --chown=node:node --from=deps /app/node_modules ./node_modules
+COPY --chown=node:node --from=build /app/packages/shared ./packages/shared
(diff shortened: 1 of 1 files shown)


lee-to (commented):
## Code Review: perf(docker): use COPY --chown instead of recursive chown -R

### Verdict: LGTM ✓

Clean, focused optimization. The change is correct and consistent across all three stages (api, agent, mcp).

### What's good

- **Root cause addressed properly**: `chown -R` over `/app` (including 9k+ files in `node_modules`) was creating an expensive separate layer. Using `COPY --chown=node:node` sets ownership at copy time with zero extra I/O — this is the canonical Docker best practice.
- **Residual `chown` is intentional and correct**: The remaining `chown node:node /data /home/www /app` (non-recursive, 3 dirs only) handles directories created by `mkdir -p` that aren't covered by COPY.
- **Entrypoint left as-is — correct**: `docker-entrypoint.sh` isn't given `--chown` because it runs as root initially (gosu pattern), and the runtime `chown -R` in the entrypoint covers mounted volumes at startup, which is a different concern.
- **Consistent across all stages**: api, agent, mcp all follow the same pattern.

### Minor notes (non-blocking)

1. **Cache invalidation tradeoff**: Adding `--chown` to `COPY --from=deps` means that if the base image's `node` UID changes, the layer is still correct. No concern here — just noting the mechanism is sound.
2. **No behavioral change**: File ownership is identical at runtime. The only difference is build speed and layer size.

No issues found. Ship it.

PR #68, opened 2026-04-12
stabilize Codex runtime defaults and MCP port handling

ichinya:
## Summary

This PR hardens the remaining Codex runtime and MCP edge cases found during the stabilization audit.

### What changed

- added stable fallback behavior for invalid Codex `approvalPolicy` and `sandboxMode` overrides in both CLI and SDK adapters
- added DEBUG/WARN logging for resolved Codex permission settings and ignored invalid overrides
- normalized `MCP_PORT` parsing across the API route, local dev scripts, and the MCP package entrypoints
- kept MCP install behavior aligned by falling back to `stdio` when `MCP_PORT` is invalid instead of generating a broken HTTP config
- added regression coverage for:
  - invalid Codex permission overrides
  - invalid `MCP_PORT` values
  - partial `/settings/mcp/install` failures
- fixed the root coverage script so `npm run ai:validate` runs end-to-end in the monorepo
- updated runtime/MCP docs to match the current behavior

## Validation

- `npm run test --workspace=@aif/runtime -- src/__tests__/codexCli.test.ts src/__tests__/codexSdk.test.ts`
- `npm run test --workspace=@aif/mcp -- src/__tests__/env.test.ts`
- `npm run test --workspace=@aif/api -- src/__tests__/settings.test.ts`
- `npm run ai:validate`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ npm run init
 npm run dev
 ```
 
-Set `MCP_PORT` in your shell or root `.env` before `npm run dev` if you also want the MCP HTTP server in local development.
+Set `MCP_PORT` in your shell or root `.env` before `npm run dev` if you also want the MCP HTTP server in local development. Use an integer port between `1` and `65535`; invalid values are ignored by the dev launcher and the settings install route falls back to the local `stdio` entry instead of writing an HTTP MCP endpoint.
 
 ### With Docker
 
@@ -55,7 +55,7 @@ cd aif-handoff
 docker compose up --build
 ```
 
-Development starts three services by default. If `MCP_PORT` is set, it starts a fourth service for MCP over HTTP. Docker starts all four services.
+Development starts three services by default. If `MCP_PORT` is set to a valid integer port, it starts a fourth service for MCP over HTTP. Docker starts all four services.
 
 | Service   | URL                               | Description                                  |
 | --------- | --------------------------------- | -------------------------------------------- |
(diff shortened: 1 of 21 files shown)


lee-to (commented):
## Code Review

Overall this is a solid stabilization PR — good defensive hardening of Codex permission overrides and MCP port validation, with thorough regression tests. A few observations:

### Issues

**1. DRY violation: `resolveMcpPort` is copy-pasted 3 times**
The same validation logic exists in `scripts/dev.mjs`, `packages/api/src/routes/settings.ts`, and `packages/mcp/src/env.ts` (plus `packages/mcp/scripts/dev-http.mjs` with a slightly different variant). The "keep in sync" comments are a maintenance smell — consider extracting this into a shared utility. If the `.mjs` scripts can't import from `@aif/shared`, at minimum the two TypeScript copies (`settings.ts` and `env.ts`) should share a single implementation.

**2. Inconsistent validation approach between CLI and SDK adapters**
In `cli.ts`, approval/sandbox validation uses `normalizeCodexApprovalPolicy()` / `normalizeCodexSandboxMode()` — clean `Set`-based normalizers. In `sdk.ts`, the same validation is done with inline `=== "never" || === "on-request" || ...` comparisons repeated in both the warn helper call and the existing `if` block. This means the valid-values list is defined in two places within `sdk.ts` and diverges from the `Set`-based approach in `cli.ts`. Consider reusing the CLI normalizers (or extracting them to a shared Codex utils module).

**3. `warnOnInvalidCodexPermissionOverride` vs `warnOnInvalidCodexThreadOverride` — near-identical functions**
These two functions in `cli.ts` and `sdk.ts` do the same thing with a slightly different type signature (`source` field). Could be one shared function.

### Nits

- Log messages like `"WARN [runtime:codex] ..."` and `"DEBUG [runtime:codex] ..."` embed the log level in the message string, but pino already adds the level. The `[runtime:codex]` prefix is also redundant with the logger name `runtime:codex` that's already in the structured fields. Consider just using descriptive messages like `"Ignoring invalid Codex approvalPolicy override"`.

- `dev-http.mjs` throws on invalid `MCP_PORT` while `dev.mjs` silently returns `null`. Both are dev scripts — the inconsistency makes sense (one is explicitly HTTP-mode, one is optional), but worth a brief code comment explaining why.

### Positives

- Good test coverage: parameterized `it.each` for invalid port values, partial install failure test, CLI+SDK invalid override tests
- The transport-aware behavior in `mcp/src/env.ts` (warn+fallback in stdio, throw in http) is a nice design choice
- Docs are thorough and consistent with the code changes
- The `package.json` coverage script fix (`--workspaces --if-present` instead of turbo passthrough) is a good catch

LGTM with the DRY concern as the main actionable item for a follow-up.

ichinya:
Addressing the review feedback from [#pullrequestreview-4095234476](https://github.com/lee-to/aif-handoff/pull/68#pullrequestreview-4095234476).

Changes in `f6d755e`:
- extracted shared `parseMcpPortSetting()` into `@aif/shared` and reused it from the TypeScript MCP/API paths
- extracted shared Codex permission helpers into `packages/runtime/src/adapters/codex/permissions.ts` and reused them from both CLI and SDK adapters
- removed the redundant `WARN` / `DEBUG` prefixes from the new permission-related log message strings
- added explicit comments documenting why `scripts/dev.mjs` is lenient while `packages/mcp/scripts/dev-http.mjs` is strict

Re-ran targeted checks:
- `npm run test --workspace=@aif/runtime -- src/__tests__/codexCli.test.ts src/__tests__/codexSdk.test.ts`
- `npm run test --workspace=@aif/mcp -- src/__tests__/env.test.ts`
- `npm run test --workspace=@aif/api -- src/__tests__/settings.test.ts`

`npm run ai:validate` is still blocked by an existing unrelated web test failure:
- `@aif/web`
- `RuntimeProfileForm.test.tsx > shows model filter for large catalogs and filters case-insensitively`

So the review cleanup itself is covered

lee-to:
Thanks for the detailed breakdown! The changes sound like they address all the review points:

- `parseMcpPortSetting()` in `@aif/shared` → fixes the DRY concern ✓
- `permissions.ts` shared module for Codex CLI/SDK → fixes the inconsistent validation and duplicate warn helpers ✓
- Removed redundant `WARN`/`DEBUG` prefixes → nit fixed ✓
- Comments on dev script leniency difference → nit fixed ✓

However, I don't see commit `f6d755e` in the PR yet — the branch still shows only the original 2 commits. Could you push the fixup?

Re: the `@aif/web` test failure — as long as it's a pre-existing issue unrelated to this PR, that's fine. Might be worth a separate issue if one doesn't exist already.

ichinya:
For some reason, the push took an unexpectedly long time. The commit finally posted.

lee-to (approved):
All review feedback addressed cleanly in f6d755e:

- **DRY fix**: `parseMcpPortSetting()` extracted to `@aif/shared/mcpPort.ts` with a discriminated union return type (`unset | valid | invalid`) — nice API, both `settings.ts` and `mcp/env.ts` now consume it. The `.mjs` scripts keep their local copy with an explanatory comment — acceptable since they can't import from `@aif/shared`.

- **Shared permissions module**: `permissions.ts` consolidates `normalizeCodexApprovalPolicy`, `normalizeCodexSandboxMode`, and `warnOnInvalidCodexPermissionOverride` — both CLI and SDK import from it. The SDK now uses the `Set`-based normalizers instead of inline `===` chains, which also removed the duplicated valid-values lists.

- **Log message cleanup**: `WARN`/`DEBUG` prefixes removed from message strings, `[runtime:codex]` tag dropped — pino handles both.

- **Dev script comments**: `dev-http.mjs` now documents why it throws on invalid port vs `dev.mjs` silently skipping.

LGTM, good to merge.

PR #69, opened 2026-04-12
Fix/projects retry on api restart

rootbdfy:
## Summary

`docker compose restart` causes the web UI to lose project state and show an empty "Select project" screen with no way to recover without a manual page reload.

- **Problem:** `docker compose restart` makes the API briefly unavailable (10-15s). If the browser page is open or reloaded during that window, React Query's default 3 retries (~7s) exhaust before the API recovers. The `useProjects()` query permanently enters error state — the UI shows "Select project" with an empty board and never self-heals.
- **Root cause:** `useProjects()` used React Query defaults (`retry: 3`), which is insufficient for bootstrap data that the entire UI depends on.
- **Fix:** Set `retry: true` (infinite) with exponential backoff (2s → 4s → 8s → capped at 15s). The query keeps retrying until the API is reachable instead of giving up.

## Test plan

- [x] Playwright: stop API → reload page → wait 15s → start API → UI auto-recovers to selected project
- [x] Playwright: `docker compose restart` during page reload (race condition) → UI recovers after brief 502
- [x] Playwright: page stays open during restart → project selection persists
- [x] Full test suite passes (`npm run ai:validate`)

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,6 +1,17 @@
 # ---- Base ----
 FROM node:22-slim AS base
 WORKDIR /app
+# Proxy (optional) — propagated to all downstream stages via ENV,
+# so apt-get / npm / npx all pick it up during build.
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+ARG NO_PROXY
+ENV HTTP_PROXY=${HTTP_PROXY} \
+    HTTPS_PROXY=${HTTPS_PROXY} \
+    NO_PROXY=${NO_PROXY} \
+    http_proxy=${HTTP_PROXY} \
+    https_proxy=${HTTPS_PROXY} \
+    no_proxy=${NO_PROXY}
 
 # ---- Dependencies ----
 FROM base AS deps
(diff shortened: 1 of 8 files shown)


lee-to:
## Code Review

### Overall

The PR addresses three concerns: (1) infinite retry for the critical `useProjects` query, (2) HTTP proxy env-var forwarding in runtime adapters, and (3) Docker `COPY --chown` optimization. All three are reasonable changes. A few observations below.

---

### useProjects retry — looks good

The `retry: true` + exponential backoff (2s → 4s → 8s → cap 15s) is the right approach for bootstrap data. Comment is clear and explains the *why*. No issues here.

### Docker — `COPY --chown` and proxy support

- **`--chown` optimization:** Replacing `chown -R` with `COPY --chown` avoids a full recursive walk — good improvement for layer size and build time.
- **Proxy ARGs in Dockerfile:** The `ARG`/`ENV` block is correctly placed in both `base` and `web` stages (web uses Alpine, not base). Both case variants (`HTTP_PROXY` + `http_proxy`) are covered.
- **⚠️ Missing `build.args` in `docker-compose.yml`:** The Dockerfile declares `ARG HTTP_PROXY` etc., but `docker-compose.yml` doesn't pass them under `build.args`. Docker Compose does NOT auto-forward host env vars as build args — users will need to either:
  - Add `args:` under each service's `build:` block, or
  - Use `docker compose build --build-arg HTTP_PROXY=...` manually.

  Without this, the proxy ARGs will always be empty during `docker compose build`. Consider adding:
  ```yaml
  build:
    args:
      - HTTP_PROXY
      - HTTPS_PROXY
      - NO_PROXY
  ```
  (Compose resolves bare names from the host environment.)

### Proxy env allowlists — works, but repetitive

The same 6-entry proxy block is copy-pasted identically across 5 files (`claude/cli.ts`, `claude/options.ts`, `codex/cli.ts`, `codex/sdk.ts`, `codex/modelDiscovery/process.ts`). This is not a blocker, but worth noting as a future DRY candidate — a shared constant or helper would prevent them from drifting apart.

### OpenRouter adapter — no action needed

OpenRouter uses Node `fetch()` directly (no subprocess spawning), so proxy vars from the process environment are inherited automatically. No allowlist needed — correct as-is.

---

### Verdict

**Approve with one suggestion:** wire up the proxy build args in `docker-compose.yml` so `docker compose build` picks them up without manual `--build-arg` flags. The rest is clean.

rootbdfy:
Блин, хз как сюда все коммиты прилепились. Должен был только один - для решения с отвалом api. Сегодня первый раз в жизни PR сделал.

lee-to:
Спасибо за фикс! тут много из других пр, я пр закрою а фикс внесу сам с небольшими доработками

Issue #63, opened 2026-04-12
Add scheduled task execution and auto-queue processing mode


lee-to:
## Summary

Add the ability to schedule task execution at specific times and introduce an **auto-queue mode** that processes all queued tasks sequentially by their order number — one after another, without manual intervention.

## Motivation

Currently tasks must be triggered manually or picked up by the coordinator's polling loop. There is no way to:
- Schedule a specific task to run at a given time (e.g. "start planning at 09:00")
- Tell the system to drain the entire backlog automatically, processing tasks in order

This limits the system's autonomy and makes it harder to run unattended pipelines (e.g. overnight batch processing of a roadmap).

### 1. Scheduled Task Execution
- Allow setting a `scheduledAt` timestamp on a task (via API + UI)
- The coordinator checks for tasks whose `scheduledAt` has passed and triggers them automatically
- A task with a schedule should transition from its current stage into the next actionable stage when the time arrives

### 2. Auto-Queue Processing Mode
- A project-level or global toggle: **Auto Mode** (on/off)
- When enabled, the coordinator picks the next eligible task by `orderIndex` (ascending) and processes it through the full pipeline (Planning → Implementation → Review → Done)
- Only one task is active at a time — the next task starts only after the current one reaches `done` (or `failed`)
- Can operate at the **roadmap level**: generate a roadmap, then auto-process every resulting task in sequence
- The mode can be paused/resumed via API and UI

### API (`@aif/api`)
- [ ] Add `scheduledAt` field to task create/update endpoints
- [ ] Add `PATCH /api/projects/:id/auto-mode` endpoint (enable/disable)
- [ ] Add `GET /api/projects/:id/auto-mode` endpoint (current state)
- [ ] Broadcast auto-mode state changes via WebSocket

### Database (`@aif/shared`)
- [ ] Add `scheduled_at` column to tasks table (nullable datetime)
- [ ] Add `auto_mode` column to projects table (boolean, default false)

### Coordinator (`@aif/agent`)
- [ ] Add scheduled-task check to the polling loop (cron tick → find due tasks → trigger)
- [ ] Add auto-queue logic: when auto-mode is on, pick next task by `orderIndex` and drive it through stages
- [ ] Ensure sequential processing — only one task active per project in auto mode
- [ ] Emit events/logs for scheduled triggers and auto-queue transitions

### Frontend (`@aif/web`)
- [ ] Add "Schedule" option to task detail (date/time picker for `scheduledAt`)
[...]

PR #70, opened 2026-04-12
Feature/scheduled tasks auto queue

lee-to:
## Summary

Adds two related capabilities to the autonomous task pipeline (issue #63):

- **Scheduled execution** — tasks expose a `scheduledAt` ISO-8601 timestamp (normalized to UTC `Z` on intake). When the time arrives, the coordinator fires the task from `backlog` into `planning` (the same transition as the human `start_ai` event), clears `scheduledAt`, and broadcasts `task:scheduled_fired`.
- **Auto-queue mode** — a per-project toggle that fills the pipeline by **pool depth**. Sequential projects (`parallelEnabled: false`) advance strictly one task at a time — the next backlog task starts only after the previous reaches `done`/`verified`. Parallel-enabled projects fill the pipeline up to `COORDINATOR_MAX_CONCURRENT_TASKS` and refill as soon as a slot frees up. Both cases share one code path: `limit = parallelEnabled ? COORDINATOR_MAX_CONCURRENT_TASKS : 1`.

### Scope note vs. issue #63

The issue listed "parallel auto-mode execution" as out of scope. After review discussion in this PR, we extended the scope to allow auto-queue to also work with parallel-enabled projects (they share a single state-machine, atomic-claim primitive, and counter — splitting them would have been more code, not less). UI copy, docs, and tests reflect both modes. The strict-sequential semantic that #63 specifies is preserved exactly for non-parallel projects (the default).

Out of scope (per the issue): recurring cron schedules, cross-project orchestration.

## Changes

**Schema + migration (v12)**
- `tasks.scheduled_at` (TEXT nullable, ISO-8601 UTC) + `idx_tasks_scheduled_at` composite index on `(scheduled_at, status)`
- `projects.auto_queue_mode` (INTEGER boolean, default `0`)
- Named `autoQueueMode` to avoid collision with the existing per-task `tasks.auto_mode` AI-processing flag
- `Task.scheduledAt`, `Project.autoQueueMode`, optional `scheduledAt` on `CreateTaskInput`/`UpdateTaskInput`
- Three new `WsEventType` entries: `task:scheduled_fired`, `project:auto_queue_mode_changed`, `project:auto_queue_advanced`
- `scheduledAt: null` added to `CLEAN_STATE_RESET` in the state machine — any transition out of `backlog` (start_ai, scheduler, auto-queue, retry_from_blocked, etc.) clears the schedule in the same write

**Data layer (`@aif/data`)**
- `listDueScheduledTasks(nowIso)` — backlog + not paused + `scheduledAt <= nowIso`
- `claimBacklogTaskForAdvance(taskId)` — **CAS-style** conditional UPDATE that only succeeds when the row is still `backlog` AND `paused = false` at the moment of the write. Both scheduler and auto-queue use this primitive so the same row can never be advanced twice in a single cycle (or by parallel coordinator instances)
- `countActivePipelineTasksForProject(projectId)` — counts non-terminal (`planning`, `plan_ready`, `implementing`, `review`, `blocked_external`) tasks; powers pool-depth math
- `getAutoQueueMode` / `setAutoQueueMode` / `listAutoQueueProjects`
- `nextBacklogTaskByPosition(projectId)` — skips paused tasks and tasks with future `scheduledAt`
- `updateScheduledAt`, `clearScheduledAt`
- `updateTaskPositionOnly(id, pos)` — reorder writes ONLY position, never bumps `updatedAt`, so list views sorted by "Updated" stay stable when user clicks reorder arrows

**Coordinator (`@aif/agent`)**
- `processDueScheduledTasks()` — for each due backlog task, atomic-claims into `planning` via `claimBacklogTaskForAdvance`, appends `[scheduler]` activity log, broadcasts `task:scheduled_fired` and `task:moved`. Skips silently when the row was already claimed by another pass.
- `processAutoQueueAdvance()` — fill-the-pool loop: `limit = parallelEnabled ? COORDINATOR_MAX_CONCURRENT_TASKS : 1`. Same atomic-claim primitive; on a failed claim the active counter is re-read so pool depth stays accurate. Single tick fills the entire pool (a parallel project doesn't need three poll cycles to reach capacity). Broadcasts `project:auto_queue_advanced` per advance.
- Both passes run in `pollAndProcess` in the same cycle: scheduler fires its set, then auto-queue tops up the remaining pool slots (scheduler-fired tasks now count toward `active`).

**API**
- `createTaskSchema`/`updateTaskSchema` accept `scheduledAt`. Schema **normalizes to UTC `Z`** on intake (`z.transform(s => new Date(s).toISOString())`) so DB lexical compare matches instant compare. Past timestamps rejected with `400`; `null` clears.
- `GET`/`PATCH /projects/:id/auto-queue-mode` (broadcasts `project:auto_queue_mode_changed`)
- `POST /projects/:id/broadcast` for project-scoped WS events (used by coordinator)
- `PATCH /tasks/:id/position` rewritten to use `updateTaskPositionOnly` — does NOT bump `updatedAt` (reorder is metadata, not content)
- `notifyProjectBroadcast` helper in agent notifier

**Web**
- `TaskSettings`: datetime-local picker for `scheduledAt` (backlog only), priority Select picker (None..Critical, 0..5)
- `TaskCard`: "Starts ..." banner for backlog with `scheduledAt`, reorder ↑/↓ arrows, Pause/Resume button (paused = excluded from auto-queue and scheduler, gated on backlog status only)
- `TaskListTable`: "Order" column with reorder arrows + pause toggle for backlog rows; scheduled banner inline with title; auto-switches list sort to "By Status" when user clicks an arrow so the new position is actually visible
- `TaskDetailHeader`: scheduled banner under status/priority badges; pause button no longer hidden in backlog (now hidden only in `done`/`verified`)
- `Board`: Auto-Queue toggle moved into project create/edit dialog (`ProjectSelector`) under "Parallel Execution" — natural home for per-project setting; old board-header toggle removed
- `AddTaskForm`: priority Select picker, resets to None on submit success AND on X-close AND on form re-open
- `useAutoQueueMode` / `useSetAutoQueueMode` hooks; `useWebSocket` invalidates `["projects"]` and `["autoQueueMode", id]` on `project:auto_queue_mode_changed`
- All UI built by composing existing primitives (`Switch`, `Badge`, `Input`, `Select`, `Button`) — no new visual components, no Pencil sync needed

**Bug fixes uncovered during review**
- `scheduledAt` cleared on any transition out of `backlog` (was sticky after `start_ai`)
- `PATCH /tasks/:id/position` no longer bumps `updatedAt`
- `ProjectSelector` edit dialog: `previousAutoQueue` now looked up via `editingId`, not `selected` (was firing/skipping `setAutoQueueMode` against the wrong baseline when editing a non-selected project)
- `AddTaskForm` priority is now reset on dismiss/reopen flow, not just on successful submit

**Tests (+~70 new)**
- `@aif/data`: scheduler queries, auto-queue queries, CAS primitive idempotency + paused/non-backlog refusal + scheduledAt-clear-in-same-write, pool-depth count
- `@aif/agent`: dedicated `autoQueue.test.ts` with sequential semantics (advance after done, hold during planning/plan_ready/implementing/review, blocked_external counts, paused skip, scheduled-future skip), parallel pool-depth (fill in single tick, partial refill, capacity bail, backfill on done, paused exclusion), scheduler+auto-queue interaction (scheduler fires first then auto-queue tops up; paused due scheduled NOT fired), CAS race protection (single task that's both due and next-by-position advances exactly once), toggle path
- `@aif/api`: scheduledAt validation (future accept, past reject, non-ISO reject, **UTC normalization from offset values**, null clear), auto-queue-mode endpoints (GET/PATCH happy path, 400 invalid, 404 unknown), reorder-without-updatedAt-bump
- `@aif/web`: TaskSettings (schedule picker visibility, ISO conversion on save, Clear→null, priority preselect + change → onSave payload), AddTaskForm (priority default 0, chosen value flows to createTask, **reset on X-close+reopen**), TaskCard (scheduled banner, reorder callbacks + boundaries + non-bubble + non-backlog hide, pause/resume toggle), TaskListTable (Order column, midpoint reorder math, onReorderBacklog callback, pause toggle, banner inline, click bubbling), ProjectSelector (Auto-Queue Mode switch alongside Parallel)

**Docs**
- `docs/api.md`: auto-queue-mode endpoints, `scheduledAt` field, three new WS events
- `docs/architecture.md`: Scheduled Execution + Auto-Queue Mode (with pool-depth model and atomic-claim composition described explicitly)
- `packages/web/CHECKLIST.md`: new rule — every new WebSocket event must be wired through `useWebSocket.ts` (invalidate affected react-query keys; explicit handler for non-task scoped events). Never ship a WS event with no UI consumer.

## AI Model

- [x] Claude Opus 4.6

## Test Plan

- [x] Tests pass (`npm test` — full workspace green: data 145, api 220, agent 226, shared 141, runtime 594, web 549, mcp 81)
- [x] Lint passes (`npm run lint` — 0 errors, 0 warnings)
- [x] Coverage met per package (`npm run coverage`)
- [x] Documentation updated (api.md + architecture.md + CHECKLIST.md)
- [ ] Manually verified end-to-end: scheduled-task firing, sequential auto-queue strict-after-done, parallel pool-depth fill, paused exclusion, reorder arrows in both views, schedule + priority pickers, project dialog auto-queue toggle


ichinya:
Thanks for the PR — this looks like it covers most of the scope from #63: `scheduledAt`, project-level `autoQueueMode`, new API/WS events, scheduling UI, and auto-queue behavior.

I like the direction, but I think this still needs changes before merge. There are a few behavioral issues in the backend and one bug in the UI that look blocking to me:

1. `scheduledAt` currently accepts any offset-based datetime, but later it is compared as `TEXT` in the DB. That means non-UTC values can break due checks and ordering, because lexical string comparison is not equivalent to instant comparison unless the value is normalized first.
2. The sequential guarantee from #63 does not seem fully enforced yet. Auto-queue currently skips only when there is an active lock, but that is weaker than “wait until the current task is done or failed”.
3. The documented behavior says “run due scheduled task first; if none is due, auto-queue the next backlog item”, but the poll loop currently runs both steps unconditionally in the same cycle.
4. There is still a read-then-write race in scheduler/auto-queue flows because task selection and status update are not guarded by a conditional/CAS-style update.
5. There is a UI bug in `ProjectSelector`: when editing a non-selected project, submit compares against `selected?.autoQueueMode` instead of the project being edited.

Once those are addressed, this should be in much better shape.

lee-to:
Thanks — all five are valid blockers. Addressed in `87f0826`:

**1. UTC normalization for `scheduledAt`**
You're right — accepting any offset and storing the raw string broke `lte()` against `new Date().toISOString()` because lexical TEXT compare ≠ instant compare. The schema now normalizes on intake:

```ts
export const scheduledAtSchema = z
  .string()
  .datetime({ offset: true })
  .transform((s) => new Date(s).toISOString())   // → "...Z"
  .refine((iso) => Date.parse(iso) > Date.now(), { message: "must be future" })
  .nullable().optional();
```

`+03:00` values now land as `Z` and the scheduler scan matches reliably. New API test `normalizes non-UTC scheduledAt to UTC Z form` covers it.

**2. Sequential guarantee from #63**
This was already addressed by the pool-depth rewrite (`eca5ab7`) before your review, but it's worth restating since the diff is large. The condition is no longer "no active lock" — it's:

```ts
limit  = parallelEnabled ? COORDINATOR_MAX_CONCURRENT_TASKS : 1
active = count(tasks WHERE status IN
               ('planning','plan_ready','implementing','review','blocked_external'))
while active < limit: advance
```

For a sequential project (`limit=1`) the next backlog task can only fire after the previous one reaches a terminal status (`done`/`verified`). The inter-stage lock-gap that used to allow early advance is now closed because we count by status, not by lock window. Covered by the `holds while previous in planning/plan_ready/implementing/review` and `advances next task only after previous reaches done` tests in `autoQueue.test.ts`.

**3. "Run scheduler first; if none is due, auto-queue advances" wording**
Agreed — the docs were misleading. The pollAndProcess loop runs **both** unconditionally, and that's actually correct given the pool-depth model: the scheduler claims due tasks, then auto-queue runs and tops up the **remaining** slots (the scheduler-fired tasks now count toward `active`). I rewrote the architecture.md section:

> Auto-queue and scheduled execution compose in the same poll cycle: `processDueScheduledTasks()` runs first and fires every backlog task whose `scheduledAt` is due, then `processAutoQueueAdvance()` runs and **tops up the remaining pool slots**. Order matters because once the scheduler advances a task, it counts as in-flight and reduces how many slots auto-queue still needs to fill. Both passes use the same atomic `claimBacklogTaskForAdvance` write so a row is moved out of `backlog` exactly once even when both passes target the same task in the same cycle.

**4. Read-then-write race in scheduler/auto-queue**
Confirmed — `listDue…` then per-task `updateTaskStatus` was vulnerable to two passes (or two coordinator instances) racing the same row. Replaced with a CAS primitive in `@aif/data`:

```ts
export function claimBacklogTaskForAdvance(taskId: string): boolean {
  const result = getDb()
    .update(tasks)
    .set({ status: "planning", scheduledAt: null, /* CLEAN_STATE_RESET */, updatedAt: nowIso })
    .where(and(eq(tasks.id, taskId), eq(tasks.status, "backlog"), eq(tasks.paused, false)))
    .run();
  return result.changes > 0;
}
```

Both `processDueScheduledTasks` and `processAutoQueueAdvance` now wrap their per-task work in `if (!claimBacklogTaskForAdvance(id)) continue;`. Broadcasts, log entries, and counters only run when the claim succeeded. Auto-queue re-reads `active` after a failed claim so the pool counter stays accurate. Tests:
- `claimBacklogTaskForAdvance returns true exactly once and is idempotent on retries`
- `…refuses paused / non-backlog tasks`
- `…clears scheduledAt in the same write`
- agent-level `CAS race protection: scheduler and auto-queue cannot double-advance the same task in one cycle`

**5. `ProjectSelector` edit-mode bug with `selected?.autoQueueMode`**
Real bug — opening the edit dialog for a non-selected project compared the new toggle value against the wrong baseline, silently firing or skipping the `setAutoQueueMode` mutation. Fixed by looking up the project being edited via `editingId`:

```ts
const editingProject = projects?.find((p) => p.id === editingId);
const previousAutoQueue = editingProject?.autoQueueMode ?? false;
```

`ai:validate` passes (lint, all tests across packages, coverage, build). Ready for another look.

ichinya:
Nice work overall — the scheduledAt/data/API plumbing mostly looks coherent, and the PR is close. I still think this should stay in Request changes for two reasons.

First, the current auto-queue behavior no longer matches the contract in #63 or in the PR description. Both describe auto-queue as sequential only: advance the next backlog task when there is no active/locked task, with parallel auto-mode execution explicitly out of scope. The current coordinator implementation now fills the pipeline up to the parallel capacity for parallel-enabled projects, so a parallel project can auto-start multiple backlog tasks in a single poll cycle. That is a real product behavior change, not just an implementation detail.

I think we need to pick one direction and make it consistent everywhere:
- either keep auto-queue strictly sequential for all projects, as described in the issue/PR/UI copy
- or explicitly re-scope the feature and update the PR description, issue expectations, UI text, and tests to say that parallel projects may auto-advance multiple backlog tasks

Second, the latest priority-picker commit introduced a small but user-visible reset bug in AddTaskForm. `priority` is now part of the create payload and it is reset after a successful submit, but it is not reset in the normal dismiss/reopen flow: the sync-on-open effect does not restore it, and the X-close handler also skips resetting it.

Repro:
1. Open Add task
2. Set Priority = Critical
3. Close the form with X
4. Reopen the form

The previous priority is still selected.


Once the auto-queue contract is aligned and the AddTaskForm reset path is fixed, I think this will be in good shape.

lee-to:
Thanks for the second pass — both addressed in `4fc932d` plus the updated PR description.

**1. Auto-queue contract drift**

You're right that the implementation drifted from the strict-sequential wording in #63. We discussed two paths:

- **A)** Pull back to strict sequential everywhere, ignoring `parallelEnabled` for auto-queue
- **B)** Keep the pool-depth model but explicitly extend the scope, updating PR description / UI copy / docs / tests

Going with **B**. Reasoning: the two flags compose naturally through one primitive (`limit = parallelEnabled ? COORDINATOR_MAX_CONCURRENT_TASKS : 1`), splitting them would be more code and more state-machine surface, and the strict-sequential semantic from #63 is preserved exactly for non-parallel projects (the default). What changed:

- **PR description** rewritten with a "Scope note vs. issue #63" section that calls out the extension explicitly
- **UI copy** in `ProjectSelector` Auto-Queue Mode tooltip now says: *"Sequential projects start the next task only after the previous reaches done. Parallel-enabled projects fill the pipeline up to the parallel-execution cap."* — no surprises when a parallel project starts more than one task
- **Issue #63** comment: noting the extension and linking back here
- **Docs** (`architecture.md`) already describe the pool-depth model with the atomic-claim composition
- **Tests** cover both modes (sequential strict-after-done semantics in `autoQueue.test.ts > sequential project`, pool-depth synergy in `parallel project (synergy with parallelEnabled = true)`)

If you'd rather we go with A and ship a separate follow-up issue/PR for the parallel synergy, happy to do that — just say the word.

**2. AddTaskForm priority reset on dismiss/reopen**

Confirmed and fixed. `priority` was reset on submit-success but missed in two paths: the X-close handler and the `syncGen`-driven open effect. Now reset in both — same lifecycle as `title`, `description`, etc. New test `resets priority back to None when dismissed via X and reopened` covers the repro you described.

`ai:validate` green. Ready for another look.

ichinya (approved):

PR #71, opened 2026-04-13
feat(tasks): mode-driven defaults for review/docs/tests + scoped review sidecars

lee-to:
## Summary
- Adds a shared `defaultsForMode()` helper so planner mode drives `skipReview` / `planDocs` / `planTests` consistently across UI (AddTaskForm, TaskSettings) and API (`POST /tasks`, roadmap import).
- Full mode → `{skipReview:false, planDocs:true, planTests:true}`; Fast mode → `{skipReview:true, planDocs:false, planTests:false}`; roadmap-generated tasks force `skipReview:true`.
- Centralizes the review-stage scope rule in `subagentQuery.ts`: when `profileMode === "review"` the review/security sidecars automatically get a system-prompt append restricting them to the current task's diff. No longer depends on editing each `.claude/agents/*.md` file.
- Fixes a latent AddTaskForm reset bug (reset set `plannerMode` to `full` while initial was `fast`).

## Test plan
- [x] `npm run ai:validate` passes (lint, build, tests, coverage)
- [x] New assertions: `POST /tasks` default matrix (fast/full/explicit-false), roadmap import `skipReview=true`
- [x] AddTaskForm: default fast flags, Full/Fast flip, reset-on-success returns to fast defaults
- [x] TaskSettings: mount preserves saved values, manual Full radio flips flags, PATCH payload carries them
- [x] Manual smoke check in UI (Kanban + task detail)

### DIFF
--- a/packages/agent/src/constants.ts
+++ b/packages/agent/src/constants.ts
@@ -6,3 +6,16 @@ export const PROJECT_SCOPE_SYSTEM_APPEND =
   "Project scope rule: work strictly inside the current working directory (project root). " +
   "Do not inspect or modify files in the orchestrator monorepo or in parent/sibling directories " +
   "unless the user explicitly asks for that path. Avoid broad discovery outside the current project root.";
+
+/**
+ * Scope rule injected automatically into review-style subagent queries
+ * (profileMode === "review"). Keeps review/security sidecars focused on the
+ * current task's diff instead of auditing unrelated code paths.
+ */
+export const REVIEW_DIFF_SCOPE_SYSTEM_APPEND =
+  "Review scope rule: review ONLY code that changed as part of this task's implementation " +
+  "(the diff introduced by the current plan's tasks). Do NOT audit unrelated files, " +
+  "pre-existing code paths, or broader project concerns. If a concern is outside the changed " +
+  'scope, note it briefly as "out of scope" and move on. Reference changed files/lines ' +
+  "explicitly. Ignore pre-existing issues unless they are directly aggravated by the change. " +
+  "Your job is to validate the delta, not the whole codebase.";
--- a/packages/agent/src/subagentQuery.ts
+++ b/packages/agent/src/subagentQuery.ts
@@ -29,7 +29,7 @@ import {
 } from "@aif/runtime";
 import { getEnv, logger } from "@aif/shared";
 import { logActivity } from "./hooks.js";
-import { PROJECT_SCOPE_SYSTEM_APPEND } from "./constants.js";
+import { PROJECT_SCOPE_SYSTEM_APPEND, REVIEW_DIFF_SCOPE_SYSTEM_APPEND } from "./constants.js";
 import { createStderrCollector } from "./stderrCollector.js";
 import { writeQueryAudit } from "./queryAudit.js";
 import { getActiveStageAbortController } from "./stageAbort.js";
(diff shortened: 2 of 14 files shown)


ichinya:
Overall this is a nice cleanup. I like that defaultsForMode() is centralized and now applied consistently between the create UI and POST /tasks, and the added tests cover the main create/detail/API paths.

I do see one blocking issue though: roadmap import still bypasses POST /tasks and calls createTask() directly. That means the new mode-driven normalization is not actually shared across all task-creation entry points. The comment in roadmapGeneration.ts says planner mode is left untouched and parallel-enabled projects already force full via POST /tasks, but this flow never goes through POST /tasks. As written, I don't think imported tasks are guaranteed to inherit the same plannerMode / planDocs / planTests defaults as normal task creation, especially for parallel-enabled projects.

I’d fix this by moving the normalization into a shared create helper or into createTask() itself, or by explicitly computing plannerMode + defaults inside importGeneratedTasks(). I’d also add a test asserting that roadmap-imported tasks in a parallel-enabled project get full-mode settings while still forcing skipReview=true.

Non-blocking follow-up: PUT /tasks/:id still looks API-asymmetric with POST — changing plannerMode without explicit flags won’t re-resolve defaults. The current UI appears to compensate by sending all three flags on mode switch, so I’d treat that as follow-up rather than blocking this PR.

lee-to:
Thanks for the review — blocker fixed in 45699ad.

**What changed:**
- `importGeneratedTasks` no longer relies on the "parallelEnabled projects already force full via POST /tasks" assumption — the import path bypasses POST, so normalization is now applied inline here: `plannerMode = project.parallelEnabled ? "full" : "fast"`, then `defaultsForMode(plannerMode)` fills `planDocs`/`planTests`, and `skipReview: true` is always forced.
- Added a test for parallel-enabled projects: imported tasks get `plannerMode=full`, `planDocs=true`, `planTests=true`, `skipReview=true`.
- Extended the existing test for regular projects: `plannerMode=fast`, `planDocs=false`, `planTests=false`, `skipReview=true`.

I didn't pull the normalization into a shared helper / into `createTask()` in `@aif/data` this round: there are only two entry points right now (POST /tasks and `importGeneratedTasks`), and both pass `plannerMode` explicitly. If a third one shows up I'll move it into a shared helper to avoid duplicating the rule.

**On the non-blocking PUT /tasks/:id point:** agreed, leaving it as follow-up. The UI currently compensates by sending all three flags on a mode switch, so the asymmetry isn't user-visible in practice, but the API behavior should be aligned in a separate PR.

ichinya:
LGTM overall.

My previous blocker around roadmap import is resolved by the follow-up commit: importGeneratedTasks() now applies the mode-driven defaults explicitly when it bypasses POST /tasks, while still forcing skipReview=true for roadmap imports.

I like the shape of this change: defaultsForMode() is centralized in shared, create-time API defaults now preserve explicit false values, the web create/settings flows use the same helper, and the review-diff scoping is injected centrally for review-mode subagent queries.

Non-blocking follow-up: PUT /tasks/:id is still slightly API-asymmetric with POST /tasks. If an external client sends only plannerMode, the route does not currently re-resolve skipReview/planDocs/planTests from mode defaults. The current UI compensates by sending those flags explicitly on mode switch, so I would not block this PR on that.

PR #77, opened 2026-04-14
fix(chat): render AskUserQuestion prompts in chat UI (#74)

### LINKED ISSUE #74: [discussion] Question tool and permissions in chat mode. Runtime: Claude Code CLI.
In chat mode, when invoking /aif, /aif-improve, and so on, Claude sometimes wants to ask something via the Question tool. Naturally, that message does not come through in chat, which results in an unexplained pause. If you then ask Claude again, it will reply that it did not receive an answer and will make the decision for you. It may make sense to disable the Question tool in ai-factory with a special flag, so that questions are asked as plain text with selectable options instead. Also, in chat mode, it was not possible to create mcp.json during project bootstrap via /aif. Claude treated it as a sensitive file and requested permission. Naturally, that permission request did not come through in chat either. This appears to be designed for direct use in Claude Code. This needs some thought as to how best to handle it.

lee-to:
## Summary

- Adds a runtime-neutral `tool:question` event + `RuntimeToolQuestionPayload` shape in `@aif/runtime`, with a shared `buildToolUseEvents()` helper in `toolEvents.ts` so every adapter emits tool activity in the same way.
- Claude adapter (both SDK streaming and CLI stream-json) parses `AskUserQuestion` into the normalized payload and emits `tool:question` alongside `tool:use`. Parser lives in `adapters/claude/questions.ts` and is the Claude-specific part; new adapters implement an equivalent `parseXxxQuestion()` and feed the result into the shared builder — documented in `adapters/TEMPLATE.ts`.
- Chat route renders `tool:question` as a markdown block (header, question, numbered options, answer hint) via the existing `chat:token` channel. Other non-noisy tool calls (`Bash`, `Write`, `Edit`, etc.) now surface as `> 🔧 ToolName` system lines so users see progress instead of a silent pause.
- System-prompt hint appended to chat so Claude keeps calling `AskUserQuestion` and understands that the user's next message is the answer.
- Docs note in `docs/configuration.md` covers `.mcp.json` trust (separate from `AGENT_BYPASS_PERMISSIONS`) and the answer-by-text flow.

Closes #74.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run build` — clean
- [x] `npm run test` — 225+603+233 pass across agent/runtime/api
- [x] `npm run format`
- [x] `npm run ai:validate`
- [x] Manual verification on CLI transport: Claude called `AskUserQuestion` with a header, a question, and 4 options — all rendered in the chat UI with the numbered list and answer hint, and the user's reply advanced the flow.
- SDK transport covered by unit test with a mocked `@anthropic-ai/claude-agent-sdk` stream emitting an `assistant` message containing a `tool_use` block.

## Notes

- Other adapters (Codex, OpenRouter) don't have an equivalent interactive tool today; they gain support for free the moment they add a parser + call into the shared `buildToolUseEvents`. No code duplicated.
- `.mcp.json` permission prompts remain Claude Code's own trust layer — not solvable from our process; documented as a caveat.


ichinya:
**Request changes**

Thanks, good PR. Direction is correct: the “silent pause” issue is solved not by a local hack in chat, but through runtime-neutral `tool:question`; logic is moved to the general `buildToolUseEvents()`, Claude-paths (SDK and CLI) are reduced to one scheme, and chat finally shows both the question and visible progress on non-noisy tool calls. Plus it’s well covered by tests and synchronized with the documentation. 

But I wouldn't approve it yet because of one functional gap and a couple of architectural comments.

### 1. Blocker: `AskUserQuestion` doesn't survive reload / reopening of the session

In `packages/api/src/routes/chat.ts`, `tool:question` currently only goes to the live UI via `sendToken(rendered)` but doesn't reach `fullAssistantResponse`. At the same time, both the `assistantMessage` in the HTTP response and the local DB history are built from the `fullAssistantResponse` / `result.outputText`. This means that if the agent's progress consists only of `AskUserQuestion`, the question will only be visible to the current websocket client; according to the current code, it will not be saved as an assistant message after refresh or reopen. Moreover, loading the history from the runtime session only collects messages from the assistant/user text payloads, while `tool:question` is a separate event without a role, so it won't be possible to restore this prompt from the outside either. I would either add persistence for the rendered question block or explicitly materialize it as an assistant message in the history. Additionally, a regression test is required at the level of `POST /chat` → `GET /chat/sessions/:id/messages` → the question is present in the history.

### 2. The UI loses the semantics of `multiSelect` and multiple questions

The parser and types already save `multiSelect`, and the payload allows multiple questions in a single tool call. However, `formatToolQuestion()` ignores this and always displays the same hint: “Answer by number or free text in the next message.” For `multiSelect: true`, this is misleading, and it is unclear how to answer multiple questions in a row. I would either render the hint conditionally (`select one` / `you can select multiple, list the numbers separated by commas` / `answer each question in order`) or not include `multiSelect` in the public payload until the UI can display it correctly. A separate test for `multiSelect: true` is very necessary here.

### 3. Claude-specific prompt leak into the general chat path

`CHAT_ASKUSERQUESTION_HINT` is now definitely added to `buildContextAppend()`, which means it is added to any chat runtime/profile. However, the PR clearly states that the interactive question tool is currently only implemented for Claude, and that other adapters will be supported later when their own parser is added. This means that the overall chat scaffold is now implicitly dependent on Claude-specific features. This can lead to unnecessary prompt noise for other runtimes, and in the worst case, it can create false expectations about the available tools. I would do it conditionally: by runtime/provider, by capability-flag, or in general through adapter-specific system append, and not through general `buildContextAppend()`.

### 4. Tests cover happy path, but not the most risky UX-scenarios

Now there are good tests on emission (`tool:use` + `tool:question`), suppression noisy tools, dedupe by `toolUseId` and parser legacy shape. But there is not enough coverage on:

* persistence/reload history after question-only turn,
* `multiSelect`,
* multiple questions in one payload,
* a case without `toolUseId`, where chat-layer dedupe will no longer work.

### What I liked

Architecturally, the strongest point of the PR is the normalization of the event at the runtime level, rather than the chat-only special case. This ensures that the Claude CLI and Claude SDK do not duplicate logic, and future adapters can indeed connect through parser + `buildToolUseEvents()`. The documentation has also been updated in the same direction as the code, and the suppression of noisy tools really improves the chat's UX. 

---

And three more short inline-comments if you need the "on-the-fly" format:

**`packages/api/src/routes/chat.ts`**

> `tool:question` is only rendered in the websocket (`sendToken(rendered)`), but it is not included in the `fullAssistantResponse`, which is later used to construct the `assistantMessage` and DB history. As a result, AskUserQuestion-only turn doesn't survive reload/reopen. We need persistence for the question prompt and regression test on `/chat/sessions/:id/messages`. commit f8445601fdde0beb99476f111cf1a92fc94ad6ce

**`packages/api/src/routes/chat.ts`**

> Here, the UI loses the `multiSelect`/multiple-questions semantics: the payload carries them, but the renderer always shows the same single-choice hint. This can easily become a misleading UX once Claude actually sends a multi-select question. 
https://raw.githubusercontent.com/lee-to/aif-handoff/f8445601fdde0beb99476f111cf1a92fc94ad6ce/packages/runtime/src/adapters/claude/questions.ts

**`packages/api/src/routes/chat.ts`**

> `CHAT_ASKUSERQUESTION_HINT` is now added globally for any chat runtime, although only Claude has support in PR. It's better to limit this to a capability/runtime-specific branch, so as not to drag Claude-specific prompt semantics into the general chat scaffold. https://raw.githubusercontent.com/lee-to/aif-handoff/f8445601fdde0beb99476f111cf1a92fc94ad6ce/packages/api/src/routes/chat.ts

lee-to:
Thanks for the review — all four points addressed in a7a6980.

### 1. Blocker — persistence of AskUserQuestion-only turns
`packages/api/src/routes/chat.ts` now appends the rendered `tool:question` block to `fullAssistantResponse` (and flips `hasStreamedTokens = true`) so the same text lands in `createChatMessage({ role: "assistant" })`, the HTTP `assistantMessage` field, and the WebSocket stream. A question-only turn now survives reload and reopen.

Covered by new test: `persists rendered AskUserQuestion block as an assistant message so it survives reload` — asserts both `body.assistantMessage` and the persisted DB row contain the rendered question + options.

### 2. multiSelect / multiple-question UX
`formatToolQuestion()` now branches on `payload.questions.length` and `question.multiSelect`:
- single + single-select → `_Answer by number or free text in the next message._` (unchanged)
- single + multiSelect → `_You can select multiple options — list the numbers separated by commas, or answer in free text._`
- multiple questions → per-block `Select one.` / `Select one or more (N options).` headers + a global `_Answer each question in order…_` hint under the `---`-separated list.

Covered by: `renders a multi-select hint when the question has multiSelect=true` and `renders per-question selection hints and an ordered-answer hint when payload has multiple questions`.

### 3. Claude-specific prompt leak
New optional `RuntimeCapabilities.supportsInteractiveQuestions` flag. Claude SDK and CLI set it to `true`; Claude API, Codex, OpenCode, OpenRouter, and `DEFAULT_RUNTIME_CAPABILITIES` leave it unset (→ `false`). In `POST /chat` the adapter is resolved first, and `CHAT_ASKUSERQUESTION_HINT` is injected into `systemPromptAppend` only when `resolveAdapterCapabilities(adapter, transport).supportsInteractiveQuestions === true`. `buildContextAppend()` now takes `{ interactiveQuestions }` so session-discovery paths never carry the hint.

Docs updated in `docs/providers.md` → new "Interactive questions capability" section; `TEMPLATE.ts` documents the flag for future adapters.

Covered by: `includes AskUserQuestion hint in systemPromptAppend only for runtimes that support interactive questions` — asserts both branches (flag off → absent, flag on → present).

### 4. Missing test coverage
Added:
- persistence after a question-only turn (→ `createChatMessage` role=assistant, HTTP `assistantMessage`),
- `multiSelect: true`,
- multiple questions in one payload,
- `toolUseId: null` (chat-layer dedupe by id cannot suppress → still renders),
- capability-gated system-append injection.

All 562 tests pass (`npm run ai:validate` green).

ichinya:
Yes, the previous comments from the old review have already been closed: the question is now saved as an assistant message for a question-only turn, the UI can handle `multiSelect`/multiple questions, and `CHAT_ASKUSERQUESTION_HINT` is tied to the capability flag `supportsInteractiveQuestions`. 

From the new review, I would like to leave the following comments.

1. **The loss of assistant text on Claude CLI if there is both text and `AskUserQuestion` in the same assistant message.**
   In the chat route, any `tool:question` now sets `hasStreamedTokens = true` and adds only the rendered question block to `fullAssistantResponse`; fallback `result.outputText` is sent and saved only when `!hasStreamedTokens`. In Claude CLI, text `assistant`-blocks in partial-messages mode are not emitted as `stream:text`, but are accumulated in `assistantText` and only appear in `outputText` on success. So the scenario `[{type:"text", ...}, {type:"tool_use", name:"AskUserQuestion", ...}]` is now losing the text introduction: the user will only see the question, and the phrase before it will not be displayed in the UI or in the DB. The current regression test does not catch this because it only checks the question-only turn with `outputText: ""`. I would add a separate test for mixed assistant text + tool_use for the CLI path. [chat.ts](https://raw.githubusercontent.com/lee-to/aif-handoff/fix/chat-askuserquestion-74/packages/api/src/routes/chat.ts)

2. **The story is still not fully runtime-neutral: double render for future interactive tools that are not named `AskUserQuestion`.**
   By agreement, `buildToolUseEvents()` always emits `tool:use`, and for interactive questions, it also emits `tool:question`; in PR and docks, this is presented as a general scheme that new adapters can connect to “almost for free”. However, the chat route suppresses `tool:use` only when the hardcoded `toolName === "AskUserQuestion"`. For any subsequent runtime with a different interactive tool name, the UI will display both the `> ToolName` and the question block. I wouldn't solve this by the name of the tool: it's better to set an explicit flag like `interactiveQuestion: true` in `tool:use` directly in `buildToolUseEvents()`, or otherwise correlate `tool:use` and `tool:question` at the event payload level, rather than using a Claude-specific string compare. [toolEvents.ts](https://github.com/lee-to/aif-handoff/blob/fix/chat-askuserquestion-74/packages/runtime/src/toolEvents.ts)

3. **Virtual runtime sessions still lose `tool:question` when re-reading the history.**
   The fix with `fullAssistantResponse += rendered` helps regular DB-backed chat sessions, but in the `GET /chat/sessions/:id/messages` branch for virtual sessions, messages are still only built from events with `role` and `event.message`. `toolQuestionEvent()` has a `message` but no `role`, so such questions are still filtered out. That is, for “clean” runtime sessions, the history of interactive questions after reopen is still incomplete. Here, you need either the same materialization logic in the read-path, or a separate projection of `tool:question` in the assistant message when reading session events. [chat.ts](https://github.com/lee-to/aif-handoff/blob/fix/chat-askuserquestion-74/packages/api/src/routes/chat.ts)

lee-to:
Thanks for the follow-up review — all three points addressed in 7e87557.

### 1. Mixed assistant text + `AskUserQuestion` on Claude CLI
`packages/api/src/routes/chat.ts` now tracks `streamedTextLength` separately from `hasStreamedTokens`. `tool:question` still flips `hasStreamedTokens` (so the `result.outputText` fallback doesn't overwrite a question-only turn), but the intro-text fallback is keyed on `streamedTextLength === 0`. When Claude CLI's partial-messages path keeps the text assistant-block in `result.outputText` without re-emitting it as `stream:text` (while `tool_use` still fires `tool:question`), the intro text is now prepended to `fullAssistantResponse` and streamed via `sendToken()` before the rendered question. UI and DB row both see `"Let me check the options." + question block`, in that order.

Covered by two new tests:
- `prepends assistant outputText when a mixed text+AskUserQuestion turn streamed only the question (Claude CLI path)` — asserts ordering (intro before question) in `assistantMessage` and the persisted assistant row.
- `does NOT duplicate assistant text when stream:text deltas already fired alongside a question` — guards against double emission when deltas + `outputText` both carry the same text.

### 2. Runtime-neutral interactive-tool dedup
`buildToolUseEvents()` in `packages/runtime/src/toolEvents.ts` now tags every `tool:use` event with `interactive: Boolean(questionPayload)` in `data`. `chat.ts` suppresses the rendered `> Tool` line based on `data.interactive === true` instead of a hardcoded `toolName === "AskUserQuestion"` check. A future adapter with any differently-named interactive tool will correlate its `tool:use` and `tool:question` automatically via the flag — no Claude-specific string compare.

Covered by:
- Updated `suppresses noisy tool:use…` test — now also emits a `tool:use` with `name: "SomeOtherAdapterQuestion", interactive: true` and asserts it is suppressed (proving the dedup is name-independent).
- Updated `buildToolUseEvents` unit tests — check `interactive: false` for plain tools and `interactive: true` when a payload is passed.

### 3. `tool:question` projection in virtual-session read-path
Both read-paths in `GET /chat/sessions/:id/messages` now project `tool:question` events through `formatToolQuestion()` as assistant messages:
- the virtual-session branch (`parseVirtualRuntimeSessionId` → `adapter.listSessionEvents`),
- the DB-backed branch that merges runtime session events with local DB history.

After reload/reopen of a clean runtime session, previously asked questions now appear in chat history with the same markdown rendering as live-streamed ones.

Covered by: `projects tool:question events as assistant messages when reading virtual runtime session history` — asserts the returned message has `role: "assistant"` and contains the rendered question + numbered options.

All 2004 tests pass (`npm run ai:validate` green).

ichinya:
I've rechecked the current head PR. The old points are indeed closed: `tool:question` is now added to `fullAssistantResponse`, `formatToolQuestion()` already takes into account `multiSelect` and multiple questions, and `CHAT_ASKUSERQUESTION_HINT` is only enabled when the runtime declares `supportsInteractiveQuestions`; the Claude SDK/CLI has this flag set, which is `false` by default. This can be seen in the code and the new tests.

Here are some more notes:

1. `GET /chat/sessions/:id/messages` still loses `tool:question` for runtime-only / virtual sessions. Currently, `eventRole()` only accepts `event.data.role`, and both runtime history reading branches filter events without `role`. However, `toolQuestionEvent()` creates `tool:question` without `role`, and `POST /chat` already has a comment stating that replay runtime-history does not include such events with a role. That is, the fix with `fullAssistantResponse += rendered` saves a regular DB-backed chat session, but not a clean runtime-session history. I would fix this not only through persistence in `/chat`, but also through the replay-path: either materialize `tool:question` as an assistant message when reading runtime events, or add `role: "assistant"`/a special projection at the runtime event level. And we need a regression test specifically on `/chat/sessions/:id/messages` for a virtual session with `tool:question`. 

2. Suppression of double-render is still tied to the hardcode `toolName === "AskUserQuestion"`. At the same time, the contract of `buildToolUseEvents()` explicitly states that the interactive question flow is a general pattern of `tool:use` + `tool:question`, and TEMPLATE recommends that future adapters simply call the shared helper. The PR description even states that other adapters “gain support for free”. With the current code, this is not entirely true: any next provider-specific interactive tool is rendered twice — as `> ToolName` and as a question block. I would not match by the name of the tool, but would throw an explicit sign in `tool:use` like `interactiveQuestion: true` or correlate `tool:use` / `tool:question` by `toolUseId'. 

3. A small design note to the first point: `toolQuestionEvent.message` currently stores only `payload.questions[0]?.question || payload.toolName`. That is, if you then fix replay/history by simply reading `event.message`, the header, options, and structure of multiple questions will still be lost. Therefore, it is better to restore the text from `RuntimeToolQuestionPayload` for history-path rather than relying on `event.message`.

lee-to:
Addressed the latest review (99793f9):

**1. `tool:question` lost on virtual-session replay — fixed at the adapter layer.**
`listClaudeRuntimeSessionEvents` (`packages/runtime/src/adapters/claude/sessions.ts`) now walks assistant `tool_use` blocks and projects every `AskUserQuestion` into a `tool:question` event, reusing `parseClaudeAskUserQuestion` + `toolQuestionEvent`. A question-only turn and a mixed text+question turn both survive reload of a clean runtime session (`sdk:` / `runtime:` ids) through `GET /chat/sessions/:id/messages`. Non-question tool_use blocks stay out of chat replay — scope matches the review intent.

**2. Double-render suppression is by flag, not tool name.**
`chat.ts` already branches on `data.interactive === true` set by `buildToolUseEvents` when a `questionPayload` is provided. Any future adapter that wires its own interactive tool through the shared helper gets free suppression — no name match anywhere. Regression test in `chat.test.ts` (`SomeOtherAdapterQuestion`) guards against re-introducing a name check.

**3. `toolQuestionEvent.message` — defensive upgrade.**
Now joins every question text with `|` so a naive consumer reading `event.message` sees a useful summary instead of only the first question. History rendering in both chat read-paths keeps using `event.data` (`RuntimeToolQuestionPayload`), so header/options/multi-question structure is never read from `event.message`.

**Tests added**
- `claudeAdapter.test.ts` — `tool:question` projection for question-only, mixed text+question, and Bash (non-question) blocks in session history; confirms no `tool:use` leakage in replay.
- `toolEvents.test.ts` — joined-message contract + `event.data` payload identity.

**Checks:** lint, build, format, full test suite all clean locally (`npm run lint`, `npm run build`, `npm run test`, `npm run ai:validate`).

ichinya:
Yes, I've double-checked the current head PR. The main past issues have already been resolved: `tool:question` is now materialized in history-read paths, suppression of `tool:use` for interactive tools is based on the `interactive` flag, and the renderer and tests already cover `multiSelect`, multiple questions, and the `supportsInteractiveQuestions` capability flag.

However, I would still like to address the following issues:

1. **In mixed text + `AskUserQuestion` live websocket, the order is still incorrect.**
   Currently, `tool:question` is still sent to the stream immediately via `sendToken(rendered)`. After the run is completed, the missing intro from `result.outputText` is indeed prepended to `fullAssistantResponse`, but it is sent to the websocket as a separate late `sendToken(result.outputText)`. This results in an HTTP/persisted response of `intro + question`, and live `chat:token` remains `question -> intro` in the code. The current regression test for this case only checks the order in `body.assistantMessage`, not the order of the `chat:token` themselves. I would either buffer the question-block until the end of the assistant turn or at least add an explicit token-order test.

2. **`GET /chat/sessions/:id/messages` can still duplicate a mixed turn after reload of a linked DB session.**
   Claude replay already splits such an assistant turn into a regular `session-message` with text and a separate `tool:question`. Read-path in `chat.ts` materializes both as messages and then merges runtime/DB history only by exact trimmed-content equality. But the POST-path for the same scenario saves one combined assistant message: first it does `fullAssistantResponse = result.outputText + fullAssistantResponse`, then it writes this final text to the DB. As a result, the combined DB row will not match either the runtime text or the runtime question, and merge will add it as the third message. We need a regression test specifically for reload linked session after a mixed turn.

3. **Non-blocker: the materialized `tool:question` does not have a stable `id`.**
   `toolQuestionEvent()` puts in `data` just payload without `id`, and `eventId()` in chat route takes `event.data.id ?? crypto.randomUUID()`. Since read-path for `tool:question` uses `id: eventId(event)`, the same question from runtime history will get new id on every fetch/reload. For front, this usually means extra churn of message list. I would use `toolUseId` as stable id for `tool:question`.

lee-to:
Addressed all three review items in `c04833a`:

**1. Live websocket token order (mixed text + AskUserQuestion)**
Rendered question blocks are now buffered in `pendingQuestionBlocks` during the run and flushed only after the intro text is sent (either live via `stream:text` or via the post-run `outputText` prepend). The live `chat:token` stream matches the final HTTP/DB order (`intro → question`). Added an explicit token-order regression test in `chat.test.ts`.

**2. Duplicate mixed turn on reload of linked DB session**
Mixed turns are persisted as two separate assistant rows (intro text + rendered question block), mirroring Claude replay's split of the assistant turn into `session-message` + `tool:question`. `mergeRuntimeAndDbMessages` now finds exact trimmed-content matches for both rows instead of leaving a stale combined row as an unmatched third message. Added a reload-dedup regression test in `chatSessions.test.ts` that exercises the full split-persist + split-replay path.

**3. Stable `id` for materialized `tool:question`**
`eventId()` now falls back to `tool:question:<toolUseId>` when `data.id` is absent, so the same question keeps a stable client id across fetches/reloads. Added a regression test that asserts the id is identical across repeated `GET /chat/sessions/:id/messages` calls.

All checks green: lint, build, tests (1398 pass), format, `ai:validate`.

ichinya:
Notes:

1. In recovery-path, there is still a narrow edge case with partially received assistant text. In `chat.ts`, the missing text is retrieved from `result.outputText` only when `streamedTextLength === 0`, and the result is then collected as `streamedText + questionBlocksText`. However, in Claude CLI partial-messages mode, the text from assistant blocks is accumulated in `assistantText`, text deltas are separately written in `outputText`, and in the end, `assistantText` is substituted only if `outputText` is empty. This leads to an unpleasant intermediate scenario: if some of the text comes in delta-s, and another text-block is left only in the complete assistant block, it can be lost both in the live UI and in the persisted response. I would either merge `result.outputText` with streamed text by missing suffix/prefix, or add a regression test specifically for partial-delta + hidden text block. 

2. The chat layer still does not preserve the exact order of events within an assistant turn. Currently, each `tool:question` simply goes to `pendingQuestionBlocks` and is concatenated into the overall `questionBlocksText`; after the run is completed, all question blocks are sent to the end, and a single assistant row is written to the database with the entire `trimmedQuestionBlocks`. However, the read-path for runtime history materializes each `tool:question` as a separate assistant message. For the current Claude happy path, this is probably fine, but as soon as there is more than one interactive question event in a turn, or any interleaving of the form `text → question → text`, the shape live/DB will no longer match the runtime replay, and dedupe will become fragile again after reload. I would either keep ordered segments or at least a separate DB row for each question event, rather than a single shared block for the entire turn.

3. Dedupe by `toolUseId` currently only remembers the last id. In the types, `toolUseId` is explicitly described as a key for de-duping re-emits, but the chat route only compares the event with `lastToolPromptId`. This catches consecutive duplicates, but it does not catch a repeat of the same question later in the same run after another event. It is cheaper and more reliable to keep `seenToolPromptIds = new Set<string>()` per request. 

In total, I no longer see any blockers for the current Claude-path, but I would still keep these 2 edge cases + 1 nit in review.

PR #78, opened 2026-04-14
feat(chat): stop button to interrupt in-flight chat runs

lee-to:
## Summary

- Adds a Stop button to the chat input that replaces Send while a run is streaming. Clicking it aborts the underlying runtime call and unwinds the request cleanly.
- Backend keeps a `conversationId → AbortController` map, exposes `POST /chat/:conversationId/abort`, and classifies the thrown `AbortError` so aborted runs emit `chat:error` with code `"aborted"` (HTTP 409) instead of being reported as failures.
- Frontend hook exposes `abortStream()`, `ChatPanel` renders the Stop button (Square icon, destructive variant — composition of existing UI primitives, no new ones).

## Test plan

- [x] `npm run lint`
- [x] `npm run build`
- [x] `npm run test` — 232 API tests pass, including two new ones (404 for unknown conversation, full round-trip aborting an in-flight run)
- [x] `npm run format`
- [x] `npm run ai:validate`
- [ ] Manual verification: start a long run, press Stop, confirm the stream ends immediately and the UI reflects the aborted state.

## Notes

- Independent of #77 (AskUserQuestion rendering) — no conflicts.
- Honors `AbortController` on both Claude transports: SDK forwards the signal to `query()`, CLI kills the spawn.

### DIFF
--- a/packages/api/src/__tests__/chat.test.ts
+++ b/packages/api/src/__tests__/chat.test.ts
@@ -681,4 +681,50 @@ describe("chat API", () => {
       ),
     ).toBe(true);
   });
+
+  it("persists the runtime session link on abort so the next turn can resume", async () => {
+    mockAdapterRun.mockImplementation(async (input: RuntimeRunInput) => {
+      // Runtime emits the session id before the abort fires.
+      input.execution?.onEvent?.({
+        type: "system:init",
+        timestamp: new Date().toISOString(),
+        data: { sessionId: "runtime-session-preserved" },
+      });
+      await new Promise<void>((_resolve, reject) => {
+        input.execution?.abortController?.signal.addEventListener("abort", () => {
+          const err = new Error("The operation was aborted");
+          (err as Error & { name: string }).name = "AbortError";
+          reject(err);
+        });
+      });
+      return { outputText: "", sessionId: "runtime-session-preserved" };
+    });
+
+    mockUpdateChatSession.mockClear();
+
+    const conversationId = crypto.randomUUID();
+    const chatPromise = app.request("/chat", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        projectId: "project-1",
+        message: "first turn",
+        clientId: "client-1",
+        conversationId,
+      }),
+    });
+
+    await new Promise((r) => setTimeout(r, 20));
+    await app.request(`/chat/${conversationId}/abort`, { method: "POST" });
+    const chatRes = await chatPromise;
+    expect(chatRes.status).toBe(409);
+
+    expect(
+      mockUpdateChatSession.mock.calls.some(
+        (call) =>
+          (call[1] as { runtimeSessionId?: string } | null)?.runtimeSessionId ===
+          "runtime-session-preserved",
+      ),
+    ).toBe(true);
+  });
 });
--- a/packages/api/src/__tests__/chat.test.ts
+++ b/packages/api/src/__tests__/chat.test.ts
@@ -586,4 +586,53 @@ describe("chat API", () => {
       }),
     );
   });
+
+  it("returns 404 when aborting an unknown conversation", async () => {
+    const res = await app.request("/chat/unknown-conversation-id/abort", { method: "POST" });
+    expect(res.status).toBe(404);
+  });
+
+  it("aborts an in-flight chat run and surfaces chat:error with code 'aborted'", async () => {
+    let capturedController: AbortController | undefined;
+    mockAdapterRun.mockImplementation(async (input: RuntimeRunInput) => {
+      capturedController = input.execution?.abortController;
+      // Simulate adapter honoring the AbortController by throwing AbortError
+      await new Promise<void>((_resolve, reject) => {
+        capturedController?.signal.addEventListener("abort", () => {
+          const err = new Error("The operation was aborted");
+          (err as Error & { name: string }).name = "AbortError";
+          reject(err);
+        });
+      });
+      return { outputText: "", sessionId: "runtime-session-1" };
+    });
+
+    const conversationId = crypto.randomUUID();
+    const chatPromise = app.request("/chat", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        projectId: "project-1",
+        message: "long task",
+        clientId: "client-1",
+        conversationId,
+      }),
+    });
+
+    // Give the route a tick to register the AbortController before we abort.
+    await new Promise((r) => setTimeout(r, 20));
+    const abortRes = await app.request(`/chat/${conversationId}/abort`, { method: "POST" });
+    expect(abortRes.status).toBe(204);
+
+    const chatRes = await chatPromise;
+    expect(chatRes.status).toBe(409);
+    const body = await chatRes.json();
+    expect(body.code).toBe("aborted");
+
+    expect(
+      mockSendToClient.mock.calls.some(
+        (call) => call[1]?.type === "chat:error" && call[1].payload?.code === "aborted",
+      ),
+    ).toBe(true);
+  });
 });
--- a/packages/api/src/__tests__/chat.test.ts
+++ b/packages/api/src/__tests__/chat.test.ts
@@ -628,10 +628,56 @@ describe("chat API", () => {
     expect(chatRes.status).toBe(409);
     const body = await chatRes.json();
     expect(body.code).toBe("aborted");
+    // The response must expose the server-resolved sessionId so the client
+    // can promote a fresh "new chat" that was stopped before the happy-path
+    // onSessionResolved handshake.
+    expect(typeof body.sessionId === "string" || body.sessionId === null).toBe(true);
+    expect(body.conversationId).toBe(conversationId);
 
     expect(
       mockSendToClient.mock.calls.some(
         (call) => call[1]?.type === "chat:error" && call[1].payload?.code === "aborted",
+      ),
+    ).toBe(true);
+  });
+
+  it("persists partial streamed output when a run is aborted mid-stream", async () => {
+    mockAdapterRun.mockImplementation(async (input: RuntimeRunInput) => {
+      input.execution?.onEvent?.({
+        type: "stream:text",
+        message: "Partial reply",
+        timestamp: new Date().toISOString(),
+      });
+      await new Promise<void>((_resolve, reject) => {
+        input.execution?.abortController?.signal.addEventListener("abort", () => {
+          const err = new Error("The operation was aborted");
+          (err as Error & { name: string }).name = "AbortError";
+          reject(err);
+        });
+      });
+      return { outputText: "", sessionId: "runtime-session-2" };
+    });
+
+    const conversationId = crypto.randomUUID();
+    const chatPromise = app.request("/chat", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        projectId: "project-1",
+        message: "long task",
+        clientId: "client-1",
+        conversationId,
+      }),
+    });
+
+    await new Promise((r) => setTimeout(r, 20));
+    await app.request(`/chat/${conversationId}/abort`, { method: "POST" });
+    const chatRes = await chatPromise;
+    expect(chatRes.status).toBe(409);
+
+    expect(
+      mockCreateChatMessage.mock.calls.some(
+        (call) => call[0]?.role === "assistant" && call[0].content === "Partial reply",
       ),
     ).toBe(true);
   });
--- a/packages/api/src/__tests__/chat.test.ts
+++ b/packages/api/src/__tests__/chat.test.ts
@@ -680,6 +680,60 @@ describe("chat API", () => {
         (call) => call[0]?.role === "assistant" && call[0].content === "Partial reply",
       ),
     ).toBe(true);
+
+    // The 409 body must surface the partial reply so no-WS clients can render
+    // what was saved server-side without a round-trip to load messages.
+    const body = await chatRes.json();
+    expect(body.assistantMessage).toBe("Partial reply");
+  });
+
+  it("returns saved attachments in the 409 abort response", async () => {
+    mockPersistAttachments.mockResolvedValue([
+      {
+        name: "spec.md",
+        mimeType: "text/markdown",
+        size: 10,
+        path: "storage/chat/s1/spec.md",
+      },
+    ]);
+    mockAdapterRun.mockImplementation(async (input: RuntimeRunInput) => {
+      await new Promise<void>((_resolve, reject) => {
+        input.execution?.abortController?.signal.addEventListener("abort", () => {
+          const err = new Error("The operation was aborted");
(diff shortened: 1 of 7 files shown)


ichinya:
**Request changes**

Thanks, the feature is useful and generally well-assembled: the PR added Stop instead of Send during the stream, the backend added `POST /chat/:conversationId/abort` with the `AbortController`-registry, and the API tests already have both 404 for unknown conversation and round-trip scenario with `code: "aborted"`.

1. **Blocker: Stop may interrupt the wrong run if there are multiple streams in different sessions at the same time.**
   `useChat` is already designed as a per-session streaming state: it has `activeStreamsRef`, `sessionStreamsRef`, `isSessionStreaming()`, in-flight state restoration when switching sessions, and `isStreaming` based on the current selected session. But the new `abortStream()` takes the `conversationId` from the `currentConversationIdRef`, which is updated as “most recently dispatched message” and is not bound to the current `streamKey`/`activeSessionId`. The scenario is as follows: we start a stream in session A, switch to B, send a request there, and return to A. The Stop button in A will send an abort to B. I would calculate the abort target from the current `streamKey` (or from `sessionStreamsRef` / `conversationIdForNoSession`), rather than from a single global ref, and I would definitely add a web test for switching between sessions with two in-flight runs. 

2. **Blocker: aborting the first message in “New chat” does not resolve the active session.**
   `handleNewChat()` puts the UI in new-chat mode via `newSession()` + `clearActiveSession()`, and `useChatSessions` in this mode keeps `resolvedSessionId = null` until someone explicitly calls `setActiveSessionId`. In the happy path, this is done by `useChat` via `onSessionResolved(result.sessionId)` after a successful response from `/chat`. But the backend creates a chat session in advance, before the runtime starts, and sends out `chat:session_created`, whereas the abort-branch returns `409` with only `{ error, code }` without `sessionId`. As a result, after Stop, a new chat session already exists on the server, but the current screen may remain without an active `sessionId`; such a thread is easy to “lose” from the main view until the user manually selects it in the list. I would either return the `sessionId` in the abort-response, or resolve the active session earlier/via a separate event.

3. **Strong UX/story note: partial output after abort on a fresh session doesn't seem to survive a reload.**
   In the frontend, tokens are only stored in the in-memory stream state (`sessionStreamsRef`/`messages`). In the backend, both the `runtimeSessionId` and the assistant message are only stored after the `result` is successfully completed; the abort branch exits the `catch` earlier. And loading history for a regular DB session comes from the DB and refers to runtime messages only if the session already has a `runtimeSessionId/agentSessionId`. For a fresh auto-created session stopped before `result`, it looks like a loss of already streamed assistant text after reopen/reload. If such a discard is intended - ok, but it should be explicitly fixed; if not - partial response should be saved on abort.

4. **Non-blocking: there is not enough web-level coverage for new regressions.**
   Currently, only API tests for unknown conversation and round-trip abort have been added to the diff; there are no changes in the web tests, although the main risks are in `useChat`/`ChatPanel`. I would add at least two scenarios: “abort in first-message new chat” and “abort in the selected session with two parallel in-flight streams”.

After fixing the first two points, I would already be in favor of merge: the idea itself is good, the backend part with the classification of `AbortError` looks neat, and the Stop button in the chat is definitely needed.

lee-to:
Addressed all four points in 8dfc016:

1. **Abort targeting (blocker)** — `abortStream()` now resolves the conversation from `activeStreamsRef` by matching the currently-viewed session's `streamKey` (or `conversationIdForNoSession`), not a global "last dispatched" ref. Covered by a new web test that starts two in-flight streams in sessions A and B, switches back to A, and asserts the abort targets A's conversation.

2. **New-chat abort resolves session (blocker)** — the 409 response now includes `sessionId` and `conversationId`. `ApiError` carries the response body, and `useChat`'s catch branch promotes the server-created session via `onSessionResolved` + `chat:session_created` dispatch when `code === "aborted"`. New web test covers first-message new-chat abort resolving the sessionId.

3. **Partial output persistence** — `fullAssistantResponse` hoisted out of the try so the abort branch can `createChatMessage({ role: "assistant", content: partial })` when any tokens were streamed before the abort. API test asserts the partial reply is persisted.

4. **Web-level coverage** — added the two scenarios above to `useChat.test.ts`.

Lint + build + full test suite (564 web tests, 234 API tests) all green locally.

ichinya:
The second commit, 8dfc016, actually closed the previous comments: abort now targets the current session, 409 on abort returns `sessionId` and `conversationId`, partial output is preserved, and new web/API tests have been added.

In the review, I would now leave the following comments:

1. **Blocker: Early Stop may still not work.**
   `POST /chat/:conversationId/abort` returns 404 if the conversation is not yet in `activeChatRuns`, and `activeChatRuns.set(chatConversationId, abortController)` happens only after project lookup, runtime resolution, and auto-create session. Even the API test specifically waits 20 ms “to register the AbortController before we abort”. This means that clicking Stop immediately after Send may still result in a no-op: the client will swallow the 404, and the original `/chat` will continue executing. I would either register the controller immediately after validating the body/calculating the `conversationId`, or I would make a `pendingAbort` token that `/chat` checks before dispatching in runtime. [api/src/routes/chat.ts](https://github.com/lee-to/aif-handoff/blob/feat/chat-abort/packages/api/src/routes/chat.ts)

2. **Strong note: the abort-path does not link the DB session to the runtime-session.**
   For a fresh session, `createChatSession(...)` is called without `runtimeSessionId`; the next turn is summarized only from `dbSession.runtimeSessionId ?? dbSession.agentSessionId`; the reference to the runtime-session is written to the DB only after a successful `result`; in the abort-branch, the partial assistant text is now saved, but the `runtimeSessionId` is not updated. According to the current code, this means that after the abort of the first message, the next request will be sent without a resume context; if the runtime managed to create its own session before the abort, this context will be lost. I would either save the runtime-session id in the abort-path, if the adapter can provide it before the run is completed, or explicitly mention in the PR that the abort of the first turn breaks runtime continuity. [api/src/routes/chat.ts](https://github.com/lee-to/aif-handoff/blob/feat/chat-abort/packages/api/src/routes/chat.ts)

3. **Non-blocking UX note: abort now looks like a regular assistant-reply.**
   The frontend adds `{ role: "assistant", content: message }` to the transcript for any `chat:error`, and the backend sends a `chat:error` with the text `"Chat run aborted by user"` for abort. At the same time, only partial assistant reply is saved in the abort database. As a result, before reload, the user sees a separate bubble “Chat run aborted by user”, and after reload, it disappears. For `code === "aborted"` I would not write this to the history as an assistant message, but instead display a separate status/banner. [web/src/hooks/useChat.ts](https://github.com/lee-to/aif-handoff/blob/feat/chat-abort/packages/web/src/hooks/useChat.ts)

lee-to:
Addressed all three points in afb813b:

1. **Early-Stop race (blocker)** — the `AbortController` is now registered immediately after body validation, before project lookup / runtime resolution / session auto-create. An early Stop click now trips the already-aborted signal inside `adapter.run()` instead of 404ing against an unregistered controller. The handler body is wrapped in a single outer `try/catch/finally` so cleanup (`activeChatRuns.delete`) holds across all early-return paths.

2. **Runtime session continuity on abort (strong)** — the chat route now captures the `sessionId` from the runtime `system:init` event into `runtimeSessionIdFromEvents`. The abort branch calls `updateChatSession({ runtimeSessionId })` using that value, so the next turn resumes the same runtime thread instead of starting a fresh one. New API test `persists the runtime session link on abort` asserts the link is written.

3. **Aborted no longer looks like a regular assistant reply (UX)** — `useChat` skips pushing the `"Chat run aborted by user"` bubble when `code === "aborted"` on both the WS error path and the HTTP 409 catch. `ChatPanel` surfaces a dedicated `Stopped` banner via `chatErrorCode === "aborted"` (same pattern as `CHAT_USAGE_LIMIT`). The streamed partial reply above the banner is still persisted to DB and survives reload.

Local checks: `npm run lint`, `npm run build`, `npm run test` (234 API + 564 web), `npm run format` — all green.

ichinya:
1. **Blocker: post-request cleanup in `useChat` is still not bound to the current session.**
   WS-handlers are already session-aware via `isCurrentStream(streamKey)`, but in the HTTP-path after `sendChatMessage()`, there are unconditional `setIsStreaming(false)` and, for abort, `setChatErrorCode("aborted")`; the same applies to success fallback and forced-stop timers. Since `ChatPanel` renders Stop and banner by `isStreaming` and `chatErrorCode`, a late abort or HTTP fallback from background session A can remove Stop and show `Stopped` in an already open session B. I would wrap this tail in the same 'isCurrentStream(activeStreamKey)' guard and add a test for the scenario “A terminates while the user is already watching B".

2. **Strong point: abort with attachments seems to leave the current user bubble without a download link until reload.**
   The optimistic user message is created only from the metadata of the file, and the clickable attachment in the `MessageBubble` appears only when there is a `att.path` and a `sessionId`. The upgrade to server-resolved `result.attachments` is only done in the happy path; in the abort catch, there is a cleanup of the stream-state and banner, but there is no patch or refetch. According to the code, if you stop the request with files, the backend will already save the `savedAttachments` in the DB, but the current UI may remain with a pathless chip until you reload or reopen the session. I would either return `attachments` and 409 `aborted`, or do a refetch of the current session after abort.

3. **Non-blocking, but real: no-WS abort still doesn't show the saved partial reply in the current transcript.**
   `useChat` can continue without `clientId`, and the backend doesn't send `chat:token` in this mode at all. On success, this is compensated by HTTP-fallback through `result.assistantMessage`, but on `409 aborted`, the front only sets `chatErrorCode="aborted"` and displays the banner “Any partial reply above has been saved”. However, the API test separately confirms that the partial output on abort is indeed saved. As a result, if there is no WS, the partial reply may be saved on the server but not displayed in the current UI until a reload is performed. In this case, I would either add a partial `assistantMessage` in the aborted response or perform a refetch of messages after an abort if no tokens are received.

lee-to:
Addressed all three review points in 9da2f14.

**1. Blocker — HTTP-path cleanup is now session-scoped.** Extracted `isCurrentStream(streamKey)` to hook scope (`useChat.ts`) and gate every post-request UI mutation behind it: the HTTP fallback timer, the forced-stop timer, the happy-path attachments patch, the happy-path session-ref promotion (`currentSessionIdRef`/`onSessionResolved`), and the catch-block `setIsStreaming`/`setChatErrorCode`/`setMessages` calls. A background session terminating no longer hides Stop or shows `Stopped` on a different viewed session. Covered by the new test _"does not clear Stop or show Stopped on session B when session A's run aborts in the background"_.

**2. Strong point — aborted uploads keep their download link.** Hoisted `savedAttachments` in `chat.ts` so the 409 body echoes server-resolved attachment paths, and the catch in `useChat.ts` patches both the React state and the stream-scoped message list with them. Covered by _"returns saved attachments in the 409 abort response"_ (API) and _"upgrades user message attachments from the 409 abort body"_ (web).

**3. Non-blocking — no-WS abort now shows the saved partial.** The 409 response now carries `assistantMessage: partial || null`, and the catch appends it to the transcript when no WS tokens were accumulated. Covered by the extended _"persists partial streamed output when a run is aborted mid-stream"_ (API) and _"appends the partial assistant reply from the 409 abort body when no WS tokens arrived"_ (web).

All checks: `npm run lint` clean, `npm run build` green, 235 API tests pass, 567 web tests pass, `npm run format` clean.

ichinya:
**Abort with attachments is still racey in the regular WS-path.** On abort, the backend still raises `chat:error` with `code="aborted"`, and the front-end `handleError` immediately clears `activeStreamsRef` and `sessionStreamsRef`. The new HTTP abort-catch upgrades server-resolved attachment paths only if `state` still exists (`if (isAbortedError && state)`), and only then does it do `setMessages(...attachments...)`. So if the `chat:error` arrives before the HTTP `409`, the stream state is already deleted and the patch will not happen.

This is user-visible: `MessageBubble` makes the attachment clickable only when there is `att.path && sessionId`, and the backend already has time to save the `savedAttachments` with the `path` in the DB user-message before the runtime-call. This means that after Stop, the current bubble may still be without a download-link until reload, even though the file has already been saved on the server. The current test, `upgrades user message attachments from the 409 abort body`, only checks for a clean HTTP `409` reject and does not simulate the actual order of `chat:error` + `409`, so this race is not yet covered. I would either not remove the stream-state in `handleError` before processing the HTTP abort body, or I would throw the resolved attachments into the WS abort event and patch the UI right there, plus I would add a test for the sequence `chat:error` → HTTP `409`.

lee-to:
Addressed in 6f1b4bf.

**Race fix** — decoupled the React-state attachment patch from the in-flight `state` presence in `useChat.ts`. The outer guard is now `if (isAbortedError)` and the `state` check wraps only the stream-scoped patch (for session-switch restoration). The React patch runs off the `userMessage` closure reference, so it still lands when WS `chat:error` with `code="aborted"` cleared `activeStreamsRef`/`sessionStreamsRef` before the 409 arrived. Kept the partial-assistant append gated on `state`+`!hasAccumulatedTokens` to avoid re-appending text that WS tokens already streamed into React state.

**Regression test** — new `upgrades user message attachments when chat:error races ahead of the 409 abort body` in `useChat.test.ts` dispatches `chat:error` (code=aborted) first, then rejects the pending `sendChatMessage` with a 409 carrying attachments, and asserts the user bubble got the resolved `path`. Verified the test fails without the fix and passes with it.

All checks green: `npm run lint`, `npm run build`, `npm run test` (569 web tests, all useChat tests pass), `npm run format` clean.

ichinya:
1. **Blocker: instant Stop before auto-create session leaves an orphaned optimistic transcript.**
   `POST /chat` now registers `AbortController` **before** project lookup / runtime resolution / session auto-create. In the abort branch, the response returns `sessionId: chatSessionId`, and the partial/runtime-link is also tied to `if (chatSessionId)`. On the front, `useChat` optimistically puts the first user message in `messages`, and in the abort-catch, it promotes the new chat only when `abortData.sessionId` is present; otherwise, it simply stops streaming and leaves the transcript as is. As a result, an early Stop on the **very first** message may leave a user bubble in the UI that does not exist on the server; the next send will visually continue this transcript, even though the backend has only seen the second turn. Additionally, I see that the current abort tests still call `/abort` only after `setTimeout(..., 20)`, so the pre-session window is not covered. I would either ensure that `createChatSession` is called before entering the abortable slow path, or that `409 { code: "aborted", sessionId: null }` rolls back the optimistic first turn / resets the draft state, and adds an API+web test for immediate stop without delay.

2. **Non-blocking, but I would correct it: partial assistant from 409 is still lost in the race `chat:error` → HTTP 409.**
   For attachments, the code already makes a special mirror in the React state, even if `chat:error` has managed to clear the stream-state, but the append partial assistant is still tied to `state && !hasAccumulatedTokens`. The backend specifically returns `assistantMessage` in the 409 body when aborting, but if the stream-state is already cleared, this text will not be added to the current transcript. In the tests, I see separate coverage for no-WS partial and for attachment-race, but not for a similar race with `assistantMessage`. Here, the same approach as for attachments is suggested: mirror partial directly into the React state for the current viewed stream, plus a separate test for the `chat:error` sequence before 409 with assistantMessage.

ichinya (approved):

PR #79, opened 2026-04-14
feat: universal language directive injection for AI prompts

### LINKED ISSUE #73: [bug] Tasks and their descriptions, as well as review artifacts, are generated in English even though the project language is set to Russian
## Description

file specifies that everything should be generated in Russian, except for technical terms. However, in most cases, everything is still generated in English.

## Steps to Reproduce

1. Create a project using the aif skill and specify that you want to work in Russian. 
2. Check that handoff has picked up the config. 
3. Try generating a ROADMAP — the tasks in it will be in English. As you go through the handoff flow, the task accumulates review artifacts. They are also in English.

## Expected Behavior

Artifacts in Russian.

## Actual Behavior

Artifacts in English.

## Environment

- OS: Debian 13
- Node.js version: 24.14.1
- npm version: 11.11.0

lee-to:
## Summary

- Read `language.{artifacts,technical_terms}` from `.ai-factory/config.yaml` and inject a short directive into `execution.systemPromptAppend` inside `RuntimeRegistry.wrapAdapter()` — a single choke point that covers every AI-backed call path (subagents, roadmap generation, commit generation, fast-fix, chat, reviewGate) across all transports (SDK/CLI/API)
- Default `artifacts: en` → empty directive → zero behavioural change for existing users
- Directive is appended after existing `systemPromptAppend` so project-scope and review-diff-scope rules keep their emphasis; prompt/session/agent-definition paths are untouched

Closes #73

## Test plan

- [x] `buildLanguageDirective` unit tests (en/unset/ru/fr/case-insensitive/BCP-47/unknown code)
- [x] `projectConfig` tests for language block defaults, partial parsing, case-normalization, invalid `technical_terms` fallback
- [x] Integration tests in `registry.test.ts`: run + resume injection, missing `projectRoot` no-op, `en` no-op, malformed YAML → WARN + no throw
- [x] Lint + build green on `@aif/shared` and `@aif/runtime`
- [ ] Manual smoke test with `language.artifacts: ru` on a real task (reviewer / roadmap generation)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -135,7 +135,7 @@ Keys:
 
 - `artifacts` — BCP-47-ish language code. Values are validated against a conservative
   `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$` pattern after trim+lowercase (both `-` and `_` are
-  accepted as subtag separators, so `en-US` and `en_GB` parse identically); tags that fail the
+  accepted as subtag separators, so `en-US` and `en_US` parse identically); tags that fail the
   pattern (typos, non-ASCII strings) silently fall back to the default `en` rather than being
   embedded raw in the system directive. Any regional tag whose primary subtag is `en` (`en-US`,
   `en_GB`, …) is also treated as a no-op. Any other valid tag appends a short system directive
@@ -386,7 +386,7 @@ The config is editable via the **Global Settings** dialog in the web UI (gear ic
 | Key               | Default | Options                                                                                                                                                                                                                       |
 | ----------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `ui`              | `en`    | Reserved for future UI localisation (currently informational).                                                                                                                                                                |
-| `artifacts`       | `en`    | BCP-47-ish tag, validated against `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$` after trim+lowercase. `-` and `_` are interchangeable separators (`en-US` == `en_GB`). Any `en*` primary subtag and invalid tags are treated as no-op. |
+| `artifacts`       | `en`    | BCP-47-ish tag, validated against `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$` after trim+lowercase. `-` and `_` are interchangeable separators (`en-US` == `en_US`). Any `en*` primary subtag and invalid tags are treated as no-op. |
 | `technical_terms` | `keep`  | `keep`, `translate`                                                                                                                                                                                                           |
 
 **`paths`** — custom paths for AI Factory artifacts (relative to project root):
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -146,12 +146,13 @@ Keys:
   of those tokens where a good equivalent exists.
 - `ui` — reserved for future UI-side localisation; currently informational only.
 
-The directive is appended to `execution.systemPromptAppend` — never to `prompt` — so resume
-sessions, slash commands, and agent definitions continue to work unchanged. Existing appends
-(project-scope, review-diff-scope) are preserved verbatim and placed BEFORE the language
-directive so scope rules keep their emphasis.
+Injection is uniform at the registry layer: the directive is written to
+`execution.systemPromptAppend` (never to `input.prompt`), so resume sessions, slash commands,
+and agent definitions continue to work unchanged. Existing appends (project-scope,
+review-diff-scope) are preserved verbatim and placed BEFORE the language directive so scope
+rules keep their emphasis.
 
-Transport delivery matrix (how each adapter surfaces `systemPromptAppend` to the model):
+How each adapter then delivers that append block to the model depends on the transport:
 
 | Adapter    | SDK                          | CLI                      | API            |
 | ---------- | ---------------------------- | ------------------------ | -------------- |
@@ -161,8 +162,10 @@ Transport delivery matrix (how each adapter surfaces `systemPromptAppend` to the
 | OpenCode   | n/a                          | n/a                      | system message |
 
 The Codex SDK/CLI paths do not expose a dedicated system-prompt slot, so the adapter prepends
-the append block to the user prompt separated by a blank line. This keeps the registry-level
-injection guaranteed to reach the model on every Codex transport, matching the API path.
+the append block to the user prompt separated by a blank line. That happens inside the
+adapter, after the registry has set `systemPromptAppend` — the caller and the registry never
+mutate `input.prompt` themselves. This keeps the registry-level injection guaranteed to reach
+the model on every Codex transport, matching the API path.
 
 To disable: leave `artifacts: en` (the default).
 
@@ -378,13 +381,13 @@ The config is editable via the **Global Settings** dialog in the web UI (gear ic
 
 ### Sections
 
-**`language`** — controls AI-generated content language:
+**`language`** — controls AI-generated content language. See [Project Language](#project-language) for the full directive contract, validation rules, and per-transport delivery matrix:
 
-| Key               | Default | Options                                                    |
-| ----------------- | ------- | ---------------------------------------------------------- |
-| `ui`              | `en`    | `en`, `ru`, `de`, `fr`, `es`, `zh`, `ja`, `ko`, `pt`, `it` |
-| `artifacts`       | `en`    | Same as `ui`                                               |
-| `technical_terms` | `keep`  | `keep`, `translate`                                        |
+| Key               | Default | Options                                                                                                                                                                                                                       |
+| ----------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ui`              | `en`    | Reserved for future UI localisation (currently informational).                                                                                                                                                                |
+| `artifacts`       | `en`    | BCP-47-ish tag, validated against `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$` after trim+lowercase. `-` and `_` are interchangeable separators (`en-US` == `en_GB`). Any `en*` primary subtag and invalid tags are treated as no-op. |
+| `technical_terms` | `keep`  | `keep`, `translate`                                                                                                                                                                                                           |
 
 **`paths`** — custom paths for AI Factory artifacts (relative to project root):
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -134,11 +134,12 @@ language:
 Keys:
 
 - `artifacts` — BCP-47-ish language code. Values are validated against a conservative
-  `^[a-z]{2,3}(-[a-z0-9]{2,8})*$` pattern after trim+lowercase; tags that fail the pattern
-  (typos, non-ASCII strings) silently fall back to the default `en` rather than being embedded
-  raw in the system directive. Any regional tag whose primary subtag is `en` (`en-US`, `en_GB`,
-  …) is also treated as a no-op. Any other valid tag appends a short system directive asking
-  the model to write artifacts in that language.
+  `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$` pattern after trim+lowercase (both `-` and `_` are
+  accepted as subtag separators, so `en-US` and `en_GB` parse identically); tags that fail the
+  pattern (typos, non-ASCII strings) silently fall back to the default `en` rather than being
+  embedded raw in the system directive. Any regional tag whose primary subtag is `en` (`en-US`,
+  `en_GB`, …) is also treated as a no-op. Any other valid tag appends a short system directive
+  asking the model to write artifacts in that language.
 - `technical_terms` — `keep` (default) instructs the model to leave identifiers, API/function
   names, file paths, CLI flags, environment variables, code snippets, and log/error strings in
   English even when the rest of the text is translated. `translate` allows natural translation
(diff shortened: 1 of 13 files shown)


ichinya:
In general, I like the direction: centralizing language-policy injection in `RuntimeRegistry.wrapAdapter()` is a good idea, because the change actually goes through both wrappers, `run()` and `resume()`. The backward compatibility is also well-thought-out: the `projectConfig` has a `language` block with a default of `artifacts: "en"`, and `buildLanguageDirective()` returns an empty string for unset/`en`, so the behavior does not change when the default configuration is used. The documentation and registry-level tests generally support this design. 

The main blocker is that the PR currently promises more than it actually guarantees. The documentation explicitly states that it covers “every call path … across all transports (SDK/CLI/API),” but I don’t see this match in Codex. The Codex adapter itself declares `SDK`, `CLI`, and `API`, with the default transport being `CLI`. In the `API` path, `buildMessages()` explicitly adds `input.execution.systemPromptAppend` to the system message. However, in the `SDK`-path, I only see `buildThreadOptions()`, `buildTurnOptions()`, and then `thread.runStreamed(input.prompt, turnOpts)` without explicitly calling `systemPromptAppend`. In the `CLI`-path, `normalizeCliArgs()` works around `prompt/model/session`, and the process itself only writes `input.prompt` to stdin; I also do not see any processing of `systemPromptAppend`. This means that registry injection is present, but for the most common Codex scenario, it does not seem to have an effect on the model. For merge, I would ask either to bring the Codex CLI/SDK to the same semantics, or to narrow the scope of PR/documentation/tests and not promise universal behavior where transport does not consume this field. 

The second point is that `language.artifacts` is normalized too freely now. In the doc, it is described as “BCP-47-ish language code”, but `normalizeLanguage()` accepts any non-empty string, and `buildLanguageDirective()` for unknown values substitutes raw value in system directive. This means that typos or garbage values will be appended as-is; plus, `en-US`/`en_GB` will still inject the English directive, even though the builder already knows how to extract the primary subtag for the displayed language name. I would validate the value against a conservative language tag pattern and make a "no-op or not" decision based on the primary subtag, rather than an exact match with `"en"`. 

The third point is that the current tests check the boundary well at the registry level, but not at the actual transports. The diff shows that the tests use a fake capturing adapter and check the filling of `execution.systemPromptAppend`; this does not prove that the built-in transport-specific adapter paths actually consume this field. Given the discrepancy with Codex, I would add transport-level regression tests at least for the Codex API/CLI/SDK, or explicitly document the compatibility matrix if some transports do not support system-level append by design.

Non-blocking: `technical_terms` is normalized more strictly than `ui` and `artifacts` — only the exact lowercase value `"translate"` is accepted, so `"Translate"` or `" translate "` will quietly roll back to `"keep"`. This is a minor issue, but it looks inconsistent with the rest of the config normalization. 

In summary, I would recommend using `Request changes`; after closing the transport-gap in Codex, the rest of the improvements are no longer considered blockers.

lee-to:
Thanks for the thorough review — addressed all three blockers in eb5542e:

**1. Codex SDK/CLI transport gap.** Both transports now prepend `execution.systemPromptAppend` to the user prompt (SDK: before `thread.runStreamed`; CLI: before stdin write), since neither exposes a dedicated system-prompt slot. The registry's language directive now actually reaches the model on every Codex transport, matching the API path.

**2. `language.artifacts` validation.** `normalizeLanguage()` now validates `ui`/`artifacts` against a conservative BCP-47-ish pattern (`^[a-z]{2,3}(-[a-z0-9]{2,8})*$`) after trim+lowercase — typos and non-ASCII garbage silently fall back to the default `en` instead of being embedded verbatim. `buildLanguageDirective()` also treats any tag whose primary subtag is `en` (`en-US`, `en_GB`, …) as a no-op so regional English no longer injects a redundant directive.

**3. Transport-level regression tests.** Added tests to `codexSdk.test.ts` and `codexCli.test.ts` that assert `systemPromptAppend` reaches `runStreamed` args and CLI stdin respectively. Added `projectConfig` tests for BCP-47 validation and `languagePolicy` tests for `en-US`/`en_GB` no-op.

**Non-blocking.** `technical_terms` is now normalized with trim+lowercase for parity with `ui`/`artifacts`, so `"Translate"` / `" translate "` map to `translate` as expected.

Docs (`docs/configuration.md`) updated with a transport delivery matrix explaining how each adapter surfaces `systemPromptAppend`, plus the validation behaviour.

`npm run ai:validate` passes (lint + build + tests).

ichinya:
I've rechecked the PR based on the current state: the previous major blocker regarding the Codex SDK/CLI has been resolved. A separate fix commit, eb5542e, has been added to the PR, and the docs now explicitly describe the delivery matrix for Codex. Additionally, the code for `systemPromptAppend` is now included in the Codex CLI through prepending it to stdin and in the Codex SDK through prepending it to `runStreamed(...)`; separate tests have been added to support this. 

 Here are the comments I would like to make at this point.

1. **There is a real gap for `codexCliArgs` with `{prompt}` / `--prompt`.**
   In `normalizeCliArgs()`, custom CLI-args are substituted in `{prompt}` with `input.prompt`, without `execution.systemPromptAppend`. Then, `shouldWritePromptToStdin()` disables writing to stdin if prompt is already in args or if there is `--prompt`/`--prompt=...`. The only place where append is actually combined with prompt is in the stdin writing branch. That is, for a supported scenario with a custom `codexCliArgs` language directive, it doesn't reach the model again. This is particularly noticeable because the tests already capture `--prompt={prompt}` → `--prompt=Implement feature` and separately refer to `codexCliArgs` as a “full escape hatch”. If the PR wants to continue being referred to as “universal language directive injection,” I would either merge `systemPromptAppend` with prompt **before** template-substitution or honestly document that a custom `codexCliArgs` disables the guarantee of appends delivery. [runtime/src/adapters/codex/cli.ts](https://github.com/lee-to/aif-handoff/blob/feature/language-directive-injection/packages/runtime/src/adapters/codex/cli.ts)

2. **There is a syncronization of docs vs implementation by the format of language tag.**
   The docs say that validation goes by the pattern `^[a-z]{2,3}(-[a-z0-9]{2,8})*$`, that is, only with `-`. But the real code accepts both `-` and `_`: `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$`. Moreover, the docs and tests themselves give `en_GB` as a valid no-op example. This doesn't break the behavior, but the documentation text is currently misleading. 

3. **There are still not enough tests for the specific branch where the gap remains.**
   Currently, there is a test for the default CLI stdin-path with `systemPromptAppend`, a test for custom `codexCliArgs` with `--prompt={prompt}`, and run/resume tests at the registry level. However, there is no transport-level regression test for the combination of `systemPromptAppend + custom codexCliArgs`, which is why the current hole is not being caught. Separately for CLI/SDK, there is no explicit transport-level test for `resume + systemPromptAppend`, although the code should behave the same way there. I would add at least one test for a custom `--prompt={prompt}` with append, because this would immediately fix the escape hatch semantics. [runtime/src/__tests__/codexCli.test.ts](https://github.com/lee-to/aif-handoff/blob/feature/language-directive-injection/packages/runtime/src/__tests__/codexCli.test.ts)

To sum up: **the main previous blocker is already closed**, but I would still leave **Request changes** due to the `codexCliArgs` exception; the remaining comments are more of a polishing nature.

lee-to:
Thanks for the careful re-check — all three points addressed in c826660.

**1. `codexCliArgs` `{prompt}` / `--prompt` gap (the real blocker).** Fixed in `packages/runtime/src/adapters/codex/cli.ts`. The prompt is now composed once at the top of `runCodexCli()` (`composePrompt()` prepends `execution.systemPromptAppend` to `input.prompt`) and threaded through both branches:
- `normalizeCliArgs()` substitutes the composed prompt into `{prompt}` placeholders, so the language directive reaches the model via the custom-args escape hatch too.
- The stdin fallback writes the same composed prompt; `shouldWritePromptToStdin()` checks against it so we never send the prompt twice.

**2. Docs vs implementation drift.** `docs/configuration.md` now shows the actual pattern `^[a-z]{2,3}(?:[-_][a-z0-9]{2,8})*$` with an explicit note that both `-` and `_` are accepted, matching `projectConfig.ts` and the `en_GB` example.

**3. Missing transport-level tests.** Added:
- `codexCli.test.ts` — custom `codexCliArgs` with `--prompt={prompt}` + `systemPromptAppend` (the exact gap case): asserts args carry the composed prompt AND stdin is not written.
- `codexCli.test.ts` — resume + `systemPromptAppend`: asserts the composed prompt is written to stdin on the `exec resume` path.
- `codexSdk.test.ts` — resume + `systemPromptAppend`: asserts `resumeThread` is used and `runStreamed` receives the composed prompt.

`npm run ai:validate` green (562 tests). Ready for another pass.

ichinya:
I've rechecked the PR in its current state. The old comments about the Codex SDK/CLI and the `--prompt={prompt}` branch have already been closed: the PR has a separate fix `c826660`, and the description of this commit explicitly lists the fixing of `codexCliArgs`, updating the docs, and transport-level tests for CLI/SDK run+resume

I would leave two comments now.

1. The code edge case in `codexCliArgs` is still there.
 There is still one generic `{prompt}` edge case in the custom `codexCliArgs` path. `normalizeCliArgs()` substitutes `{prompt}` in any argument, but `shouldWritePromptToStdin()` suppresses stdin only for an exact `args.includes(prompt)` match or `--prompt` / `--prompt=*`. Because that check happens after substitution, `arg.includes("{prompt}")` is effectively dead code there, so shapes like `["run", "--json", "--payload=prefix {prompt} suffix"]` would still receive the composed prompt on stdin as well. I’d either carry a `usesPromptPlaceholder` flag out of `normalizeCliArgs()` and skip stdin whenever `{prompt}` was used anywhere, or narrow the contract/tests to `--prompt` specifically.

2. The docs still contain self-contradictions in `language.*`.
   Docs still contradict themselves on `language.*`. The new “Project Language” section says `artifacts` accepts a BCP-47-ish tag pattern and describes behavior for tags like `en-US` / `en_GB`, but the later `Project Config (config.yaml)` table still presents `ui` / `artifacts` as a small fixed enum and `artifacts` as “Same as ui”. Also, “the directive is appended to `execution.systemPromptAppend` — never to `prompt`” reads awkwardly right above the matrix that says Codex SDK/CLI surface it by prepending to the user prompt. I’d either remove the duplicated table or update the later section so it matches the actual runtime behavior and validation rules.

lee-to:
Thanks for the careful re-read. Addressed both in 0be1e89:

1. **`codexCliArgs` `{prompt}` edge case** — `normalizeCliArgs()` now carries a `usesPromptPlaceholder` flag out alongside the substituted args. Detection happens pre-substitution, so composite shapes like `["run", "--json", "--payload=prefix {prompt} suffix"]` now correctly suppress stdin (the old `arg.includes("{prompt}")` post-substitution check was indeed dead code). Added a regression test for that composite shape.

2. **`language.*` docs contradictions** — the Project Config (config.yaml) `language` table now mirrors the BCP-47-ish validation rules and links back to the Project Language section instead of repeating a stale enum. Also reworded the "never to `prompt`" sentence so it no longer contradicts the delivery matrix: the **registry** always writes to `execution.systemPromptAppend` and never touches `input.prompt`; the **Codex SDK/CLI adapter** then prepends that block to the user prompt internally, since those transports have no dedicated system-prompt slot.

Full check suite (lint + build + vitest w/ coverage + format) green locally.

ichinya:
There is still one false-positive edge case in the Codex CLI stdin suppressor.

`shouldWritePromptToStdin()` currently skips stdin when `args.includes(prompt)`, but on the default path `args` always contain generic tokens like `exec`, `--json`, and optionally the model id. That means a prompt that exactly matches one of those tokens would be treated as “already delivered” and would never be written to stdin, even though no prompt placeholder was used.

I’d remove the generic `args.includes(prompt)` branch entirely, or replace it with an explicit signal from `normalizeCliArgs()` that a standalone prompt arg was produced by template substitution. A small regression test with a prompt like `"exec"` on the default path would lock this down.

Minor docs nit: the “separator interchangeability” example currently uses `en-US` vs `en_GB`, which differs not only by separator but also by region. If the point is just `-` vs `_`, `en-US` vs `en_US` (or `en-GB` vs `en_GB`) would be less confusing.

lee-to:
Good catch — addressed both in b02cab7.

1. **`shouldWritePromptToStdin` false-positive.** Removed the `args.includes(prompt)` branch entirely. On the default path `args` always contain generic tokens (`exec`, `--json`, the model id), so a user prompt equal to one of those would be wrongly classified as "already delivered" and never reach the CLI. The `usesPromptPlaceholder` flag (covers `{prompt}` in any shape) plus the explicit `--prompt` / `--prompt=*` check already cover every legitimate embed path, so the generic equality branch was both unsafe and redundant. Added a regression test with `prompt: "exec"` that proves stdin still receives the prompt on the default path.

2. **Docs nit.** Swapped the separator example from `en-US` vs `en_GB` to `en-US` vs `en_US` in both the Project Language section and the Project Config (config.yaml) table, so the point is the pure `-`/`_` interchangeability — the region-tag no-op story is still covered by the separate `en-US` / `en_GB` list further down.

Full check suite green (562 runtime tests + all package test files pass).

ichinya:
One remaining edge case in `normalizeCliArgs()`: placeholder substitution still returns after the first matching placeholder type in each arg. So mixed templates like `--meta={model}:{session_id}` or `--prompt={prompt} ({model})` are only partially expanded. In the second case this is worse than a cosmetic issue, because `usesPromptPlaceholder` will suppress stdin, so the partially unresolved arg is what reaches Codex.

I’d apply substitutions cumulatively per arg (prompt → model → session_id on the same string) and add a regression test for multiple placeholders inside one arg.

ichinya (approved):

PR #80, opened 2026-04-17
fix: harden runtime limit auto-pause state, redaction, and stale UI handling

### LINKED ISSUE #24: [feature] Expose runtime limits in UI and auto-pause tasks at 10% remaining
## Problem
The system already detects usage-limit, quota, and rate-limit exhaustion as external failures, but the UX is reactive:
- the limit is not surfaced clearly in the main interface
- work continues until a hard limit error happens
- only then does the system react

This is late and inefficient for long autonomous runs.

## Proposed Solution
Expose runtime usage limits in the UI and manage task execution proactively.

Requested behavior:
- display total limit, remaining limit or percentage, reset time, and current status in the UI when available
- when remaining limit reaches `10%` or lower, stop starting new work for the affected runtime or task
- move the task into a controlled paused or blocked state instead of waiting for hard exhaustion
- store the reset time when known and use it as the resume time
- automatically resume the task after the limit window resets and return it to its previous working status

## Alternatives Considered
Keep the current reactive behavior that only blocks tasks after a hard usage-limit failure, or rely on the chat-level usage-limit notice alone. Both approaches still waste the last portion of available quota and make autonomous execution less predictable.

## Additional Context
The codebase already has useful building blocks for this: chat UI can display a usage-limit message, usage and quota errors are already classified as external failures, and blocked tasks already support `retryAfter` plus automatic release after the retry window.

ichinya:
## Summary

This PR closes the runtime-limit review findings around state persistence, reset handling, client redaction, and stale UI status.

## What changed

- preserve persisted runtime-limit state after successful runs unless the provider sends an authoritative recovery signal
- stop proactive gating from treating status-only snapshots without a valid future `resetAt` as permanent blocks
- harden provider reset-hint normalization so invalid/out-of-range values are dropped instead of throwing during limit handling
- redact provider-controlled upstream error bodies before they reach chat clients
- redact provider-controlled upstream error bodies before they are persisted into blocked task UI / activity log
- degrade expired runtime-limit snapshots in the UI to a neutral `Expired` state instead of showing stale `Blocked` / `Near Limit`

## Review findings addressed

1. success path no longer clears authoritative limit state without a recovery signal
2. missing reset timestamps no longer become permanent gates
3. provider-controlled reset hints no longer throw before limit handling
4. chat no longer reflects raw upstream provider bodies to the browser
5. blocked task UI no longer persists raw upstream provider bodies
6. runtime-limit UI no longer stays blocked after the reset window has elapsed

## Verification

- `packages/agent`: `npm run lint`, `npm test`
- `packages/web`: `npm run lint`, `npm test`
- repo-level: `npm run ai:validate`

## Notes

This is a hardening pass for runtime limit handling across agent, API, runtime normalization, and web display logic. The behavior is now aligned across coordinator gating and UI presentation.

fixed #24


ichinya:
I managed to get the limits from the codex.
image

next, you need to deal with claude and "anthropic" subscriptions

ichinya:
added grouping by CODEX CLI/SDK accounts
image

ichinya:
Added account grouping in Codex (runtime profiles are grouped)
image

ichinya:
Added "definition" of anthropic-like runtime
image

ichinya:
added getting a quota for glm.
image

ichinya:
a limit has been added to the chat if there is
image

ichinya:
Spark is separate from Codex
image

lee-to:
## Code review

Reviewed the full diff against main (+15413 / −357, 99 files). The six review findings in the PR description are each addressed with matching code and targeted tests, and CI (Build / ESLint / Tests / MCP Unit / npm audit) is green.

### What I verified against the stated findings

1. **Success path no longer clears authoritative limit state without a recovery signal** — `executeSubagentQuery` only calls `refreshRuntimeProfileLimitState` on success when it actually extracted a snapshot from the result events; otherwise it emits `"Preserving runtime limit state after successful subagent execution without an authoritative recovery signal"` and leaves persisted state alone (`packages/agent/src/subagentQuery.ts:961-983`). Covered by the new test at `subagentQuery.test.ts:491`.

2. **Missing reset timestamps no longer become permanent gates** — `evaluateRuntimeLimitGate` explicitly refuses to block when `providerBlockedHint.source === "none"` or the exact-threshold hint has no source, and only gates when `isFuture` (`packages/data/src/index.ts:2038-2094`). `resolveRuntimeLimitFutureHint` backs this via `candidateFromResetAt` / `candidateFromRetryAfter` and the `isFuture` check (`packages/shared/src/runtimeLimitUtils.ts:422-494`).

3. **Provider-controlled reset hints no longer throw before limit handling** — `normalizeTimestamp` in `packages/runtime/src/adapters/claude/limit.ts:62-89` drops non-finite / out-of-range / `NaN` values with a warn-log and returns `null`, instead of letting `new Date(...)` produce `Invalid Date` downstream. Covered by `claudeLimit.test.ts`.

4. **Chat no longer reflects raw upstream provider bodies to the browser** — `classifyChatError` in `packages/api/src/routes/chat.ts:277-316` always returns a fixed user-safe string keyed off `RuntimeErrorCategory`; the raw message is only used for a redacted server-side log via `redactProviderText`.

5. **Blocked task UI no longer persists raw upstream provider bodies** — `stageErrorHandler.classifyStageError` persists only `buildUserSafeExternalReason(err)` as the blocked reason and diagnostics are redacted before activity-log writes (`packages/agent/src/stageErrorHandler.ts:53-236`).

6. **Runtime-limit UI no longer stays blocked after the reset window has elapsed** — `packages/web/src/lib/runtimeLimits.ts:218-260` switches `displayState` to `"expired"` (neutral `tone: "info"`, label `"Expired"`) when the snapshot is an active limit signal but the future hint is already in the past. Covered by `runtimeLimits.test.ts`.

### Additional strengths

- `sanitizeProviderMeta` is well layered: per-provider allow-list + generic allow-list + forbidden-key denylist + nested recursion depth / object-key / array-item / byte caps (`runtimeLimitUtils.ts:1-265`). Truncation is explicit (`_truncated: true`) rather than silent.
- `redactProviderText` covers the usual secret shapes (JWTs, `sk-…`, `gh[pours]_…`, `AIza…`, `ya29.…`, AWS access keys, Slack tokens, bearer tokens) plus a key-value scrubber for `access_token`, `api_key`, `password`, etc.
- `buildRuntimeLimitSignature` uses `stableSortObjectKeys` on the full normalized snapshot, so cache dedupe is order-independent and survives provider-meta re-ordering.
- `ClaudeProviderFamily` / `ClaudeProviderIdentity` resolution correctly restricts local-settings reads to `sdk` / `cli` transports, and `accountFingerprint` uses a SHA-256 of `origin|apiKey` truncated to 16 hex chars — not reversible and scoped by origin.
- Good test coverage on the new modules: `runtimeLimitUtils.test.ts` (364 lines), `claudeLimit.test.ts` (204), `claudeZaiQuota.test.ts` (257), `claudeProviderIdentity.test.ts` (106), `openaiRateLimits.test.ts` (232), `limitEvents.test.ts` (87), plus 335 new lines in `coordinator.test.ts` and 472 in `subagentQuery.test.ts`.
- `stageErrorHandler.ts` is a clean extraction of the error-classification branch out of the coordinator — single responsibility, easier to test.

### Suggestions / non-blocking concerns

1. **Duplicate limit-refresh helpers between agent and api.** `extractRuntimeLimitSnapshotFromEvent`, `observeRuntimeLimitEvent`, `extractLatestRuntimeLimitSnapshot`, `extractRuntimeLimitSnapshotFromError`, `buildRuntimeLimitCacheSignature`, `buildRuntimeLimitBroadcastCacheKey`, and the bulk of `refreshRuntimeProfileLimitState` exist in both `packages/agent/src/subagentQuery.ts` and `packages/api/src/services/runtime.ts` with nearly identical logic (only the broadcast key shape differs). This is exactly the kind of cross-package drift risk the Runtime Adapter Sync Rule in `CLAUDE.md` warns about. Consider hoisting these into `@aif/runtime` (or `@aif/shared`) so a future fix (e.g. another redaction rule) doesn't need to be applied twice and stay in sync.

2. **URL / email redaction is aggressive.** `VALUE_TOKEN_PATTERNS` treats every `https?://…` and every email as a secret, so legitimate diagnostic text (`"see https://docs.anthropic.com/errors/rate-limit"`, `"contact support@…"`) gets scrubbed from logs too. Defensive by design, but it will make server-side debugging a bit harder — worth documenting or relaxing for the `"internal"` exposure path if you later want richer diagnostics.

3. **Top-level allow-list doesn't extend into nested shapes.** `sanitizeProviderMeta` allow-lists top-level keys, but `sanitizeProviderMetaValue` only denies forbidden keys inside nested objects — anything else survives (subject to depth/size caps). If a new adapter ever nests quota data under a structured object, unexpected fields can still land. Probably fine given the byte cap, but worth a note.

4. **`RuntimeUsageDialog.tsx` is 948 lines in one file.** Per the UI component rules in `CLAUDE.md`, feature-level compositions are fine, but this file is large enough to justify splitting into sub-components (per-provider summary, window list, account grouping) for readability and incremental rendering cost.

5. **`runtimeLimitBroadcastCache` deletion in the failure handler is racy.** In `refreshRuntimeProfileLimitState`, when `notifyProjectRuntimeLimitBroadcast` rejects, the cache entry for that key is deleted; a concurrent caller that set the same signature just before can then re-broadcast the next time. At worst this is one extra broadcast (no correctness impact), but the `.then(…).catch(…)` path could capture the signature it set and only delete when the cache still holds that value.

6. **PR scope.** 99 files and ~15k additions touching agent, api, runtime, shared, web, and docs in a single PR makes reverts / bisects painful. If there's a future pass on anything here, splitting into (a) shared redaction + runtime-limit utils, (b) per-adapter quota/identity integrations, (c) UI surfaces would be easier to review and backport.

### Verdict

LGTM — the stated findings are genuinely closed with corresponding tests, redaction is defense-in-depth, and the gating logic now correctly distinguishes "authoritative block" from "stale/status-only snapshot". The duplication between agent and api is the only thing I'd ask to be followed up on before the next runtime-limit change lands, but it's not a blocker for this PR.

lee-to:
Conflicts after merged #82

lee-to:
## Code review — follow-up

Reviewed the 5 new commits since the previous review ([08:11 UTC 2026-04-21](https://github.com/lee-to/aif-handoff/pull/80#issuecomment-2825)). CI is green (Build / ESLint / Tests / MCP Unit / npm audit).

### Response pass on prior findings

| # | Prior finding | Status | Where |
|---|---|---|---|
| 1 | Duplicate limit-refresh helpers between `agent` and `api` | ✔ Addressed | Hoisted into `packages/runtime/src/limitState.ts`; both consumers now import from `@aif/runtime`. Covered by `packages/runtime/src/__tests__/limitState.test.ts` (113 lines). |
| 2 | `redactProviderText` scrubs URLs/emails too aggressively for logs | ✔ Addressed | New `redactProviderTextForLogs()` in `packages/shared/src/runtimeLimitUtils.ts:303-311` (preserves emails/URLs). Strict `redactProviderText()` stays on client-facing paths. Consumers updated: `subagentQuery.ts`, `stageErrorHandler.ts`, `api/services/runtime.ts`, `api/routes/chat.ts`, codex/openrouter SSE parse. |
| 3 | Top-level allow-list did not extend into nested shapes | ✔ Addressed | `PROVIDER_META_NESTED_SCHEMAS` now enforces per-shape allow-lists for `modelUsageSummary` / `toolUsageSummary` (`runtimeLimitUtils.ts:195-232`). New test at `runtimeLimitUtils.test.ts:209-256`. |
| 4 | `RuntimeUsageDialog.tsx` at 948 lines | ✔ Addressed | Split into `RuntimeUsageDialog.tsx` (~35 lines), `RuntimeUsageEntryCard.tsx` (599), `runtimeUsageDialogModel.ts` (297). Reuses existing `Badge` / `Collapsible` primitives per the UI reuse rule. |
| 5 | Racy broadcast-cache deletion on notify failure | ✔ Addressed | New `clearRuntimeLimitBroadcastCacheKeyIfUnchanged()` in `subagentQuery.ts` only deletes when the cache still holds the exact signature it set. Regression test at `subagentQuery.test.ts:653-729`. |
| 6 | PR scope / revert-ability | ⚠ Unresolved | Not actionable retrospectively; merge with #82 grew scope further. Non-blocking — just note for next runtime-limit pass. |

### Additional delta — strengths

- **[FIX] log-marker leaks scrubbed** across `serverBootstrap.ts`, `data/index.ts`, `vite.config.ts`, and a test assertion. Mojibake `вЂ"` → `—` fixed in `ChatPanel.test.tsx:258` and the `codexCallbackSchema` JSDoc.
- **Dead `runtimeMismatch = false` + its gated alert UI** removed from `ChatPanel.tsx:176` — the gating was always false, so the block was unreachable.
- **N+1 removed in `GET /tasks`**: `toTaskRouteResponse` now takes `systemDefaultRuntimeProfileId` as a parameter so `getAppDefaultRuntimeProfileId("task")` runs once per list call, not per row (`packages/api/src/routes/tasks.ts:48-92`). Covered by the new test `tasks.test.ts:103-147`.
- **Runtime profile list gets a stable, documented `scope=visible`** (default) with `compareVisibleRuntimeProfiles` tiebreaker on `projectId → createdAt → id`, and `docs/api.md` is updated accordingly.
- **`deleteRuntimeProfile` invalidates the cached app-settings row** (`data/index.ts:1619-1620`), so app-default slot clearing via the DB trigger is observed by the next read. Covered by the new test in `runtimeProfiles.test.ts:369-408`.

### New suggestions / non-blocking concerns

1. **`appSettings` module-level cache has a silent-staleness trap.** The cache in `packages/data/src/index.ts:51-52` is invalidated inside `updateAppSettings()` (line 695) and `deleteRuntimeProfile()` (line 1619), but **not** by raw drizzle writes — three test files do `db.update(appSettings).set({ … }).run()` directly (`tasks.test.ts:109`, `tasks.test.ts:585`, `runtimeProfiles.test.ts:456`). Currently safe because those writes happen before any cached read, but a future caller that reads settings during bootstrap and then writes via raw SQL will silently observe a stale row. Consider exporting a `resetAppSettingsCache()` for tests / raw-SQL paths, or keying the cache on `appSettings.updatedAt`, or dropping the cache altogether — a SQLite singleton lookup is already O(1) and the new benchmark gain is mostly from the `toTaskRouteResponse` hoist, not this cache.

2. **Silent behavior change in `sanitizeProviderMetaValue` for unknown nested containers.** Previously, arrays/objects nested under a top-level allow-listed key were walked recursively with only the forbidden-key denylist applied. Now, without an explicit entry in `PROVIDER_META_NESTED_SCHEMAS`, any nested container collapses via `sanitizeOpaqueProviderMetaContainer` into a JSON-stringified blob. Existing top-level keys all carry primitives, so nothing breaks today, but a future adapter emitting a structured object under a new top-level key gets a stringified string rather than a typed object — and the reader has no clear signal that this happened. Worth one line in `docs/providers.md` → "Provider metadata sanitization" stating that any new structured nested shape must register a schema entry to be preserved as a typed object.

3. **`redactProviderText` vs `redactProviderTextForLogs` — guidance missing.** Two redactors with different contracts (strict vs log-friendly) now coexist, but neither has a JSDoc header nor a mention in `docs/providers.md`. Risk: a new adapter author picks `redactProviderText` for server logs and loses URL / email diagnostic context. A short comment on each exported function ("client-safe / strict — use for anything returned to clients" vs "server logs — preserves URLs and emails for diagnostics") would close this.

4. **Mock-drift in `packages/api/src/__tests__/runtimeService.test.ts`.** The new helpers (`buildRuntimeLimitCacheSignature`, `buildRuntimeLimitBroadcastCacheKey`, `observeRuntimeLimitEvent`, `extractLatestRuntimeLimitSnapshot`, `extractRuntimeLimitSnapshotFromError`) are now mocked in both `subagentQuery.test.ts` and `runtimeService.test.ts` with hand-rolled copies of the real implementations. Given they now have direct coverage in `packages/runtime/src/__tests__/limitState.test.ts`, those mocks are added churn that can mask behavior drift — consider unmocking them so integration tests exercise the actual helpers.

### Verdict

**LGTM now** — all five actionable prior findings are closed with matching tests, dev-marker leaks / mojibake have been swept, dead code removed, and the docs/api.md update matches the new `scope=visible` surface. The four new items above are follow-ups, not blockers.

PR #82, opened 2026-04-20
feat: enhance global and project runtime profile management

### LINKED ISSUE #81: [feature] make the runtime global for the project as well
## Problem

Runtimes are currently tied to individual projects. We need to implement the ability to make them global and create global runtimes.

## Proposed Solution

Save agent settings so they don't need to be reconfigured each time a new project is created.

ichinya:
## Summary

This PR adds app-wide runtime defaults and completes the global/project runtime-profile workflow across the API, data layer, agent resolution, and web UI.

Runtime selection now resolves in this order:

`task override -> project default -> app default -> env fallback`

## What changed

- Added persisted app-wide runtime defaults for:
  - task
  - plan
  - review
  - chat
- Wired runtime resolution to use app defaults before environment fallback
- Enforced runtime-profile scope validation across:
  - settings
  - projects
  - tasks
  - chat sessions
- Fixed app default resolution so disabled or missing profiles are not reported as effective defaults
- Restricted app-default selectors to enabled global profiles only
- Sanitized stale disabled app-default selections in the UI before save

## UI updates

- Split global and project runtime-profile management more clearly
- Added `Make Global` for project-scoped profiles
- Added `Copy to Project` for global profiles
- Show global profiles inside project runtime settings
- Added delete action for global profiles from the project screen with explicit confirmation

image

## Why

Previously, global runtime defaults were only partially implemented:
- runtime execution did not consistently consume app defaults
- `/settings` could report disabled profiles as active defaults
- the UI allowed selecting disabled global profiles that the API would reject
- global profiles were not fully manageable from the project context

This PR closes those gaps and makes global profiles reusable and manageable without per-project reconfiguration.

## Verification

- `npm run lint --workspace=@aif/web`
- `npm run test --workspace=@aif/web`
- `npm run ai:validate`

## Tests

Added/updated coverage for:
- app runtime default resolution
- settings/runtime-defaults behavior with disabled profiles
- global runtime default UI sanitization
- project/global runtime scope actions in the web UI

Closed #81


lee-to:
## Code review

Solid PR overall. The app-level runtime default chain (`task override → project → app → env`) is wired end-to-end consistently — same `getAppDefaultRuntimeProfileId(mode)` call threads through `runtimeProfiles.ts`, `tasks.ts`, `chat.ts`, `runtime.ts`, and `subagentQuery.ts`. The scope-validation service and the `trg_runtime_profiles_delete` trigger (now covering `app_settings`) give referential safety without FKs. Test coverage across API / data / UI is comprehensive, all CI checks pass.

A few concrete issues worth addressing before merge:

### Must fix

1. **`[FIX]` debug markers leaked into production strings** — look like TODO tags to anyone reading logs:
   - `packages/data/src/index.ts:711,723,730` — three `[FIX] App runtime default points to …` log lines inside `getAppDefaultRuntimeProfileId`.
   - `packages/web/vite.config.ts` — `throw new Error("[FIX] Failed to load @tailwindcss/vite: ...")`.
   Rephrase as plain warnings/errors.

2. **Dead code + unreachable UI in `ChatPanel.tsx`** — `handleSend` was simplified to always pin (`pinActiveSession()` + `forceNew = false`) but the component still has `const runtimeMismatch = false;` (line 177) and the JSX branch `{runtimeMismatch && (<banner>Runtime changed — next message will start a new session</banner>)}` (lines 314–319). Remove both — they can never fire. The `ChatPanel.test.tsx` assertion that checks this banner is absent is also looking at a mojibake string (`"Runtime changed вЂ" next…"`, CP1251→UTF-8 corruption of the em-dash); it passes only because `queryByText` returns `null` for the garbled string, not because the UI logic works — so the test gives a false sense of coverage.

### Should fix

3. **N+1 on task list** — `toTaskRouteResponse` calls `getAppDefaultRuntimeProfileId("task")` per task, which runs `getAppSettings()` → `ensureAppSettingsRow()` (SELECT + `INSERT OR IGNORE`) every call. `GET /tasks` will do O(N) round-trips for the same singleton. Hoist the app-default lookup out of the `map` in `tasksRouter.get("/")` and pass it in, or memoize per request.

4. **`ensureAppSettingsRow()` runs an idempotent insert on every read** even though migration 13 already seeds row id=1. Safe, but wasteful — consider a once-per-process flag.

### Nits

5. `packages/api/src/schemas.ts` — em-dash (`—`) in the Codex callback comment got converted to an ASCII hyphen in this PR. Looks like an unintentional encoding conversion, not part of the feature.
6. `runtimeProfileListQuerySchema` accepts `scope: "global" | "project" | "visible"` but `docs/api.md` only documents `global`/`project`. Add `visible` (or note it as the implicit default).
7. `RuntimeExecutionContext.selectionSource` gains `"profile_id"` — double-check no downstream switch/`if` branches drop it silently (I scanned the diff only).
8. `resolveAdapterForTask` still types `mode: "task" | "plan" | "review"` (no `chat`), which is fine for current callers but worth a comment since the underlying `getAppDefaultRuntimeProfileId` is wider.

Nice UX touches: the "App default" / "Project default" / "Env fallback" empty-state labels, the `Make Global` / `Copy to Project` pair, and sanitizing stale disabled selections in the dialog before save.

lee-to:
## Follow-up on the response commits

Thanks for the turnaround. Walking through the two new commits:

**Addressed**
- `[FIX]` markers stripped from `packages/data/src/index.ts` (3 warnings) and the `vite.config.ts` tailwind-load error. ✔
- Dead `runtimeMismatch` constant + unreachable banner JSX removed from `ChatPanel.tsx`, stale `AlertTriangle` import cleaned up. ✔
- N+1 fixed in `tasksRouter.get("/")` — `getAppDefaultRuntimeProfileId("task")` is hoisted once and threaded into `toTaskRouteResponse` as a parameter. ✔
- Request-lifetime cache for `app_settings` keyed on the DB instance, with invalidation on `updateAppSettings` and `deleteRuntimeProfile`. Nice touch — covers the "wasteful insert-or-ignore per read" concern and survives `createTestDb()` swaps. ✔
- Em-dash restored in `packages/api/src/schemas.ts`. ✔
- `scope=visible` documented in `docs/api.md` + clarifying comment on `resolveAdapterForTask` about task-only scope. ✔

**New breakage — CI Tests job is red**

The newly added test `packages/api/src/__tests__/runtimeProfiles.test.ts > "lists visible profiles when scope=visible"` is failing:

```
AssertionError: expected [ 'project-visible-profile', …(1) ] to deeply equal [ 'global-visible-profile', …(1) ]
```

Expected ordering is `[global-visible-profile, project-visible-profile]`, actual is `[project-visible-profile, global-visible-profile]`. `listRuntimeProfiles({ projectId, includeGlobal })` doesn't guarantee ordering by scope — the test encodes an assumption the implementation doesn't make.

Pick one:
- Relax the assertion to `expect.arrayContaining([...])` (+ `length` check) — simplest, tests visibility not ordering.
- Or add an explicit `ORDER BY` (e.g., globals first, then project-local) in the list query and keep the assertion — has the side benefit of a stable UI.

Second commit (`f4e0980`) also invalidates the app-settings cache inside `deleteRuntimeProfile`. Since the delete trigger in migration 13 can null out `app_settings.default_*_runtime_profile_id`, the cache invalidation is correct — without it a subsequent `getAppDefaultRuntimeProfileId` could still return the now-nulled id. Good catch.

Once the `scope=visible` test is green this looks ready.

ichinya:
Checked latest commit `a0a0b15`. No findings on my side.

The change makes the `scope=visible` ordering explicit in the API route instead of relying on incidental DB order: globals first, then `createdAt`, with `id` as a deterministic tie-breaker. That matches the test contract and removes the flakiness behind the failing CI assertion.

Re-ran the targeted API test locally:
`npx vitest run --config packages/api/vitest.config.ts --configLoader bundle packages/api/src/__tests__/runtimeProfiles.test.ts`
Result: 19/19 tests passed.

lee-to:
CI is green now. Deterministic ordering (globals first, then `createdAt`, then `id`) is a cleaner fix than relaxing the assertion — stable UI order is a free side benefit. LGTM on my end.

PR #3, opened 2026-04-03
feat(ui): add Label, Checkbox, Switch, Radio form atoms

lee-to:
## Summary
- Add 4 form control atoms: Label, Checkbox, Switch, Radio
- Include tests for all components
- Migrate TaskSettings and AddTaskForm to use new primitives

## Components
- `ui/label.tsx` — Label with required asterisk support
- `ui/checkbox.tsx` — Styled checkbox with forwardRef
- `ui/switch.tsx` — Toggle switch with aria-checked
- `ui/radio.tsx` — Styled radio with forwardRef

PR #84, opened 2026-04-23
feat(sessions): enhance session file handling and optimize limit snapshot

ichinya:
## Summary

Optimizes Codex session scanning for `/runtime-profiles` and `/chat/sessions`.

The previous pipeline applied `limit`/newest-first logic after hydrating metadata from all Codex JSONL session files. With thousands of sessions this caused cold requests to scan hundreds of MiB of session data.

## Changes

- Added newest-first `filePath + mtimeMs + size` discovery before JSONL parsing.
- Pushed `limit` and `projectRoot` filtering down before session meta hydration.
- Added per-file session meta cache keyed by `filePath|mtimeMs|size`.
- Changed latest Codex rate-limit discovery to tail-read JSONL chunks instead of full-file reads.
- Kept compatibility fallback for session-id lookup when filename id does not match `session_meta.id`.
- Added regression tests for lazy scan limiting and tail-read behavior.
- Added AI Factory patch note and branch-scoped implementation plan.

## Validation

- `npm test --workspace=@aif/runtime -- codexSessions`
- `npm run lint --workspace=@aif/runtime`
- `npm run build --workspace=@aif/runtime`
- `npm test --workspace=@aif/api -- runtimeProfiles chatSessions sessionCache`
- `npm run ai:perf`
- `npm run ai:validate`

## Perf Notes

Latest local `ai:perf` sample:

- `chat/sessions`: cold `208.2ms`, warm `12.9ms`
- `runtime-profiles`: cold `688.4ms`, warm `10ms`

Remaining known cost: directory/stat walk is still `O(number of session files)` to sort by `mtime`; JSONL content reads are now bounded/lazy. Persistent sidecar indexing remains the next step if stat-walk cost becomes significant at 10k+ sessions.


ichinya:
Perf after optimization relative to the run with the feature enabled before the fixes:

- `runtime-profiles` cold: ~75% faster
- `runtime-profiles` cold total: ~56% faster
- Dashboard DOMContentLoaded: ~59% faster
- Dashboard FCP: ~51% faster
- Dashboard LCP: ~3% faster
- `chat/sessions` cold: ~17% faster
- `chat/sessions` warm: ~25% faster
- Overall perf run: ~19% faster

ichinya:
Final validation update.

All correctness checks are green:

- `npm test --workspace=@aif/runtime -- codexSessions`: 15/15 passed
- `npm test --workspace=@aif/api -- codexIndex runtimeProfiles`: 40/40 passed
- `npm run test`: passed
- `npm run ai:load`: passed all 3 k6 scripts
- `npm run ai:validate`: passed end-to-end with k6 enabled


npm run ai:load passed all 3 k6 scripts
* chat-sessions: p95 67.73ms, p99 375.84ms
* runtime-profiles: p95 129.11ms, p99 163.01ms
* tasks: p95 918.69ms, p99 933.62ms

ichinya:
Review: Changes requested due to migration conflict, otherwise looks solid.

The performance direction is strong. The PR pushes limit/newest-first logic before JSONL hydration, adds per-file meta caching, tail-reads limit snapshots, and the follow-up chunking fix correctly addresses SQLITE_MAX_VARIABLE_NUMBER for production-size session directories. The reported validation and k6 numbers look good.

Blocking issue:
#84 updates DB/schema version to 16 for Codex index tables, while #85 also uses migration v16 for tasks.branch_name. Please rebase/renumber/merge the migrations so both changes are applied deterministically from pre-v16 DBs. Add a migration regression test covering the final combined schema.

Non-blocking:
Please document the background indexer / kill-switch behavior in release notes, since the remaining directory/stat walk is still O(number of session files) and this runs against large ~/.codex/sessions directories.

Once the migration conflict is resolved, I would approve this PR.

ichinya:
Addressed the review feedback.

- Resolved the migration collision by moving the Codex index read-model migration off `v16` to the next trailing version (`v17`) and updated the DB migration regression expectations.
- Removed sensitive scope identifiers from Codex overlay debug logs: `accountFingerprint`, `projectRoot`, and `headKey` are no longer emitted there. The logs now keep aggregate scope/count/source-selection diagnostics only.
- Added regression coverage for the logging redaction path.

While re-running load checks, `/tasks` exposed a separate performance regression. Root cause was per-task effective runtime resolution plus noisy fallback logging under list load. I fixed that with batch runtime-profile resolution for task lists and corrected fallback logging so normal project defaults are not logged as fallback events.

Validation run:

- `npm test --workspace=@aif/shared -- db`
- `npm test --workspace=@aif/data -- codexLogging`
- `npm test --workspace=@aif/data -- runtimeProfileResolution`
- `npm test --workspace=@aif/api -- tasks`
- `npm run ai:load`
- `npm run ai:validate`

Fresh k6 result: all 3 scenarios passed; `/tasks` p95 is now about `86ms` with `0` failed checks.

ichinya:
Follow-up PR check after addressing the review findings:

- Migration conflict is resolved: the Codex index migration now uses the next trailing version (`v17`) instead of colliding with PR #85's `v16`, and the DB migration expectations were updated.
- Sensitive Codex overlay debug fields are removed: `accountFingerprint`, `projectRoot`, and `headKey` are no longer logged; the diagnostics keep aggregate counts/source-selection data only.
- Added regression coverage for the logging redaction path and runtime-profile batch resolution behavior.
- Fixed the `/tasks` k6 regression found during validation by batching effective runtime-profile resolution for task lists.

Current PR state checked again: mergeable, not draft, and all GitHub checks are green (Build, ESLint, Tests, MCP Unit, mcpunit, npm audit). Latest validation included `npm run ai:load` and `npm run ai:validate`; fresh k6 `/tasks` p95 is about `86ms` with `0` failed checks.

lee-to:
## Code review

Strong perf work. Pushing limit/newest-first filtering before JSONL hydration, adding the SQLite-backed Codex index with idle backfill + non-blocking head warm-up, and switching `/runtime-profiles` to a TTL+generation overlay cache are all the right shapes. CI is green and the previously raised migration collision (`v16` ↔ PR #85) is resolved at `v17` with `CREATE TABLE IF NOT EXISTS` keeping it idempotent.

One must-fix before merge: the `[FIX:…]` developer markers leaked into production log strings.

### Must fix

1. **Drop `[FIX:…]` developer markers from log message strings.** They're development-iteration tags — they leak into production logs and violate the code-cleanliness convention that markers like `[FIX]`/`[TODO]`/`HACK` don't ship in source. The strings carry no operational value beyond the human-readable tail.
   - `packages/api/src/services/codexIndex.ts:498-499` —
     ```ts
     status === "appended"
       ? "DEBUG [FIX:codex-index-append] Parsed Codex appended session range"
       : "DEBUG [FIX:codex-index-full-range] Parsed Codex full session range with cursor state",
     ```
   - `packages/api/src/services/codexIndex.ts:610` — `"DEBUG [FIX:codex-index-cleanup] Deleted stale Codex limit rows for changed files"`
   - `packages/api/src/shutdown.ts:36` — `"DEBUG [FIX:codex-index-shutdown] Codex indexer stopped during shutdown"`
   - `packages/data/src/index.ts:2528` — `"[FIX:tasks-runtime-batch] Resolved effective runtime profiles for task list"`

   Replace each with a plain message (e.g. `"Parsed Codex appended session range"`, `"Resolved effective runtime profiles for task list"`). The accompanying structured payload (`reason`, `status`, counts) already carries the operational signal — no leading `DEBUG ` prefix needed either, since pino renders the level itself.

### Should fix

1. **`head` mode still walks the full sessions directory before slicing to `headFileLimit`.** `listCodexSessionFileInfos({ limitNewest })` calls `listSessionFileInfos(SESSIONS_DIR)` (recursive readdir + stat per file), then `files.slice(0, limitNewest)` — `packages/runtime/src/adapters/codex/sessions.ts:533-540`. On a 6k-rollout dir that's still thousands of `stat` calls per warm-up tick, undermining the `headFileLimit=200` budget. Either push the cap into `listSessionFileInfos` (heap-K maintained during traversal) or sort lazily by directory mtime first and only stat the candidates that survive the cap.

2. **`idx_codex_session_files_dirty` is created but appears unused.** `packages/shared/src/db.ts:898` adds `idx_codex_session_files_dirty ON codex_session_files(missing, mtime_ms, size_bytes)`, but no query in `@aif/data` orders/filters by that composite (`listCodexSessionFileStates` does `desc(updatedAt)`, `…ByPaths` is `inArray(filePath, …)`). Either drop the index or wire the dirty-scan path that motivated it.

3. **`loadIndexedCodexRuntimeMessages` drops the adapter when mapping events.** `packages/api/src/routes/chat.ts` calls `mapRuntimeEventsToChatMessages(runtimeEvents, input.chatSessionId)` (no adapter), while the adapter-backed branches call it with the adapter so `extractMessageContent(content, adapter)` can run. For Codex events this currently round-trips through `extractMessageContent(rawContent, undefined)` — works because Codex JSONL events already carry plain strings, but it's an undocumented divergence. Pass the resolved adapter through, or add a `// codex events ship as plain strings` comment so the next reader doesn't trip on it.

### Nits

- `docs/architecture.md` enumerates the new `idx_codex_sessions_*` and `idx_codex_limit_*` indexes but skips `idx_codex_session_files_session_id` and `idx_codex_session_files_dirty`. If the dirty one survives Should-Fix #2, list both for completeness.
- `pruneCodexLimitHistoryRetention(maxRowsPerHead)` is still exported but the indexer now calls only `pruneCodexLimitHistoryByHead` per touched head. The whole-table sweep is now dead-on-the-hot-path; either add a callsite (e.g. a cold cron sweep) or mark the export as legacy/test-only so it doesn't get re-introduced as default behaviour.
- Test naming nit: `packages/data/src/__tests__/index.test.ts:1175` uses `describe("codex index repositories", …)` with `expect(sessionChanges).toBeGreaterThan(0)` rather than asserting the exact `1`. Tighter assertions catch silent over-write regressions.

### Context gates

- **Architecture:** pass — `api` and `agent` continue to access DB only via `@aif/data`; new `codexIndex` service in `packages/api/src/services/` correctly uses the data-layer repository. Indexer runs out-of-band of request handlers, matching the documented goal.
- **Rules (`.ai-factory/RULES.md`):** pass — migration is append-only at `v17` (no v16 reuse), no `as T` nullable strips, no string-based error classification added.
- **Roadmap:** warn — this is perf work without an explicit milestone in `.ai-factory/ROADMAP.md`. Non-blocking, but worth a one-line milestone if Codex indexing is intended to be a stable contract.
- **CHECKLIST compliance:** pass — touched packages are `runtime`, `data`, `api`, `shared`; no per-package `CHECKLIST.md` files exist in the tree, root checklist (no string-based error classification, package CHECKLIST upkeep) is honoured.
- **Docs sync:** pass with one nit — `AGENTS.md`, `docs/api.md`, `docs/architecture.md`, `docs/configuration.md`, `docs/providers.md` all updated; index list in `docs/architecture.md` omits two indexes (see Nits).

### Positive notes

- Migration collision handling is exemplary: append-only at `v17`, idempotent SQL, both fresh-DB (`ensureTables`) and migration paths covered, version assertion bumped in `db.test.ts`.
- `codexLogging.test.ts` actively asserts that overlay debug payloads do **not** contain `accountFingerprint` / `projectRoot` / `headKey` — that's the right way to enforce log hygiene under refactor. (Once the `[FIX:…]` markers are dropped, consider an analogous test asserting log strings don't match `/\[FIX:|\[TODO|FIXME/` to keep this from regressing.)
- The SQLite variable-limit fix (PR #5: chunk every bulk write to stay under `SQLITE_MAX_VARIABLE_NUMBER=999`) is a great catch — the commit message even shows the math (14 cols × 50 rows = 700 < 999). Saves a second incident.
- Generation-counter + TTL on `codexOverlayCache` (`invalidateCodexOverlayCache` bumps `currentGeneration`) is a cleaner cache-bust than a Map.clear race.

Once the `[FIX:…]` markers are stripped this is good to merge.

ichinya:
﻿Addressed the remaining review feedback in `9817ed4`.

What changed:
- Removed `[FIX:...]` developer markers and redundant `DEBUG ` prefixes from production log message strings in the Codex indexer, graceful shutdown handler, and task runtime-profile batch resolver.
- Kept the structured log payloads intact (`reason`, `status`, counts, etc.), so the operational signal remains in fields rather than message text.
- Added regression assertions around the affected log messages so temporary developer markers do not leak back into these paths.

Also covered by the earlier PR updates:
- Full/rewrite Codex parses preserve `pendingTail`/`parsedOffset` through the append parser path.
- Missing/rewritten files remove stale `codex_limit_heads` and history rows.
- Shutdown awaits `codexIndexService.stop()` before closing the server/exiting.
- Deletion-only limit-head reconciles notify visible project runtime profiles.

Validation:
- `npm test --workspace=@aif/api -- codexIndex shutdown`
- `npm test --workspace=@aif/data -- runtimeProfileResolution`
- `npm run lint`
- DB-boundary scan for direct DB imports in api/runtime/agent: no matches
- `npm run ai:validate` passed on fresh rerun

lee-to:
## Local perf reproduction (HEAD `9817ed4`)

Re-ran the suite on a different host (Mac Pro, `~/.codex/sessions` = 6034 files / 528 MB, `AIF_USAGE_LIMITS_ENABLED=true`) to cross-check the author's numbers under a different baseline.

### k6 — 20 VU × 20s, all thresholds passed

| endpoint | p95 | p90 | avg | max | fail |
|---|---|---|---|---|---|
| `chat/sessions` | 42.5 ms | 41.3 ms | 32.7 ms | 2544 ms | 0% |
| `runtime-profiles` | 47.2 ms | 45.4 ms | 35.4 ms | 72 ms | 0% |
| `tasks` | 160 ms | 96.3 ms | 88.9 ms | 844 ms | 0% |

### Playwright cold/warm

| endpoint | cold | warm | coldTotal |
|---|---|---|---|
| `chat/sessions` | 354.7 ms | 5.5 ms | — |
| `runtime-profiles` | 24.8 ms | 6.1 ms | 273 ms |

Dashboard: FCP 416 ms / LCP 696 ms / DOMContentLoaded 268 ms.

### vs prior baseline (flag=true, pre-PR, recorded 2026-04-23)

- `runtime-profiles` cold: **6794 ms → 24.8 ms** (~273× faster).
- `runtime-profiles` under load (p95): from ~7 s baseline → 47 ms.

The SQLite-index + overlay-cache path is doing exactly what it advertised on this host too.

### Differences from the PR's reported numbers

- `chat/sessions` cold 354 ms vs author's 208 ms — likely because backfill hasn't covered all 6034 jsonl yet on first request (the head warm-up loop only scans `headFileLimit=200` newest per tick, see Should-Fix #1).
- `runtime-profiles` cold 24.8 ms vs author's 688 ms — indexer was already warm by the time the cold timer started here (head warm-up had ticked several times before the test).
- `tasks` p95 160 ms vs author's 86 ms — different hardware + larger task table here. Still well below the 1200 ms threshold.

### Side-effect worth flagging

With 6034 jsonl on disk, head warm-up is firing every `idleRetryMs=250ms` because each tick reports `truncated`/`skippedForLoad` and reschedules itself. Per-tick log shows `scannedFiles: 200, durationMs: 220-260ms`. Result: a steady background load of ~220 ms of stat+SQLite work every 0.25 s until backfill catches up. That ratifies Should-Fix #1 — `listCodexSessionFileInfos` still stats all 6034 files before slicing to `limitNewest=200`, so the head budget is bypassed in practice.

Not a blocker for this PR. Recommend a follow-up that pushes the cap into the directory walk (e.g. heap-K maintained during recursion, or sort-then-stat-only-survivors).

### Verdict

Approve stands. Numbers reproduce the win on a second machine; perf claims hold.

ichinya:
I fixed all 4 review findings and added a Codex usage scan limit based on the current limit window.

 What has changed:
- Full parse now goes through the same append-range parser and saves `parsedOffset` / `pendingTail` so that the trailing JSONL fragment is not lost.
- Missing/rewrite cleanup removes stale `codex_sessions`, `codex_session_files`, `codex_limit_heads`, and `codex_limit_history`.
- Deletion-only limit updates now also disable runtime-limit overlays via broadcast.
- Shutdown waits for `codexIndexService.stop()` before closing the server.
- Backfill only scans Codex session files that are less than 7 days old, cleans up the old usage/read-model, but retains old file lookups if they are associated with saved web chats.

 Checks:
- `npm test --workspace=@aif/runtime -- codexSessions`
- `npm test --workspace=@aif/data -- index`
- `npm test --workspace=@aif/api -- codexIndex shutdown`
- `npm run lint`
- `npm run ai:validate`
- `npm run ai:load` / k6: `chat-sessions` p95 40.67ms, `runtime-profiles` p95 164.92ms, `tasks` p95 93.04ms, failures 0%.

lee-to:
## Code review — follow-up #2 (commit `9269b27`)

### Response pass — author's claim "all 4 fixed"

| Item | Status | Notes |
|---|---|---|
| Must-fix #1 (`[FIX:…]` markers) | ✔ addressed in `9817ed4` | Verified previously |
| Should-fix #1 (head walks all jsonl) | ✔ **fixed** | `listSessionFileInfos` now takes `modifiedAfterMs`; `shouldSkipSessionDirectory()` skips entire `YYYY/MM/DD` subdirs whose end-of-day < cutoff before any `readdir`/`stat`. Per-file mtime check filters survivors. Confirmed empirically: head-warmup `scannedFiles: 200 → 19` on our 6034-file corpus, durationMs ~158 ms. |
| Should-fix #2 (unused `idx_codex_session_files_dirty`) | ✖ **not fixed** | Index still defined at `packages/shared/src/db.ts:899`. The `missing=true` placeholder write path was actually *removed* in this delta (the indexer now hard-deletes via `deleteCodexSessionFilesByFilePaths` instead of upserting `missing: true`), so the column the index leads on is **never written `true` anywhere** — the index is now strictly worse than before, not just unused. Drop the index. |
| Should-fix #3 (`mapRuntimeEventsToChatMessages` adapter) | ✖ **not fixed** | `packages/api/src/routes/chat.ts:351` still calls `mapRuntimeEventsToChatMessages(runtimeEvents, input.chatSessionId)` (no `adapter`). Not blocking — works because Codex JSONL events ship plain strings — but the divergence still needs a one-line comment if you want to leave the runtime arg off. |

So 1 of 3 deferred should-fix items resolved + 1 new feature (usage scan window). Two non-blockers remain.

### New feature — 7-day usage scan window

`usageScanWindowDays = 7` default. Backfill scans only files with `mtime` in the window; pruning sweeps:
- `pruneCodexLimitRowsBeforeObservedAt(observedBeforeIso)` — strips heads/history rows older than cutoff via lexicographic ISO comparison.
- `pruneStaleCodexSessionIndexRows({ mtimeBeforeMs })` — strips session/file rows whose `mtimeMs < cutoff` **but retains rows whose `sessionId` is referenced by `chat_sessions.runtimeSessionId` / `agentSessionId`**.

The linked-retain logic is the right call — it keeps saved chats viewable indefinitely while letting orphaned indexed sessions age out.

### Concerns to flag (not blockers)

1. **`pruneStaleCodexSessionIndexRows` reads the entire `chat_sessions` table** every backfill tick to build `linkedSessionIds`. Constant scan, no `WHERE` predicate. At small chat-history scale fine; once chat_sessions grows past O(10k) it's wasted I/O. Consider a `LEFT JOIN` against `chat_sessions` keyed on `runtime_session_id` instead (or invert the lookup: select stale sessions whose `session_id NOT IN (SELECT runtime_session_id FROM chat_sessions)`).

2. **`pruneCodexLimitRowsBeforeObservedAt` is three separate statements** (select scopes / delete heads / delete history). Not wrapped in a transaction — a crash between the head delete and history delete leaves orphan history rows referencing a deleted head. Trivial to fix with `getDb().transaction(...)`.

3. **One-time data loss on upgrade.** First backfill after merge will prune all un-linked indexed Codex sessions older than 7 days. JSONL is still source of truth and the adapter fallback (`findSessionFileInfoById`) still works for legacy files, so no functional loss — but the index will rebuild over the rolling 7-day window only. Worth a one-liner in the PR description so users aren't surprised.

4. **`shouldSkipSessionDirectory` regex** `^|/)(\d{4})/(\d{2})/(\d{2})$` matches only at path tail. Codex layout is `~/.codex/sessions/YYYY/MM/DD/*.jsonl` so the test targets the leaf dir, which is exactly right. Sibling dirs (e.g. flat-layout exports) won't be filtered, but that's intentional — falls through to per-file mtime check.

### Local perf reproduction (Mac Pro, `~/.codex/sessions` = 6034 files / 528 MB, flag=true)

**k6 (20 VU × 20 s):**

| endpoint | p95 | p90 | avg | vs `9817ed4` |
|---|---|---|---|---|
| `chat/sessions` | 43.1 ms | 42.0 ms | 32.1 ms | ≈ same |
| `runtime-profiles` | 49.0 ms | 47.0 ms | 35.9 ms | ≈ same |
| `tasks` | 160 ms | 95.7 ms | 90.7 ms | ≈ same |

All thresholds passed, 0% failure.

**Playwright cold/warm:**

| endpoint | cold | warm | coldTotal |
|---|---|---|---|
| `chat/sessions` | 360.4 ms | 5.8 ms | — |
| `runtime-profiles` | 17.1 ms | 9.0 ms | 132 ms |

Dashboard: FCP 420 ms / LCP 688 ms / DCL 278 ms.

**Background load improvement:** head warm-up now scans 19 files (not 200) per tick on this corpus, completing in ~158 ms before truncation kicks in. The 250 ms retry-loop hot-spin from the prior commit is gone.

vs author's reported `runtime-profiles` p95 = 164.9 ms — ours is ~3× lower; their machine looks slower but the relative shape (improvement vs baseline) holds.

### Verdict

**Approve maintained.** Should-fix #1 resolved with measurable effect; #2 and #3 deferred with this PR is fine but should land as a follow-up — particularly #2, which is now strictly dead code rather than just unused. New 7-day window is a thoughtful addition with the right linked-chat retention policy.

Open follow-ups for:
- Drop `idx_codex_session_files_dirty` (Should-fix #2)
- Pass adapter through `mapRuntimeEventsToChatMessages` for Codex path or document why it's unnecessary (Should-fix #3)
- Wrap `pruneCodexLimitRowsBeforeObservedAt` in a transaction
- Replace full `chat_sessions` scan in `pruneStaleCodexSessionIndexRows` with a join/anti-join

lee-to:
## Bug report — `AIF_USAGE_LIMITS_ENABLED=false` is not respected end-to-end

While reproducing this PR locally a tester reported tasks being repeatedly forced into `blocked_external` with this log line:

```
Coordinator pre-start runtime gate: <scope> limit still blocked (hint=snapshot_reset_at)
```

…even though `AIF_USAGE_LIMITS_ENABLED=false` was set. The only workaround was to clear `runtime_profiles.runtime_limit_snapshot_json` manually via DB.

I traced the path. **Not introduced by this PR — pre-dates it (commit `0535345`)** — but flagging here because the indexer warm-up that this PR un-broke (PR #5, chunked SQLite writes) makes the bug observable in practice on machines that previously crashed before reaching subagent execution.

### Confirmed code path with flag=false

1. `packages/agent/src/subagentQuery.ts:20-24` imports `observeRuntimeLimitEvent`, `extractLatestRuntimeLimitSnapshot`, `extractRuntimeLimitSnapshotFromError` **directly from `@aif/runtime`** — i.e. the raw, unflagged versions. The gated wrappers live only in `packages/api/src/services/runtime.ts:57-76` and are not used by the agent.
2. `refreshRuntimeProfileLimitState()` (`subagentQuery.ts:123`) does **not** check `AIF_USAGE_LIMITS_ENABLED`. After every subagent run it calls `persistRuntimeProfileLimitSnapshot(...)` (success path: `subagentQuery.ts:906`; error path: `subagentQuery.ts:969`).
3. Result: `runtime_profiles.runtime_limit_snapshot_json` keeps getting written even with the master flag off — typically with `status="warning"` or `status="blocked"` and a future `resetAt` parsed from a Codex `token_count.rate_limits` event.
4. On the next coordinator tick, `coordinator.ts:873` calls `evaluateRuntimeLimitGate(profile)` **also unconditionally**.
5. `evaluateRuntimeLimitGate` (`packages/data/src/index.ts:2249`) does **not** check the flag either. It reads the persisted snapshot, sees `status=blocked` + future `resetAt`, returns `{ blocked: true, reason: "provider_blocked", futureHint.source: "snapshot_reset_at" }`.
6. `proactivelyBlockTaskForRuntimeGate` → CAS the task into `blocked_external` and emit the offending log line.

So with flag=false the system still:
- writes runtime-limit snapshots from agent runs;
- gates new tasks against those persisted snapshots;
- forces a manual DB reset to recover.

The doc claim in `docs/configuration.md` for this flag (added in this PR) says the opposite:

> the runtime service skips `observeRuntimeLimitEvent` / `extractLatestRuntimeLimitSnapshot` / `extractRuntimeLimitSnapshotFromError` during every runtime run; … Persisted snapshots already in the DB are still returned on read but never updated.

That contract is honored by the **API** runtime service (`runtime.ts:57-76` wrappers) but violated by the **agent** runtime path because it bypasses those wrappers.

### Suggested fix (single-line, blocks all callers)

The narrowest correct fix is in `evaluateRuntimeLimitGate` itself:

```ts
// packages/data/src/index.ts:2249
export function evaluateRuntimeLimitGate(
  profile: RuntimeProfile | null | undefined,
  nowMs = Date.now(),
): RuntimeLimitGateDecision {
  if (!getEnv().AIF_USAGE_LIMITS_ENABLED) {
    return {
      blocked: false,
      reason: "none",
      runtimeProfileId: profile?.id ?? null,
      snapshot: null,
      futureHint: resolveRuntimeLimitFutureHint(null, { nowMs }),
      violatedWindow: null,
      signature: null,
    };
  }
  // existing body unchanged
}
```

That alone unblocks the user-visible symptom and matches the documented contract. It also keeps the gate decision the single source of truth — coordinator and any future callers stay correct without each adding their own flag check.

### Belt-and-braces (recommended follow-up)

To match the docs "never updated" promise, also stop the agent path from writing:

```ts
// packages/agent/src/subagentQuery.ts:123
function refreshRuntimeProfileLimitState(input: { ... }): void {
  if (!getEnv().AIF_USAGE_LIMITS_ENABLED) return;
  // existing body
}
```

Or — cleaner — re-export the gated wrappers from `@aif/agent` (or move them to `@aif/shared`/`@aif/runtime` as the canonical entry points) and have both `packages/api/src/services/runtime.ts` and `packages/agent/src/subagentQuery.ts` consume those instead of the raw `@aif/runtime` exports. That removes the "two parallel call sites, one of them gated" failure mode permanently.

### Impact / severity

- User-visible: task forced into `blocked_external` after every Codex run that emits `rate_limits` warnings, even when the operator explicitly opted out via `AIF_USAGE_LIMITS_ENABLED=false`. Manual `UPDATE runtime_profiles SET runtime_limit_snapshot_json = NULL` required to unblock.
- Not a regression introduced by this PR.
- But this PR makes the bug observable on real hardware: before, the indexer warm-up crashed (see PR #5 commit message) and many users never reached the subagent path; now the indexer survives warm-up and tasks actually run, exposing the gate.

Recommend either:
1. Land the one-line fix in this PR (cheap, low risk, low diff), or
2. Spin a follow-up PR with the gate-level fix + agent re-export refactor.

Repro on your side: set `AIF_USAGE_LIMITS_ENABLED=false`, run any task that goes through Codex SDK/CLI long enough to emit a `token_count` event with `rate_limits` (e.g. a Spark-pool session near the 7d window), watch `runtime_profiles.runtime_limit_snapshot_json` populate and the next coordinator pick get blocked.

ichinya:
Update on the follow-up review items:

I addressed the remaining Codex index and usage-limit issues:

- `AIF_USAGE_LIMITS_ENABLED=false` now short-circuits the data-layer runtime-limit gate and the agent-side event/error snapshot parsing and persistence paths.
- Stale Codex index cleanup now preserves rows linked to saved web chats while avoiding a full `chat_sessions` scan.
- Old Codex limit pruning is now transactional.
- The unused `idx_codex_session_files_dirty` index is dropped via migration v18 and is no longer recreated.
- Added a short comment in the indexed Codex chat path explaining why adapter-specific message extraction is not needed there.

Validation passed:

- `npm test --workspace=@aif/data -- runtimeProfiles index`
- `npm test --workspace=@aif/shared -- db`
- `npm test --workspace=@aif/agent -- subagentQuery`
- `npm test --workspace=@aif/api -- chat`
- `npm run coverage --workspace=@aif/agent`
- `npm run lint`
- `git diff --check`
- `npm run ai:validate` (including all k6 load/perf scripts)

lee-to:
## Code review — follow-up #3 (commit `2738d7c`)

### Response pass — all five items from the previous review

| Item | Status | Notes |
|---|---|---|
| Bug: `AIF_USAGE_LIMITS_ENABLED=false` not respected end-to-end | ✔ **fixed** | `evaluateRuntimeLimitGate` short-circuits at the top (`packages/data/src/index.ts:2245-2257`), returning `{ blocked: false, reason: "none", snapshot: null, … }` whenever the flag is off. Belt-and-braces also landed: `subagentQuery.ts:700` reads the flag once per call and gates **all four** write/parse sites — `observeRuntimeLimitEvent` on event stream, `extractLatestRuntimeLimitSnapshot` + `refreshRuntimeProfileLimitState` on success, and `extractRuntimeLimitSnapshotFromError` + `refreshRuntimeProfileLimitState` on error. So with flag=false the agent path no longer writes snapshots, and the gate ignores any already-persisted ones. Good single-read-at-init shape too — no `process.env.FOO` re-reads in hot loops. |
| Should-fix #2: drop `idx_codex_session_files_dirty` | ✔ **fixed** | Migration `v18` (`packages/shared/src/db.ts:634`) issues `DROP INDEX IF EXISTS idx_codex_session_files_dirty`, idempotent. `ensureIndexes` no longer recreates it. Migration version assertion bumped to 18. New regression tests cover both the upgrade path (v17 DB with index → v18 without it) and fresh-DB creation (index absent on first init). Append-only migration rule honored — v18 is a new trailing slot, no in-place edits to merged migrations. |
| Should-fix #3: `mapRuntimeEventsToChatMessages` adapter divergence | ✔ **addressed** | Two-line comment added at `packages/api/src/routes/chat.ts:351` explaining indexed Codex JSONL events ship as plain strings, so adapter-specific extraction is not needed there. Acceptable resolution. |
| Concern #1: `pruneStaleCodexSessionIndexRows` full `chat_sessions` scan | ✔ **fixed** | Replaced full-table scan + JS `Set` build with bounded `EXISTS` / `NOT EXISTS` subqueries against `chat_sessions` keyed on `runtimeSessionId` / `agentSessionId` (`packages/data/src/index.ts:3105-3163`). The DB plans this as an indexed nested-loop on the small set of stale candidates, not an O(chat_sessions) sweep. |
| Concern #2: `pruneCodexLimitRowsBeforeObservedAt` non-transactional | ✔ **fixed** | All three statements (select scopes / delete heads / delete history) now run inside `getDb().transaction((tx) => …)` (`packages/data/src/index.ts:3245-3267`). No more orphan-history window on crash. |

CI green on all six checks (Build, ESLint, MCP Unit, npm audit, Tests, mcpunit).

### Nit (non-blocking, follow-up)

- **`linkedPaths` filter is now redundant.** In the rewritten `pruneStaleCodexSessionIndexRows`, `staleSessionRows` and `staleFileRows` are already constrained by `NOT EXISTS (SELECT 1 FROM chat_sessions …)`, so the resulting `candidatePaths` set cannot contain any path also present in `linkedPaths`. The `[...candidatePaths].filter((p) => !linkedPaths.has(p))` at `packages/data/src/index.ts:3168` is therefore a no-op for any input. `retainedLinkedSessionIds` is still load-bearing (it feeds `linkedRowsRetained` into the log payload + return value), so don't drop the linked queries — but the filter line can go, or the two `linkedPaths` adds can go. Pure cleanliness, no behavior change either way.

### Context gates

- **Architecture:** pass — `agent` continues to consume DB only through `@aif/data` (gate decision is data-layer); `subagentQuery` reads `getEnv()` from `@aif/shared` not from a runtime/adapter import.
- **Rules (`.ai-factory/RULES.md`):** pass — migration is append-only at `v18`, no `as T` nullable strips introduced, no string-based error classification, the new flag check is on a structured env var.
- **Roadmap:** n/a — same perf milestone as before, no new linkage required.
- **CHECKLIST compliance:** pass — touched packages (`agent`, `data`, `shared`, `api`); root checklist rules honored.
- **Docs sync:** pass — `docs/configuration.md` already documents the `AIF_USAGE_LIMITS_ENABLED` contract; this commit makes the agent path actually honor it. Doc claim and code now match.
- **PR size:** the new commit alone is +210 / -89 across 7 files — small and focused on the surfaced concerns. The PR's overall size is still large but every part has been reviewed across the three previous passes.
- **Risk gating (feature flag):** pass — `AIF_USAGE_LIMITS_ENABLED` defaults to off, single read at function entry, off-path is the existing pre-PR behavior. Naming (`AIF_<AREA>_<FEATURE>_ENABLED`) matches the convention.

### Positive notes

- Test coverage on the flag tracks the implementation closely — both the agent path (`subagentQuery.test.ts`: rate_limit_event yielded, asserted no persist/clear/notify mocks fire) and the gate (`runtimeProfiles.test.ts`: pre-persisted blocked snapshot, flag flipped, gate decision asserted as `{ blocked: false, reason: "none", snapshot: null }`). Good belt-and-braces — either side regressing alone gets caught.
- Migration v18 is the right textbook fix for "shipped an unused index": append a new migration that drops it, remove the recreate, regression-test both paths. No editing of v16/v17 slots.
- Wrapping the limit-prune in a transaction is the kind of small-surface fix that prevents a class of "almost-impossible-to-repro" reports later. Worth doing even though the prior race was narrow.

### Verdict

**Approve.** Every must-fix and should-fix from the three prior review passes is now resolved (5 of 5), the new contract for `AIF_USAGE_LIMITS_ENABLED=false` is end-to-end consistent (gate, agent event path, agent error path, success-path persist), and CI is green. The remaining `linkedPaths` filter nit is cosmetic and fine to land on a follow-up.

LGTM.

PR #85, opened 2026-04-24
fix(agent): enforce feature branch creation in HANDOFF_MODE auto-queue

### LINKED ISSUE #83: Auto-Queue/HANDOFF_MODE=1 игнорирует git.create_branches и не создаёт feature ветки
## Description

При включённой настройке `Create feature branches` / `git.create_branches: true` задачи, запущенные через Auto-Queue (`HANDOFF_MODE=1`), всё равно выполняются в текущем worktree без создания отдельной feature branch или worktree.

В результате несколько автономных задач могут писать изменения в одно и то же рабочее дерево, а их diff'ы смешиваются до коммита или ручной модерации.

## Steps to Reproduce

1. Включить Create feature branches в настройках проекта.
2. Включить Auto-Queue Mode.
3. Создать задачу с plannerMode=full.
4. Дать Handoff выполнить planner + implementer в autonomous-режиме (HANDOFF_MODE=1).
5. Проверить сгенерированный plan и состояние git/worktree.

## Expected Behavior

Если в конфиге проекта указано:

```yaml
git:
  enabled: true
  create_branches: true
  branch_prefix: feature/
```
и задача выполняется через Auto-Queue / autonomous Handoff, то система должна:

- создавать и переключаться на отдельную feature branch или отдельный worktree до начала имплементации;
- не запускать следующую queued-задачу в том же mutable workspace, если предыдущая задача ещё не финализирована;
- вести себя в соответствии со смыслом настройки Create feature branches.

## Actual Behavior

В Auto-Queue / HANDOFF_MODE=1:

- в плане branch только предлагается, например: feature/... (suggested, not created by autonomous Handoff planning);
- имплементация продолжается в текущем workspace;
- следующая queued-задача может стартовать до полной финализации предыдущей задачи, например пока предыдущая задача находится на ручной модерации / review;
- из-за этого изменения нескольких задач могут оказаться в одном worktree и потом смешаться в одном коммите.

## Environment

- OS: Ubuntu 24.04
- Node.js version: v24.13.1
- npm version: 11.12.1

lee-to:
Fixes #83
Upstream skill-level companion: lee-to/ai-factory#96

## Summary

- `HANDOFF_MODE=1` planner/implementer previously delegated branch creation to the `aif-plan` skill and `plan-polisher` subagent, which skipped Step 1.4 whenever `HANDOFF_MODE=1` was read as "no interactive prompts". Autonomous tasks kept running on whatever branch was checked out, so queued tasks stacked diffs into the same worktree.
- Move branch creation into deterministic agent code so the behavior no longer depends on LLM obedience.
- Implementer restores HEAD to `task.branchName` before running, recovering from any branch drift between planner and implementer stages.

## Changes

- **new** `packages/agent/src/gitBranch.ts` — `ensureFeatureBranch()` helper, slug/branch-name derivation, git ops via `execFileSync`.
- `packages/agent/src/subagents/planner.ts` — create/switch branch before subagent run when `plannerMode="full"` and `!isFix`; persist `task.branchName`.
- `packages/agent/src/subagents/implementer.ts` — `ensureFeatureBranch({switchOnly:true})` when `task.branchName` present.
- `packages/shared/src/schema.ts` + migration **v16** — append-only `branch_name TEXT` column on tasks.
- `packages/shared/src/types.ts` — `Task.branchName: string | null`.
- Test fixtures in `web/` and `shared/` updated for the new required field.

## Branch creation guards (all must be true)

1. `plannerMode === "full"` — fast mode stays on current branch by design.
2. `!task.isFix` — fix flow uses a dedicated plan file and does not branch.
3. `git.enabled === true` in `.ai-factory/config.yaml`.
4. `projectRoot` is inside a git work tree.
5. `git.create_branches === true`.

Any false → `action: "skipped"`, HEAD untouched, `task.branchName` stays `null`, and implementer's switch step no-ops.

## Known gap (follow-up)

Parallel auto-queue (`parallelEnabled=true` + `autoQueueMode=true`) still shares a single worktree across concurrent tasks — branch switches race between Task A's implementer and Task B's planner. Sequential auto-queue is correct after this PR. The parallel case needs either a guard rejecting the combination or real `git worktree`-per-task scheduling. Tracked separately.

## Test plan

- [x] `packages/agent/src/__tests__/gitBranch.test.ts` (11 new) — slug/branch derivation, skip conditions, create + switch paths, idempotency, `branch_prefix` config.
- [x] `planner.test.ts` (+2) — creates `feature/…` branch in full mode + git repo; stays on `main` in fast mode.
- [x] `implementer.test.ts` (+2) — switches HEAD to `task.branchName`; does not touch HEAD when `branchName` unset.
- [x] `npm run ai:validate` — format, lint, full test suite, coverage, build, Playwright perf, k6 load — all green.


lee-to:
Addressed review blockers in 8c02394.

## Summary of changes

1. **Implementer ordering** — `ensureFeatureBranch({switchOnly:true})` now runs at the very top of `runImplementer`, before `getProjectConfig`, `readCanonicalPlan`, `computePendingPlanLayers`, and the no-op early return. Stale plan on the wrong branch can no longer trip the "all done → skip" path.

2. **BranchIsolationError → blocked_external** — new typed error (`dirty_worktree`, `branch_missing`, `branch_drift`, `base_branch_unavailable`, `checkout_failed`, `create_failed`). `stageErrorHandler.classifyStageError` matches it first, pins the task to `blocked_external` with `retryAfter=null`, never falls into generic revert/retry.

3. **`switchOnly` no-create** — implementer/plan-checker/reviewer call with `switchOnly:true`; if the branch is missing the function throws `branch_missing` instead of silently recreating on top of whatever HEAD happens to be.

4. **Dirty-worktree hard gate**
   - `ensureFeatureBranch` calls `assertWorkingTreeClean` before any switch/create.
   - `processAutoQueueAdvance` calls `describeDirtyWorkingTree` per project before advancing the next backlog task — closes the sequential-auto-queue hole where `done` / `manualReviewRequired` can be terminal with uncommitted work.

5. **Strict base branch** — when base branch is missing or `git checkout <base>` fails, throws `base_branch_unavailable`. No silent creation from an unrelated HEAD.

6. **Stage-wide isolation** — `runPlanChecker` and `runReviewer` now restore `task.branchName` with the same `switchOnly` contract as implementer. Isolation is no longer implementer-only.

7. **Post-subagent drift assertion** — `runPlanner` calls `assertCurrentBranch(projectRoot, preparedBranch)` after `executeSubagentQuery` returns. Detects the case where `plan-polisher` or the skill silently switched branches mid-run.

8. **`HANDOFF_BRANCH_PREPARED` contract** — planner injects these lines into the subagent prompt so `ai-factory`'s `aif-plan` / `plan-polisher` can validate the Handoff-prepared branch instead of creating a second one:
   ```
   HANDOFF_MODE: 1
   HANDOFF_TASK_ID: <uuid>
   HANDOFF_BRANCH_PREPARED: 1
   HANDOFF_BRANCH_NAME: feature/<slug>-<short-id>
   ```
   Contract detailed in https://github.com/lee-to/ai-factory/issues/96#issuecomment-4314078941.

## Tests

Added under `packages/agent/src/__tests__/`:

- `gitBranch.test.ts`: dirty throws, switchOnly no-create, branch_missing, branch_drift, base_branch_unavailable.
- `implementer.test.ts`: branch restore BEFORE no-op (pending task still triggers subagent), throws on missing branch.
- `planner.test.ts`: drift throw when subagent switches HEAD, `HANDOFF_BRANCH_PREPARED` injection.
- `stageErrorHandler.test.ts`: BranchIsolationError → blocked_external for every kind, retryCount unchanged, cause-chain classification.
- `autoQueue.test.ts`: dirty-worktree pauses advance, clean worktree advances.

`npm run ai:validate` — format, lint, full test suite (+22 new), coverage, build, Playwright perf, k6 load — green.

## What's intentionally NOT in this PR

- **Parallel auto-queue + worktrees** — tracked in #86. Out of scope here.
- **Repo-level mutex beyond tasks** (chat/manual actions writing to `projectRoot`) — noted in review; out of scope for #85.
- **Surfacing `branchName` in web UI** — small follow-up; not a correctness blocker.

lee-to:
Second review pass addressed in 4adcc79.

## Blocker coverage

**1. Post-run branch assertion in every stage**
`assertCurrentBranch(projectRoot, task.branchName)` now fires after the subagent call in implementer, plan-checker, and reviewer (planner already had it). Any mid-run `git checkout` by a sidecar/skill is caught before plan/log/review state is persisted.

**2. `ai-factory#96` ownership contract aligned**
Issue body rewritten: skill must skip its own `git checkout -b` when `HANDOFF_BRANCH_PREPARED=1` and validate HEAD against `HANDOFF_BRANCH_NAME` instead. Old Step 1.4 path retained as fallback for `HANDOFF_MODE=1` without prepared-branch context (older Handoff clients).

**3. `task.branchName` is source-of-truth**
New `restorePersistedBranch()` helper — mandatory contract, never skips. Config drift (`git.enabled` / `create_branches` toggled off between stages) no longer lets the stage run on current HEAD. Two new kinds cover this: `git_disabled_with_persisted_branch`, `not_a_repo_with_persisted_branch`. Implementer / plan-checker / reviewer switched from `ensureFeatureBranch({switchOnly:true})` to `restorePersistedBranch`.

**4. Branch-name validation**
New `validateBranchName()` — calls `git check-ref-format --branch`, rejects empty strings, leading/trailing slashes, double slashes, Git-special refspecs like `@{-1}`. Produces `invalid_branch_name` errors up front instead of opaque `checkout_failed`/`create_failed` mid-flow. Called from both `ensureFeatureBranch` (new branches, including misconfigured `branch_prefix=""`) and `restorePersistedBranch` (stored-value corruption).

**5. `accept_existing_plan` bypass closed**
`handleAcceptExistingPlan` now calls `ensureFeatureBranch` and persists `task.branchName` before transitioning `backlog → plan_ready`. Fix tasks keep the legacy no-branch path. Closes the back-door route into the autonomous pipeline where the planner stage was skipped entirely.

**6. `ai-factory#96` body updated**
Done separately via `gh issue edit` — contract block added, dangerous "MUST execute `git checkout -b`" wording replaced.

## Medium / follow-ups

- **Parallel auto-queue** — already serialized for `create_branches` projects in the previous pass (`projectUsesSharedBranchIsolation` in `processAutoQueueAdvance`). Full parallel isolation via worktrees stays as #86.
- **`git pull --ff-only` policy** — still best-effort-warn; making it `BranchIsolationError("base_update_failed")` is low-risk but separate — can file a follow-up if you want that gated behind `git.update_base_before_branch`.
- **Dirty-gate UX** — `describeDirtyWorkingTree` already truncates to 5 paths + "+N more" in the activity message. Good enough for now.

## Tests

`gitBranch.test.ts` (+9):
- `validateBranchName` rejects empty / leading-trailing slashes / double slashes / `@{-1}`
- `validateBranchName` accepts conventional names
- `ensureFeatureBranch` throws `invalid_branch_name` when `branch_prefix=""`
- `restorePersistedBranch` throws `git_disabled_with_persisted_branch`
- `restorePersistedBranch` throws `branch_missing` when branch deleted
- `restorePersistedBranch` switches + is idempotent
- `restorePersistedBranch` throws `dirty_worktree`

`implementer.test.ts` (+1):
- throws `branch_drift` when subagent switches HEAD mid-run

`api/tasks.test.ts` (+1):
- `accept_existing_plan` creates + persists feature branch when project has `create_branches=true` and is a real git repo; HEAD ends up on the task's feature branch.

`npm run ai:validate` — format, lint, full test suite, coverage, build, Playwright perf, k6 load — all green.

ichinya:
Cross-PR blocker: #84 and #85 both claim schema/database version 16 for different migrations. Please rebase whichever lands second and bump/merge the migration version so both `tasks.branch_name` and Codex index tables are applied deterministically. Add an upgrade test from pre-v16 to the final combined schema.

ichinya:
Review: Changes requested, but this is much closer.

The follow-up commits fixed most of the previous branch-isolation blockers: restore is now before config/plan/no-op reads, task.branchName is treated as a source-of-truth via restorePersistedBranch, invalid branch names are validated with git check-ref-format, switchOnly no longer creates missing branches, accept_existing_plan now binds/restores a branch, and shared-worktree auto-queue is serialized.

Remaining blockers before merge:

1. Public Tests are still red. The current check shows npm run coverage failing with a RuntimeProfileForm.test.tsx timeout. Please rerun/fix until the PR is green.

2. Add post-run branch assertions. Planner asserts branch drift after the subagent returns, but implementer still appears to restore only before execution. If a subagent switches branch during implementation or checklist sync, the stage can persist plan/log/state on the wrong branch. Please assert currentBranch === task.branchName after implementation/checklist sync and before persistTaskPlanForTask / setTaskFields. Apply the same wrapper to planChecker/reviewer/commit/fast-fix mutation paths.

3. Resolve the schema version conflict with #84. #85 uses migration v16 for tasks.branch_name, while #84 also updates database version to 16 for Codex index tables. Whichever PR lands second must rebase/bump/merge migrations and add an upgrade regression test.

4. Please clarify git pull policy. `git pull --ff-only origin <base>` currently logs and continues from local base on failure. That may be acceptable, but it should be explicit configuration/documentation, or a strict BranchIsolationError when enabled.

After these fixes, I would be comfortable treating #85 as a strong sequential auto-queue fix plus safe parallel mitigation. #86 should remain open for real per-task worktrees.

lee-to:
Third review pass addressed in a89dcb0.

## Blocker coverage

**1. Post-run branch assertions on every mutation path**

Audited every `executeSubagentQuery` / `runApiRuntimeOneShot` callsite. Coverage now:

| Stage | Restore (pre-run) | Assert (post-run, before persist) |
| --- | --- | --- |
| planner | `ensureFeatureBranch` | `assertCurrentBranch` (already in place) |
| plan-checker | `restorePersistedBranch` | `assertCurrentBranch` (already in place) |
| implementer | `restorePersistedBranch` | `assertCurrentBranch` after main run **AND** after `runChecklistSyncQuery` (new — covers the case where the checklist-sync subagent itself drifts HEAD) |
| reviewer | `restorePersistedBranch` | `assertCurrentBranch` (already in place) |
| commit (`runCommitQuery`) | `ensureFeatureBranch({switchOnly:true})` | `assertCurrentBranch` after `runApiRuntimeOneShot` (new — drift returns `ok:false` instead of silently committing on wrong HEAD) |
| fast-fix (`handleFastFix`) | `restorePersistedBranch` (was `ensureFeatureBranch`) | `assertTaskBranchPostRun` before `persistTaskPlanForTask` (new) |
| approve_done / start_ai (regular transition) | `restorePersistedBranch` (was `ensureFeatureBranch`) | n/a (no subagent runs; only DB write + plan-file delete) |
| accept_existing_plan | `restorePersistedBranch` (was `ensureFeatureBranch`) on the already-bound path | n/a (no subagent runs) |

`restoreTaskBranchForMutation` in `packages/api/src/services/taskEvents.ts` now uses `restorePersistedBranch` so config drift (`git.enabled` / `create_branches` toggled off after planner) cannot release any API-side mutation to current HEAD. `task.branchName` is enforced as a source-of-truth contract on every mutation path, not just the agent stages.

**2. `git pull --ff-only` policy explicit**

New `git.strict_base_update` field in `.ai-factory/config.yaml` (default `false`):

- `false` (default) — failed pull is logged + branching continues from local base. Same behavior as before, now an explicit documented policy.
- `true` — failed pull is converted into `BranchIsolationError("base_update_failed")` (new `kind` on the error union). Stage classifier already routes any `BranchIsolationError` to `blocked_external` with `retryAfter=null`, so a stale base parks the task for an operator instead of silently building on top of an out-of-sync base.

The setting only gates the upfront pull — once a branch exists, a stale local base does not retroactively block subsequent stages.

Documented in `docs/configuration.md` → "`strict_base_update` semantics".

**3. Schema v16 conflict regression test**

I cannot deterministically resolve the cross-PR conflict here (we don't know which of #84 / #85 lands first), so I added the regression test the review asked for instead:

- New `db.test.ts` case: pre-v16 schema (with all v6→v15 columns), `pragma user_version = 15`, run `getDb()`, assert `branch_name` was added AND every prior column survived AND `user_version === 16`.
- If #84 lands first with a different v16 SQL body, this test will fail when this PR rebases — that's the forcing function the review asked for. The rebaser must move this PR's migration to the next free trailing slot (per the `MIGRATIONS` append-only rule in `CLAUDE.md`) rather than silently re-using v16.
- Same logic applies in reverse if #85 lands first.

**4. Tests green**

Re-ran `npm run coverage` from root and from `packages/web/`: 7/7 in `RuntimeProfileForm.test.tsx`, no timeouts. CI on this branch was green on the last push too — I think the previous timeout was a flake. I did bump the drift test in `planner.test.ts` to a 20s `it()` timeout because git-init + checkout chains under parallel vitest workers were occasionally crossing the 5s default.

`npm run ai:validate` (format, lint, full test suite + 5 new, coverage, build, Playwright perf, k6 load) — green.

## Test deltas

`packages/shared/src/__tests__/gitIsolation.test.ts` (+2):
- default policy warns + continues when `git pull --ff-only` fails
- `strict_base_update=true` throws `BranchIsolationError("base_update_failed")` on failed pull

`packages/shared/src/__tests__/db.test.ts` (+1):
- v15 → v16: `branch_name` added, prior columns intact, `user_version === 16`

`packages/api/src/__tests__/commitGeneration.test.ts` (+1):
- subagent switches HEAD mid-run → `runCommitQuery` returns `ok:false` with `Branch drift detected` error, post-run assertion is invoked AFTER the runtime call

`packages/agent/src/__tests__/planner.test.ts`:
- drift test `it(..., { timeout: 20_000 }, ...)` to absorb parallel-worker variance

## Out of scope, deferred

- **Real per-task worktrees** — still in #86. This PR is sequential auto-queue + safe parallel mitigation only.
- **Repo-level mutex for chat/manual actions** — every mutation path now calls `restorePersistedBranch`, but a true cross-process lock is a separate change.

ichinya:
# What's already good after the updates

1. **`#96` is now consistent with runtime ownership.**
   In issue `ai-factory#96`, the wording has already been updated: after `#85`, Handoff owns branch creation; if there is `HANDOFF_BRANCH_PREPARED=1`, the skill should validate `HANDOFF_BRANCH_NAME` instead of creating a second branch. This addresses my previous risk of "double-branch behavior" at the contract level.

2. **Most of the previous blockers are closed.**
   The author's update directly lists: post-run branch assertion for implementer / plan-checker / reviewer, `task.branchName` as a source of truth, explicit branch validation through `git check-ref-format`, branch binding in `accept-existing-plan`, drift tests, config drift, invalid branch names, and migration.

3. **The CI overall looks green.**
   Tests, Build, Lint, and MCP Unit are currently in a successful state. But there is a strangeness with Security Audit: the workflow page shows Success, but inside the annotations there is “1 error” and `npm audit` with exit code 1. I wouldn’t call it a blocker if branch protection considers check green, but I would ask the author to clarify/clean the annotation so that there is no false signal in the security gate.

4. **Parallel is now safer, but not solved.**
   Coordinator serializes branch-isolated shared-worktree projects and adds dirty-worktree gate before auto-queue advance. This is a good mitigation. But `#86` is still the real full fix: per-task `git worktree`, `tasks.worktree_path`, and all stages should run from the task-specific cwd.

# Remaining logical blockers

## Blocker 1. `commitGeneration` still uses `ensureFeatureBranch` instead of `restorePersistedBranch`

This is the most serious remaining issue.

 In `commitGeneration.ts`, the commit runtime currently uses:

```ts
ensureFeatureBranch({
  explicitBranchName: task.branchName,
  switchOnly: true,
})
```

instead of `restorePersistedBranch`. The problem is that `ensureFeatureBranch()` can still return `skipped`, for example when `git.enabled=false` or if projectRoot is no longer recognized as a git repo. This is fine for initial branch provisioning, but **it is not allowed for a task with a persisted `branchName`.**

 A dangerous scenario:

```text
1. Planner created feature/task-abc and recorded task.branchName.
2. Before commit, the stage config changes: git.enabled=false.
3. Commit stage calls ensureFeatureBranch(... switchOnly=true).
4. Helper returns skipped.
5. Commit runtime runs on the current branch.
6. Post-run assert can catch drift, but the commit may already be created elsewhere.
```

**Fix:** The commit path should only use `restorePersistedBranch(projectRoot, task.branchName, ...)`, as in implementer/reviewer/planChecker. Post-run assert is useful, but it shouldn't be the first point where we find out that a commit has already occurred on the wrong branch.


## Blocker 2. Planner re-run with an existing `task.branchName` should also use `restorePersistedBranch`

In `planner.ts`, the current logic is as follows: if `plannerMode === "full"` and `!task.isFix`, then `ensureFeatureBranch({ explicitBranchName: task.branchName ?? null }) is called`; if the branch is prepared, the prompt is injected with `HANDOFF_BRANCH_PREPARED` and `HANDOFF_BRANCH_NAME`, and after the subagent run, there is an `assertCurrentBranch`. This is good for a new task, but it is not ideal for a replan/retry task that already has a persisted `task.branchName`.

 The problem is the same: the persisted branch should be the source of truth. If the `task.branchName` already exists, the planner should not go through the create/provision helper. It should do mandatory restore.

 Dangerous scenarios:

```text
1. task.branchName already exists, but config drift made git.enabled=false.
2. ensureFeatureBranch can skip.
3. Planner still runs subagent and injects HANDOFF_BRANCH_PREPARED.
4. Post-run assert only catches the problem after plan/subagent could already write to the wrong place.
```

Or:

```text
1. task.branchName already exists, but the branch has been deleted.
2. ensureFeatureBranch without switchOnly may attempt to recreate the branch.
3. This is no longer a restore of a persisted branch, but a silent re-creation from the current/base HEAD.
```

**Fix:** in the planner:

```text
if task.branchName exists:
  restorePersistedBranch(...)
else:
  ensureFeatureBranch(...)
```

And only after successful restore/provision, run the subagent.


## Blocker 3. `accept_existing_plan` reads the plan before restoring for an already-bound task

`accept_existing_plan` has improved: it now knows how to bind/create branch for task without `branchName` and restore branch for task with `branchName`. But the order is still dangerous for an already-bound task: the code first resolves/reads the plan file, and only then does the branch binding / restore.

Dangerous scenario:

```text
1. Task already has task.branchName = feature/task-a.
2. Current HEAD drifted to main or feature/task-b.
3. User/API calls accept_existing_plan.
4. Code reads plan file from current wrong branch.
5. Then restore’s feature/task-a.
6. Persist’s plan content read from wrong branch.
```

**Fix:** if `task.branchName` already exists, restore should be **before** `resolvePlanFilePath()` / `readFile()`. For a task without a branchName, the current logic of “reading the existing plan and then copying/binding to the new branch” may be intentional, but it should be explicitly left only for unbound path and preferably commented.

## Blocker 4. `strict_base_update` does not work if HEAD is already on the base branch

The author added a good policy-flag `git.strict_base_update`: if enabled, failure in `git pull --ff-only origin <base>` should become `BranchIsolationError("base_update_failed")`; if disabled, you can warning-and-continue from local base. This is the correct setting.

But in `ensureFeatureBranch()` pull is now inside the branch:

```text
if current !== config.base_branch:
  checkout base
  git pull --ff-only origin base
```

If the current branch is **already** `base_branch`, the code does not pull at all, even with `strict_base_update=true`. This means that the policy “strictly update base before creating feature branch” only works in one of two cases.

 A dangerous scenario:

```text
git.strict_base_update=true
current branch = main
origin/main is not available or pull would fail
ensureFeatureBranch creates feature branch from stale local main
```

**Fix:** first ensure current branch is base, then regardless of whether checkout was performed, execute optional pull/update policy:

```text
if current !== base:
  checkout base

if update/pull policy enabled:
   pull --ff-only origin base
  if failed and strict_base_update:
    BranchIsolationError("base_update_failed")

checkout -b feature/...
```

Add test: `current === base_branch`, `strict_base_update=true`, mocked pull failure → `base_update_failed`.


## Blocker 5. Merge-order risk with `#84` still needs to be explicitly committed

Previously, there was a direct conflict with v16. Now, according to the update, `#84` has already moved the Codex index migration off v16 to the next trailing version v17. This is a good thing. But the merge-order still needs to be controlled: `#85` uses v16 for `tasks.branch_name`; if `#84` with v17 is merged **before** `#85`, then the production DB, which has already been upgraded to v17, may not execute the late-added migration v16 from `#85` without rebase/renumbering.

**Fix / merge instruction:**

```text
Either merge #85 before #84,
or if #84 lands first, rebase #84 and move branch_name migration to the next version.
```

# Minor / non-blocking

## Security Audit annotation looks contradictory

Security Audit check page shows Status Success, but annotations contain error on `npm audit` with exit code 1. If this is the expected “advisory only” logic, it is better to clean/rename the step to avoid red noise. If this should be a blocking audit, then the current Success is misleading.

## `#86` should be left open

`#85` now serializes shared branch isolation properly, but this is not true parallel isolation. For true parallel mode, you still need a worktree PR: `tasks.worktree_path`, `git worktree add`, all stages through task-specific cwd, cleanup/recovery/UI.


# Briefly

Review: Changes requested, but this is very close.

The latest updates fixed most of the previous branch-isolation blockers. The direction is correct: Handoff runtime owns branch creation/restoration, task.branchName is persisted, branch isolation failures are typed, branch names are validated with git check-ref-format, dirty worktree is fail-closed, post-run drift checks were added for the main stages, accept_existing_plan now binds/restores branches, and shared-worktree auto-queue is serialized.

The updated ai-factory#96 contract also looks right: when HANDOFF_BRANCH_PREPARED=1, the skill should validate the provided branch instead of creating another one.

I still see a few correctness issues before merge:

1. commitGeneration still uses ensureFeatureBranch(... switchOnly=true) instead of restorePersistedBranch for tasks with task.branchName. ensureFeatureBranch can return skipped for git.disabled / non-git repo, which means commit runtime may run on the current branch. The post-run assertion is too late here because the commit may already have happened. For any persisted task.branchName, restore must be mandatory and fail-closed.

2. Planner re-runs with an existing task.branchName should also use restorePersistedBranch, not ensureFeatureBranch. If task.branchName exists, the branch is already the source of truth; the planner should not skip due to config drift or recreate a missing branch. Use ensureFeatureBranch only for first-time branch provisioning.

3. accept_existing_plan still appears to read the plan file before restoring an already-bound task branch. For tasks with task.branchName, restore should happen before resolving/reading the plan file; otherwise the event can copy a plan from the currently checked-out wrong branch and then persist it onto the bound branch.

4. strict_base_update does not appear to apply when current branch is already base_branch. The pull/update policy is inside the “current !== base_branch” branch, so strict_base_update=true can still create a feature branch from stale local base if HEAD is already on base. Please run the base update policy after ensuring the base branch is checked out, regardless of whether a checkout was needed.

5. Merge-order note with #84: #84 moved its Codex index migration to v17, while #85 still owns v16 for tasks.branch_name. That is fine only if #85 lands before #84. If #84 lands first, #85 must rebase and move branch_name migration to the next version, otherwise already-upgraded DBs can skip the late v16 migration.

Minor:
- Security Audit is marked Success, but the annotations still show an npm audit error/exit code 1. Please clarify whether that is expected advisory-only behavior or clean up the step output.
- #86 should remain open: this PR serializes shared-worktree branch-isolated auto-queue, which is a good mitigation, but true parallel correctness still needs per-task git worktrees.

After the four code fixes above plus the migration merge-order clarification, I would be comfortable approving this as a strong sequential auto-queue branch-isolation fix and a safe parallel mitigation.

lee-to:
Fourth review pass addressed in c821f6f.

## Blocker coverage

**1. `commitGeneration` → `restorePersistedBranch`**

Switched the bound-task path in `runCommitQuery` from `ensureFeatureBranch({switchOnly:true})` to `restorePersistedBranch`. `ensureFeatureBranch` could return `skipped` on `git.enabled=false` / non-git projectRoot and let the commit subagent run on whatever HEAD happened to be — the post-run `assertCurrentBranch` catches the drift but only after the commit may already have landed. `restorePersistedBranch` is fail-closed: `git_disabled_with_persisted_branch` / `not_a_repo_with_persisted_branch` / `branch_missing` all throw before any runtime call.

Tests updated:
- `restores persisted task branch via restorePersistedBranch before commit runtime starts` — verifies call order + arg shape.
- `returns ok:false fail-closed when restorePersistedBranch throws` — verifies the runtime is never invoked when restore fails (e.g. config drift).

**2. Planner restore vs provision split**

`packages/agent/src/subagents/planner.ts:157` — branch handling now splits on `task.branchName`:

```ts
if (task.branchName) {
  restorePersistedBranch({ projectRoot, taskId, persistedBranchName: task.branchName });
  preparedBranch = task.branchName;
} else {
  const branchResult = ensureFeatureBranch({ projectRoot, taskId, title: task.title });
  // ... persist branchResult.branchName on the task
}
```

`ensureFeatureBranch` is now only called for first-time provisioning. Replan / retry runs (where the planner already persisted a branch) go through `restorePersistedBranch`, which:
- throws `git_disabled_with_persisted_branch` if config drifted to `git.enabled=false`
- throws `branch_missing` instead of silently re-creating a deleted branch from base
- throws `dirty_worktree` instead of clobbering uncommitted state

The post-run `assertCurrentBranch(projectRoot, preparedBranch)` is unchanged.

**3. `accept_existing_plan` reorders restore-before-read**

`packages/api/src/services/taskEvents.ts:handleAcceptExistingPlan` — branch handling moved to BEFORE `resolve(planFilePath)` / `readFileSync`. Three paths:

- `task.branchName` set → `restoreTaskBranchForMutation` (which now uses `restorePersistedBranch`) BEFORE the plan read.
- No `task.branchName` and not a fix → `ensureFeatureBranch` to provision, then plan read happens on the new branch.
- Fix tasks → unchanged (no branch).

Closes the dangerous case where the plan file was read from a drifted HEAD and then persisted onto the bound branch.

**4. `strict_base_update` on `current === base_branch`**

`packages/shared/src/gitIsolation.ts:ensureFeatureBranch` — hoisted the pull out of the `current !== config.base_branch` branch:

```ts
// Step 1: ensure HEAD is on base (no-op if already there)
if (current !== config.base_branch) { /* checkout base */ }

// Step 2: ALWAYS run the pull policy
const pullResult = runGit(projectRoot, ["pull", "--ff-only", "origin", config.base_branch], ...);
if (pullResult.status !== 0) {
  if (config.strict_base_update) throw new BranchIsolationError("base_update_failed", ...);
  log.warn(...);
}

// Step 3: checkout -b feature/...
```

This closes the bypass: `strict_base_update=true` now fires whether HEAD started on a topic branch or already on stale local `base_branch`.

Test added: `throws base_update_failed when HEAD is already on base and strict_base_update=true` — leaves HEAD on `main`, no `topic-*` branch first, asserts `BranchIsolationError("base_update_failed")`.

**5. Merge-order with #84**

Acknowledged. #85 keeps v16 for `tasks.branch_name`; if #84 (v17) lands first this PR must rebase v16 → v18 (or whatever the next free slot is) per the append-only `MIGRATIONS` rule in CLAUDE.md. The v15→v16 upgrade test added in the previous round will fail at rebase if a future v16 in main has different SQL, forcing the rebase.

## Minor

- **Security Audit annotation:** the `npm audit` step exits non-zero advisory-only, but the GitHub Actions step returns 0 to avoid blocking the check. The contradictory rendering is a `gh pr checks` reporting quirk on advisory-only steps. I'd rather track that as a separate cleanup commit since it touches CI workflow files outside the branch-isolation scope of #85 — happy to file a follow-up issue.
- **#86** — agreed, stays open for real per-task `git worktree`. This PR remains sequential auto-queue + safe parallel mitigation.

## Validation

`npm run ai:validate` (format, lint, full test suite + 2 new, coverage, build, Playwright perf, k6 load) — green end-to-end.

Tangential: bumped `packages/agent/vitest.config.ts` `testTimeout` to 20s. Several agent tests do real git init/checkout/commit chains in tmp work-trees and were flaking under parallel vitest workers with the 5s default; deterministic non-git tests still finish in <100ms so the larger budget only matters on the slow path.

ichinya:
Most of the previous critical issues have been resolved: `commitGeneration` now uses `restorePersistedBranch`, `accept_existing_plan` now handles branch handling before reading the plan, `strict_base_update` has been added, post-run branch assertions have been added to the main stages, `task.branchName` has become the source-of-truth for restore, branch names are validated using `git check-ref-format`, and `ai-factory#96` now correctly states “validate Handoff-prepared branch, do not create another branch”. The author's third pass directly lists coverage for planner / plan-checker / implementer / reviewer / commit / fast-fix / approve_done / accept_existing_plan, as well as a new strict pull policy and migration regression test.

CI for the main checks now looks green: `Tests`, `ESLint`, and `Build` are successful on the latest commit. However, `Security Audit` looks contradictory: the job summary shows success, but the annotations still contain an `npm audit` error / exit code 1. I wouldn't make it a code-blocker if branch protection considers check green, but I would ask the author to clarify whether this is an advisory-only behavior or noise in the workflow.

# Remaining logical issues

## Blocker 1. Parallel serialization is still config-based, not task-state-based

Currently, `projectUsesSharedBranchIsolation(projectRoot)` returns true only if the current config says `git.enabled && git.create_branches && isGitRepo(projectRoot)`. The coordinator uses this predicate to forcefully serialize `parallelEnabled` projects with shared branch isolation.

Issue: persisted `task.branchName` is already the source-of-truth for task restore, and `restorePersistedBranch()` continues to be mandatory for already-bound tasks even if `create_branches` is later turned off. However, scheduler serialization may be turned off because `projectUsesSharedBranchIsolation()` becomes false. This results in a possible config-drift scenario:

```text
1. Project: parallelEnabled=true, autoQueueMode=true, git.create_branches=true.
2. Planner creates a branch for Task A and saves the task.branchName.
3. Until the next stages, the user/setting turns off git.create_branches.
4. restorePersistedBranch will still switch HEAD for Task A.
5. But the coordinator no longer considers the project branch-isolated and can again give parallel concurrency.
6. Multiple branch-bound tasks again mutate the same shared worktree.
```

This actually returns some of the risk from #86: HEAD-only switching in a single shared working directory remains unsafe for parallel execution. #86 itself explicitly states that the true parallel fix requires either reject/serialization or per-task `git worktree`.

**What to fix:** serialization predicate should consider not only the current config, but also active/in-flight tasks with non-null `branchName`. Options:

```text
Option A:
serialize project if git.create_branches=true OR any active/pipeline task has branchName.

Option B:
reject config changes that disable create_branches while branch-bound tasks are active.

Option C:
persist project/task isolation mode at task creation time and schedule by that persisted mode.
```

I would choose at least Option A up to #86.

## Blocker 2. Planner restore for existing `task.branchName` is still gated by `plannerMode === "full"`

In the current `planner.ts` branch block is inside:

```text
if (plannerMode === "full" && !task.isFix) {
  if (task.branchName) restorePersistedBranch(...)
  else ensureFeatureBranch(...)
}
```

This is correct for **first-time branch provisioning**: a new branch can only be created in full mode. However, this is not ideal for **already-bound task**: if the task already has a `task.branchName`, restore should be mandatory regardless of the current `plannerMode`. Otherwise, a replan/mode-drift scenario is possible:

```text
1. Task already has a branchName = feature/task-a.
2. Due to retry/replan/manual action, the plannerMode becomes fast or the stage is not running in full.
3. Planner doesn’t restore’s branch because block is gated by full.
4. Plan generation may work on current branch.
5. Later stages restore’s feature/task-a, but plan might not be created/read there.
```

If `task.branchName` is source-of-truth, restore should happen **before mode-specific planning**. Only new branch creation should remain `full`-mode-only. ([GitHub][6])

**What to fix:**

```ts
if (!task.isFix && task.branchName) {
  preparedBranch = restorePersistedBranch(...);
} else if (!task.isFix && plannerMode === "full") {
  preparedBranch = ensureFeatureBranch(...);
}
```

And add regression test: task with existing `branchName`, plannerMode `fast`, current branch drifted → planner should restore or block, but not work on current branch.

# Medium / merge notes

## 1. Merge-order with #84 still needs to be explicitly committed

`#84` has already moved the Codex index migration to v17, so the direct "both use v16" is no longer relevant in the current state. But merge-order is still important: if `#85` with `branch_name` v16 comes first, and `#84` v17 comes later, everything is fine. If `#84` v17 comes first, then `#85` cannot be merged as “late v16”; it must be rebased and moved to the next migration version.

Ready note:

```text
Merge #85 before #84, or rebase/renumber #85 if #84 lands first.
```

## 2. Security Audit annotation needs to be cleaned or explained

The main checks are green, but the audit job page shows success along with an annotation about exit code 1. This creates confusion for reviewers/security gates.

```text
If npm audit is advisory-only, please make the workflow output clearly advisory.
If it is intended to block, then the job should not report Success.
```

## 3. #86 should remain open

`#85` now serializes shared-worktree branch isolation well, but this is a mitigation, not true parallel correctness. `#86` is still needed for `tasks.worktree_path`, `git worktree add`, all stages running from task-specific cwd, cleanup/recovery/UI.

lee-to:
Fifth review pass addressed in b745cce.

## Blocker coverage

**1. Task-state-based serialization (Option A)**

Implemented Option A from the review. Both scheduler entry points (`processAutoQueueAdvance` and `resolveProjectConcurrency`) now combine config AND task state:

```ts
const usesSharedBranchIsolation =
  projectUsesSharedBranchIsolation(project.rootPath) ||
  hasActiveBranchBoundTasksForProject(project.id);
```

Toggling `git.create_branches=false` mid-pipeline can no longer open the parallel pool while `restorePersistedBranch` is still mutating the shared work tree on stage transitions. Both predicates are pessimistic — once any branch-bound task is in flight (or queued bound from `accept_existing_plan`), the project is treated as shared until those tasks reach a terminal state.

New data-layer query `hasActiveBranchBoundTasksForProject(projectId)` in `packages/data/src/index.ts`:

- Returns true if any task with `branchName IS NOT NULL` is in `backlog | planning | plan_ready | implementing | review | blocked_external`.
- Includes `backlog` because `accept_existing_plan` / replan can leave a branch-bound task in backlog briefly; the scheduler must already serialize, otherwise the pool opens for one tick before the next stage start.
- Includes `blocked_external` so retry-after cycles don't leak parallel concurrency.
- Excludes terminal states (`done`, `verified`, `paused`).

This is meant as an Option A bridge until #86 lands real per-task worktrees. After #86 the predicate flips: `worktree_path` becomes the source of truth, and a project with shared `rootPath` simply has no in-flight tasks pointing at it.

**2. Planner restore unconditional for bound tasks**

`packages/agent/src/subagents/planner.ts:164` — restore is no longer gated by `plannerMode === "full"`:

```ts
if (!task.isFix && task.branchName) {
  restorePersistedBranch({ projectRoot, taskId, persistedBranchName: task.branchName });
  preparedBranch = task.branchName;
} else if (!task.isFix && plannerMode === "full") {
  // first-time provisioning only
  const branchResult = ensureFeatureBranch({ projectRoot, taskId, title: task.title });
  ...
}
```

Two contracts, applied in order:

1. RESTORE for ANY bound non-fix task — runs in both full and fast mode. `task.branchName` is the source of truth; once persisted, every subsequent stage MUST land on it or fail loud.
2. CREATE only in full mode for unbound non-fix tasks. Fast mode stays on the current branch by design (aif-handoff#83) — first-time branch provisioning is full-mode-only.

A fast-mode replan / comment-driven re-run of an already-bound task can no longer skip the restore.

**3. Merge-order with #84**

Acknowledged as a merge-time concern, not a code fix:

> Merge #85 before #84, or rebase/renumber #85 if #84 lands first.

The v15→v16 upgrade test added in round 3 is the forcing function — if a future v16 in main has different SQL, the test will fail at rebase and force the migration to a free trailing slot per the append-only `MIGRATIONS` rule.

**4. Security Audit annotation**

Tracked separately from #85 since it touches CI workflow files outside the branch-isolation scope. Will file a follow-up to either:
- Make the `npm audit` step explicitly advisory (`continue-on-error: true` with a clear rendered status), or
- Make it blocking and fix the underlying advisory.

Not gating this PR on CI plumbing cleanup.

**5. #86 stays open** — agreed. This PR is sequential auto-queue + safe-parallel mitigation. Real parallel correctness needs `tasks.worktree_path` + `git worktree add` per #86.

## Tests

`packages/data/src/__tests__/index.test.ts` (+4):
- `hasActiveBranchBoundTasksForProject` returns false when no task has a branchName.
- True once a branch-bound task is in flight (`implementing`).
- True for a queued `backlog` task already bound (`accept_existing_plan` / replan path).
- False when bound tasks are terminal (`done`).

`packages/agent/src/__tests__/planner.test.ts` (+1):
- `restores persisted branch in fast plannerMode for already-bound task (mode-drift safe)` — task with `plannerMode=fast` and `branchName=feature/bound-fast`, drifted HEAD on `main` → after `runPlanner`, HEAD ends on `feature/bound-fast`.

`npm run ai:validate` — format, lint, full test suite (+5 new), coverage, build, Playwright perf, k6 load — green end-to-end.

ichinya:
Review: Approve with merge notes.

I rechecked the latest round and the previous correctness blockers are now addressed.

What looks good:
- task.branchName is now treated as the source of truth via restorePersistedBranch.
- Branch isolation failures are fail-closed typed BranchIsolationError cases.
- Branch names are validated through git check-ref-format.
- accept_existing_plan handles branch binding/restoration before reading the plan.
- commitGeneration restores the persisted branch before commit runtime and asserts after runtime.
- planner restores already-bound non-fix tasks regardless of plannerMode, while first-time branch creation remains full-mode-only.
- implementer/plan-checker/reviewer have restore-before-run and post-run branch drift checks.
- strict_base_update now applies even when HEAD already starts on base_branch.
- shared-worktree parallel execution is serialized not only by current config, but also by active branch-bound task state. This closes the config-drift gap until #86 adds real per-task worktrees.
- ai-factory#96 is now aligned with this PR: when HANDOFF_BRANCH_PREPARED=1, the skill validates the Handoff branch instead of creating another branch.

Merge notes:
1. Merge-order with #84 still matters. #85 owns v16 for tasks.branch_name. #84 now uses v17. If #85 lands first, this is fine. If #84 lands first, rebase/renumber #85 to the next free migration slot.
2. The Security Audit annotation is confusing but appears advisory-only per the author’s note. I would not block this PR on CI wording cleanup, but it should get a follow-up.
3. Keep #86 open. This PR is a strong sequential auto-queue fix and safe shared-worktree parallel mitigation; true parallel correctness still requires per-task git worktrees.

No remaining code blockers from me.

PR #94, opened 2026-04-28
Add Codex App Server transport support

ichinya:
## Summary

Adds `app-server` as a Codex runtime transport backed by `codex app-server` stdio JSONL RPC.

This gives AIF a local Codex execution path with structured streaming events, resumable thread IDs, runtime-aware session discovery, and transport-specific model discovery while keeping the existing SDK, CLI, and API transports intact.

Hardening commit `49134ef` (`fix(runtime): harden codex app-server transport`) makes the PR safer for review by tightening App Server env filtering, signal-aware process shutdown, timeout policy, approval-denied behavior, session discovery caching/timeouts, generated protocol validation, and the validation checklist.

Follow-up changes after `49134ef` make the protocol generator honor `CODEX_CLI_PATH` with Windows-safe command quoting/validation, remove the broad `NODE_` env prefix so `NODE_ENV` is forwarded but `NODE_OPTIONS` is explicitly blocked, and align App Server approval capabilities so denied approval request events do not advertise human approval workflow support.

## Changes

- Adds Codex App Server transport support in `@aif/runtime`
- Adds JSONL RPC client, process lifecycle handling, event mapping, session discovery, and structured error classification
- Adds generated Codex App Server protocol artifacts and a protocol generation/check script
- Wires `app-server` into runtime transport resolution, validation, capabilities, chat sessions, and runtime profile identity enrichment
- Improves command-line safety for App Server process spawning, especially on Windows
- Adds deferred turn handling, interrupt-aware turn flow, startup retry behavior, and timeout coverage
- Splits App Server env forwarding into exact keys and explicit safe prefixes; blocks similarly named secrets and `NODE_OPTIONS`
- Documents App Server approval requests as denied by design and covers command, file-change, and permissions approval denials
- Sets App Server `supportsApprovals` to `false` until a human approval bridge exists
- Adds session discovery cache/timeout coverage so discovery does not repeatedly spawn unnecessary runtime work
- Adds `npm run ai:protocol` to validation coverage
- Updates provider/configuration docs and environment examples
- Adds runtime/API/web tests around App Server transport behavior and UI usage display

## Pros of this transport

- Uses local Codex auth via `codex login`, so it can work without directly configuring an API key in AIF
- Streams structured JSONL notifications, giving better observability than plain CLI output
- Supports resumable Codex thread IDs and runtime-backed session discovery
- Behaves closer to the SDK path while still running through a local Codex process
- Keeps transport behavior explicit through runtime profiles and transport-aware capabilities
- Allows safer subprocess spawning by avoiding shell command snippets and validating unsafe Windows metacharacters
- Can expose local account/profile state and rate-limit data where the App Server protocol provides it

## Cons / tradeoffs

- Depends on the local Codex CLI and `codex app-server` being installed, authenticated, and compatible
- Adds protocol maintenance overhead because generated App Server schema/types must stay in sync with Codex CLI changes
- Subprocess lifecycle adds startup latency, timeout handling, retries, and cleanup complexity
- Usage reporting is still partial compared with API/SDK paths
- Local process execution means sandbox and approval settings must be configured carefully
- App Server approval requests are currently denied by design because no human approval bridge is implemented yet; the transport does not advertise `supportsApprovals` until that bridge exists
- App Server protocol behavior may change over time, so compatibility tests and generated artifact checks matter

## Testing / checks after `49134ef`

- `git ls-files | grep -E 'SFConflict|\(.*Conflict'`: empty output, no tracked conflict artifacts
- `npm run test --workspace=@aif/runtime -- codexAppServerProtocolGenerator.test.ts process.test.ts`
- `npm run test --workspace=@aif/runtime -- codexAdapterSdk.test.ts`
- `npm run test --workspace=@aif/runtime`
- `npm run test --workspace=@aif/api -- runtimeProfiles.test.ts`
- `npm run test --workspace=@aif/api -- chat.test.ts chatSessions.test.ts`
- `npm run test --workspace=@aif/web`
- `npm run ai:protocol`
- `npx turbo build --concurrency=1`
- `npx turbo test --concurrency=1`
- `npx turbo coverage --concurrency=1`
- `npm run ai:perf`
- `npm run ai:load`
- `npm run ai:checklist`
- `npm run ai:validate`


ichinya:
### 1. **BLOCKER: удалить все `SFConflict` файлы**

[blocker] Please remove all `(...SFConflict...)` files from the branch.

These look like conflict-resolution artifacts and should not be committed. They currently duplicate real source/test files across API, runtime, shared, and web packages. Please keep only the canonical files, re-resolve the merge against `main`, and re-run the full test/build matrix after cleanup.

### 2. **HIGH: env allowlist в `process.ts` слишком широкая**

[high] Please split env forwarding into exact keys and prefixes.

`ALLOWED_ENV_PREFIXES` currently includes exact env names like `HOME`, `USER`, `PATH`, `SHELL`, `TERM`, `TMPDIR`, `TZ`, `HTTP_PROXY`, but the check uses `key.startsWith(prefix)`. That can forward unrelated variables such as `USER_TOKEN`, `HOME_SECRET`, or `PATH_TO_SECRET` into the Codex subprocess.

Suggested fix:
- `ALLOWED_ENV_KEYS = new Set(["HOME", "USER", "LANG", "PATH", "SHELL", "TERM", "TMPDIR", "TZ", "HTTP_PROXY", ...])`
- `ALLOWED_ENV_PREFIXES = ["OPENAI_", "CODEX_", "AIF_", "HANDOFF_", "NODE_", "LC_", "XDG_", ...]`
- allow only `ALLOWED_ENV_KEYS.has(key) || ALLOWED_ENV_PREFIXES.some(prefix => key.startsWith(prefix))`
- add regression tests proving `USER_TOKEN` / `HOME_SECRET` are not forwarded.

### 3. **HIGH: process-exit checks should handle `signalCode`**


[high] Please treat signal-terminated subprocesses as already exited.

`terminateCodexAppServerProcess()` and `waitForExit()` currently check only `process.exitCode != null`. If the child was terminated by a signal, `exitCode` may remain `null` while `signalCode` is set. Please use a helper like:

const hasExited = (p: ChildProcess) => p.exitCode != null || p.signalCode != null;

and add a regression test for signal-terminated processes.

### 4. **HIGH: default run timeout `120_000` выглядит слишком агрессивно**


[high] Please align App Server run timeout with the existing runtime/stage timeout policy.

`DEFAULT_RUN_TIMEOUT_MS = 120_000` is risky for agent/stage execution. If `input.execution.runTimeoutMs` is missing, App Server runs can be killed after 2 minutes even when other transports/stage config allow longer execution.

Please either:
- inherit the existing runtime/stage default timeout,
- make this default configurable via profile/env and document it,
- or avoid adapter-specific run timeout unless the caller explicitly provides one.

Please add tests covering missing `runTimeoutMs` and long-running app-server runs.

### 5. **MEDIUM/HIGH: session discovery spawns a fresh app-server process per request**


[medium] Please consider caching/reusing the App Server session client, or throttle session discovery calls.

Session list/get/events currently spawn a new `codex app-server` process for each request. That may be expensive for UI-driven refreshes and can add latency or local account/rate-limit side effects.

At minimum, please add:
- request-level timeout coverage,
- tests for repeated session discovery,
- and either caching/throttling or a documented reason why one-process-per-call is acceptable.


### 6. **MEDIUM/HIGH: approval requests are always denied**

`eventMapper.ts` emits approval request events, but server-side approval requests return `decision: "decline"` or empty permissions. That is safe as a default, but it means any app-server run requiring command/file/permission approval will fail or stall behaviorally unless the profile is configured to avoid approvals. ([GitHub][6])


[medium/high] Approval handling needs to be explicit.

`handleServerRequest()` emits an approval event but always declines command/file approval requests and returns empty permissions for permission approval requests. If this transport is intended to support `approvalPolicy: "on-request"`, this needs a real approval bridge. If not, docs/profile validation should clearly state that App Server is currently headless and approval requests are denied by design.

Please add regression tests for:
- command approval request,
- file-change approval request,
- permission approval request,
- expected runtime result/error when approval is denied.


### 7. **MEDIUM: usage mapping drops some token breakdown fields**


[medium] Please either preserve richer token breakdown in runtime usage or document the limitation.

App Server usage normalization currently exposes only input/output/total/cost. If Codex App Server provides cached/reasoning/breakdown tokens in raw usage, consider exposing them in `raw`/extended usage fields or documenting that App Server usage is partial and not directly comparable with API/SDK transports.


### 8. **MEDIUM: generated protocol artifacts need CI drift check**

[medium] Please wire protocol generation/check into validation.

Since this PR adds generated Codex App Server protocol artifacts, please ensure there is a CI/checklist command that fails when generated files are stale, e.g. `codex:app-server:protocol:check`, and include it in the PR testing section.

ichinya:
### 1. **HIGH / PR hygiene: обновить PR description и validation section**

[high] Please update the PR description after `49134ef`.

The code now wires `ai:protocol` into `ai:validate`, but the PR testing section still lists the old validation set and does not mention the protocol drift check. Please update the PR body with the actual commands run after the hardening commit, including:

- `git ls-files | grep -E 'SFConflict|\(.*Conflict'`
- `npm run ai:protocol`
- `npm run ai:validate` or the exact sequential equivalent
- runtime/api/web targeted tests that were re-run after `49134ef`


### 2. **MEDIUM: protocol generator ignores `CODEX_CLI_PATH`**

[medium] Please make the protocol generation/check script honor `CODEX_CLI_PATH`.

Runtime process spawning resolves `options.codexCliPath` / `CODEX_CLI_PATH`, but `scripts/generate-codex-app-server-protocol.mjs` always shells out to `codex`. That can make `ai:protocol` check a different CLI than the runtime actually uses.

Suggested fix:
- read `process.env.CODEX_CLI_PATH` in `runCodex()`,
- default to `codex` only when it is not set,
- on Windows, quote/validate it with the same safety expectations as app-server process spawning,
- document that protocol artifacts are generated from the selected Codex CLI.


### 3. **MEDIUM / hardening: `NODE_` env prefix is still broad**


[medium] Please reconsider forwarding every `NODE_*` variable to the Codex subprocess.

The exact-key leak is fixed, but `ALLOWED_ENV_PREFIXES` still includes `NODE_`, which forwards `NODE_OPTIONS` and similar variables. If Codex is launched through a Node shim, this can alter child process behavior.

Please either:
- replace `NODE_` with exact keys such as `NODE_ENV`, or
- explicitly block `NODE_OPTIONS`, or
- document why forwarding all `NODE_*` variables is required and add a regression test for the intended behavior.

### 4. **LOW/MEDIUM: финальная проверка, что `SFConflict` реально отсутствует в git index**


[low/medium] Please include the `SFConflict` cleanup proof in the PR body.

The hardening commit appears to remove the conflict artifacts, but please add the output of:

git ls-files | grep -E 'SFConflict|\(.*Conflict'

Expected output: empty.

ichinya:
Request changes

Thanks for the follow-up hardening. I re-reviewed the latest head (`033b85b`) and the previous issues look addressed:

- conflict artifacts are reported as removed in the PR body;
- App Server env forwarding is now split into exact keys and explicit prefixes;
- `NODE_OPTIONS` is blocked while `NODE_ENV` is still forwarded;
- the protocol generator now honors `CODEX_CLI_PATH`;
- `ai:protocol` is included in validation;
- App Server approval-denied behavior is documented and tested.

I found one remaining capability-contract issue that should be fixed before merge.

## High: `supportsApprovals` should not be `true` while App Server denies approvals

`RuntimeCapabilities.supportsApprovals` is documented as support for human-in-the-loop approval workflows. However, the App Server docs now state that human approval bridging is not implemented and that command/file-change/permissions approval requests are denied by design.

`APP_SERVER_CAPABILITIES` currently sets:

```ts
supportsApprovals: true
```

That makes the transport advertise approval support to capability gates / UI / validation, while the implementation will actually decline approval requests.

Please either:

1. set Codex App Server `supportsApprovals` to `false` until a real approval bridge exists, and add/adjust a capability regression test; or
2. split the meaning into two separate capabilities, for example:

   * `supportsApprovalEvents`
   * `supportsHumanApprovals`

Given the current implementation and docs, I think the simplest correct fix is:

```ts
const APP_SERVER_CAPABILITIES: RuntimeCapabilities = {
  ...
  supportsApprovals: false,
  ...
};
```

lee-to:
## Code review

Follow-up review for the new commit `185958f` only. This pass checks the previous capability-contract blocker plus the small docs/test delta that landed after the last review.

### Response pass

Addressed: ✔ The previous blocker is resolved. Codex App Server now reports `supportsApprovals: false` in `packages/runtime/src/adapters/codex/index.ts:145`, the capability regression test now asserts that value in `packages/runtime/src/__tests__/codexAdapterSdk.test.ts:149`, and `docs/providers.md:269` documents why approval request events are observable while human approval workflows are not supported.

New: none found in the follow-up delta.

Still unresolved: none.

Previous review: https://github.com/lee-to/aif-handoff/pull/94#issuecomment-4337320572

### Must fix

None.

### Should fix

None.

### Nits

None.

### Context gates

- Architecture: pass — the follow-up only changes Codex runtime capabilities, a runtime unit test, and provider docs; no DB or cross-package boundary drift.
- Rules: pass — the runtime adapter sync rule is satisfied for this delta: capability, docs, and test were updated together.
- Roadmap: pass — no new roadmap contradiction in this follow-up.
- CHECKLIST compliance (touched packages: `packages/runtime`): pass — docs were synced, the touched adapter has a test update, and no new adapter contract field was introduced.
- Docs sync: pass — `docs/providers.md` now matches the App Server capability behavior.
- PR size: pass for follow-up delta — 4 additions / 4 deletions across 3 files. The full PR is large, but this response pass is scoped to the post-review delta.
- Risk gating (feature flag): n/a for this follow-up — the new commit narrows an advertised capability and does not introduce a new runtime path.

CI is green on GitHub. I could not run the targeted runtime test locally because `vitest` is not installed in this checkout's `node_modules`, but the PR `Tests` check is passing.

Verdict: LGTM now.

lee-to:
## Review findings

1. **High: Windows command injection path in App Server validation**

`packages/runtime/src/adapters/codex/index.ts` validates App Server by calling `validateCodexCliConnection`, which reaches `probeCodexCli` in `packages/runtime/src/adapters/codex/cli.ts`. On Windows that probe uses `execFileSync(..., { shell: true })` before App Server’s new metacharacter guard runs. A profile-controlled `codexCliPath` / `CODEX_CLI_PATH` containing `&`, `|`, etc. can therefore be interpreted by `cmd.exe` during validation, even though the App Server spawn path later rejects those characters. Please move the Windows safety check before the probe, or avoid `shell: true` for configured paths.

2. **Medium: App Server denies only part of the generated approval protocol**

`packages/runtime/src/adapters/codex/appServer/eventMapper.ts` handles only `item/commandExecution/requestApproval`, `item/fileChange/requestApproval`, and `item/permissions/requestApproval`. The generated protocol also includes `applyPatchApproval` and `execCommandApproval` in `packages/runtime/src/adapters/codex/appServer/generated/ServerRequest.ts`. If Codex emits those request names, AIF returns JSON-RPC error `-32000` instead of the intended denial, so unattended runs can fail in a transport/protocol error path rather than a clean permission-denied path. Please add explicit safe responses, likely `{ decision: "denied" }`, and tests for both methods.

Context gates: architecture OK; runtime/API/web boundaries look aligned. Checklist/rules gate has the approval-contract warning above. Roadmap alignment OK for runtime-adapter expansion.

Checks I ran:

- `npm run test --workspace=@aif/runtime -- codexAppServerProtocolGenerator.test.ts process.test.ts jsonlRpcClient.test.ts run.test.ts sessions.test.ts eventMapper.test.ts codexModelDiscoveryRpc.test.ts` — passed, 65 tests
- `npm run ai:protocol` — passed; Codex emitted sandbox PATH warnings, artifacts are in sync
- `npm run test --workspace=@aif/api -- runtimeProfiles.test.ts chat.test.ts chatSessions.test.ts tasks.test.ts` — passed, 199 tests
- `npm run test --workspace=@aif/web -- Header.test.tsx RuntimeProfileForm.test.tsx` — passed, 37 tests

PR #134, opened 2026-06-23
fix(queue): restore FIFO backlog ordering

### LINKED ISSUE #133: [bug] Auto-Queue default backlog ordering is still LIFO and cross-project
## Summary

`Auto-Queue` is supposed to pick the next backlog task by ascending `position`.
That part is logically correct.

The remaining bug is in the default `createTask()` position assignment:
new backlog tasks are still inserted at the **front** of the queue with `min(position) - 100`, so Auto-Queue behaves like **LIFO** for normal task creation.

There is also a second bug: the default position query is not scoped by `projectId`, so backlog tasks from one project can influence the positions assigned in another project.

Related context:
- `#108` addressed the ROADMAP import symptom by assigning explicit positions during import.
- The generic `createTask()` path still has the same root problem on current `main`.

## Current Behavior

In `packages/data/src/index.ts`, `createTask()` still does this for the default backlog position:

```ts
position: input.position ?? (() => {
  const row = db
    .select({ minPos: min(tasks.position) })
    .from(tasks)
    .where(eq(tasks.status, "backlog"))
    .get();
  return (row?.minPos != null ? Number(row.minPos) : 1000) - 100;
})(),
```

At the same time, `nextBacklogTaskByPosition(projectId)` picks the backlog task with the smallest `position`.

That means:
- create task A, then B, then C in the same project
- positions become smaller for newer tasks
- Auto-Queue picks `C` first, then `B`, then `A`

So the effective behavior is LIFO, not FIFO.

The cross-project leak is caused by the same query using:

```ts
.where(eq(tasks.status, "backlog"))
```

instead of scoping by `projectId`.

## Expected Behavior

- New tasks created through the normal `createTask()` path should be appended to the **end** of the backlog for their own project.
- Auto-Queue should continue to pick the smallest `position`, which would then produce FIFO ordering for newly created tasks.
- Backlog position calculation must be isolated per project.

## Proposed Fix

In `packages/data/src/index.ts`, update the default position calculation in `createTask()`:

```ts
position: input.position ?? (() => {
  const row = db
    .select({ maxPos: max(tasks.position) })
    .from(tasks)
    .where(
      and(
        eq(tasks.projectId, input.projectId),
        eq(tasks.status, "backlog"),
      ),
    )
    .get();

  return (row?.maxPos != null ? Number(row.maxPos) : 1000) + 100;
})(),
```

This keeps Auto-Queue logic unchanged and fixes the actual source of the problem.

## Stability Improvement

`nextBacklogTaskByPosition()` currently sorts only by `position`.
It would be safer to make the selection stable:

```ts
.orderBy(asc(tasks.position), asc(tasks.createdAt), asc(tasks.id))
```

## Existing Data

Already-created backlog tasks may still have reversed positions.
A one-time normalization migration or maintenance script would help.

Example SQLite normalization:

```sql
WITH ordered AS (
  SELECT
    id,
    ROW_NUMBER() OVER (
      PARTITION BY project_id
      ORDER BY created_at ASC, id ASC
    ) AS rn
  FROM tasks
  WHERE status = 'backlog'
)
UPDATE tasks
SET position = (
  SELECT rn * 100
  FROM ordered
  WHERE ordered.id = tasks.id
)
WHERE id IN (SELECT id FROM ordered);
```

## Suggested Tests

- Create tasks `A`, `B`, `C` in the same project and assert their positions increase.
- Assert `nextBacklogTaskByPosition(projectId)` returns `A`.
- Create backlog tasks in another project and assert they do not affect positions in the first project.
- Add coverage for the normalization/migration path if implemented.

## Reproduction

1. Enable Auto-Queue for a project.
2. Create backlog tasks `A`, then `B`, then `C` through the normal task creation flow.
3. Let the coordinator advance the next backlog task.
4. Observe that `C` is picked first because it has the smallest `position`.

The key point is that Auto-Queue itself is already doing the right thing by picking the smallest `position`; the bug is that new tasks are currently inserted at the wrong end of the queue.

ichinya:
## Summary

Fixes #133.

This PR restores FIFO backlog behavior for ordinary task creation without changing Auto-Queue semantics.

Changes:
- append default backlog positions per project instead of prepending them
- keep Auto-Queue logic unchanged: it still selects the backlog task with the smallest `position`
- make backlog selection deterministic by ordering on `position`, `createdAt`, and `id`
- preserve roadmap import as the explicit front-of-queue path via explicit positions
- add an opt-in backlog normalization utility for legacy data
- document the new backlog ordering and remediation workflow

## Testing

Ran:
- `npm run test --workspace=@aif/data -- src/__tests__/index.test.ts src/__tests__/normalizeBacklogPositions.test.ts`
- `npm run test --workspace=@aif/agent -- src/__tests__/autoQueue.test.ts`
- `npm run test --workspace=@aif/api -- src/__tests__/tasks.test.ts src/__tests__/roadmapGeneration.test.ts`
- `npm run coverage --workspace=@aif/data`
- `npm run lint`
- `npm run build`

## Notes

- Legacy backlog positions are not rewritten automatically because manual backlog reordering is already supported.
- The normalization utility is intentionally opt-in and supports dry-run by default.
- `npm run ai:validate` still fails at the repository-wide `prettier --check` step because of pre-existing formatting drift in unrelated files outside this PR.

### DIFF
--- a/.ai-factory/plans/feat-auto-queue-backlog-fifo.md
+++ b/.ai-factory/plans/feat-auto-queue-backlog-fifo.md
@@ -1,0 +1,133 @@
+# Implementation Plan: Auto-Queue Backlog FIFO Ordering
+
+Branch: feat/auto-queue-backlog-fifo
+Created: 2026-06-23 09:09
+
+## Settings
+
+- Testing: yes
+- Logging: verbose
+- Docs: yes
+
+## Roadmap Linkage
+
+Milestone: "none"
+Rationale: Current roadmap milestones are already completed; this plan is a post-delivery correctness fix for backlog ordering and legacy backlog data.
+
+## Overview
+
+Issue `#133` is the remaining root-cause fix after the ROADMAP-specific patch from `#108`.
+
+Current `upstream/main` behavior is internally inconsistent:
+
+- `nextBacklogTaskByPosition()` correctly advances the backlog task with the smallest `position`
+- ordinary `createTask()` inserts new backlog tasks with `min(position) - 100`
+- the default create flow therefore behaves like LIFO instead of FIFO
+- the default position query is also global across all backlog tasks, so one project's backlog can influence another project's newly created task positions
+
+There is one important constraint: backlog order is now user-editable via manual reorder APIs/UI. Because of that, a blanket migration that rewrites every existing backlog task by `createdAt` risks destroying intentional operator ordering. Legacy cleanup should therefore be an explicit, opt-in remediation path rather than an automatic migration.
+
+### Key Decisions
+
+- Keep Auto-Queue semantics unchanged: the coordinator should continue to pick the smallest backlog `position`.
+- Fix the source of the bug in `createTask()` by appending ordinary backlog tasks to the tail of their own project's backlog.
+- Preserve explicit `position` overrides for special callers such as roadmap import.
+- Make `nextBacklogTaskByPosition()` deterministic for tied positions by ordering on `position`, `createdAt`, then `id`.
+- Do not add a schema/data migration just to rewrite existing backlog positions; the schema is unchanged and legacy ordering remediation should stay operator-invoked.
+- Treat legacy backlog normalization as an operator action, not an automatic schema/data migration.
+- Preserve the current ROADMAP import special-case ordering unless product requirements explicitly change it in a separate issue.
+- Treat the UI/API ordering change for ordinary newly created backlog tasks as intentional: after this fix, normal task creation should place new items at the backlog tail instead of the head.
+
+### Expected Change Surface
+
+- `packages/data/src/index.ts`
+- `packages/data/src/__tests__/index.test.ts`
+- `packages/data/src/normalizeBacklogPositions.ts` or `packages/data/src/tools/normalizeBacklogPositions.ts` (new helper/utility logic)
+- `packages/data/src/__tests__/normalizeBacklogPositions.test.ts` (new)
+- `packages/agent/src/__tests__/autoQueue.test.ts`
+- `packages/api/src/__tests__/tasks.test.ts`
+- `packages/api/src/__tests__/roadmapGeneration.test.ts`
+- `packages/api/src/services/roadmapGeneration.ts` (only if comments or helper usage need alignment)
+- `packages/data/package.json` and/or root `package.json` (launcher wiring only if needed)
+- `scripts/normalize-backlog-positions.mjs` (optional thin launcher only if implementation chooses a root entrypoint)
+- `docs/architecture.md`
+- `docs/api.md`
+- `docs/configuration.md` or another operator-facing doc only if the normalization utility introduces operator-facing flags/env beyond the invocation itself
+
+### Risks / Open Questions
+
+- Rewriting existing backlog positions automatically would clobber manual reorder state; the implementation should avoid hidden rewrites.
+- ROADMAP import currently uses explicit front-of-queue positioning to preserve phase/sequence semantics. This plan assumes that special behavior stays in place.
+- If maintainers want "every creation path appends to backlog tail" with no special cases, that is a separate product decision and should not be silently bundled into this fix.
+- The implementation should choose a runtime path for the remediation utility that does not rely on plain Node importing workspace TypeScript source directly; if a launcher is added, it must either execute built `dist` output or opt into a supported TS runtime explicitly.
+
+## Commit Plan
+
+- **Commit 1** (after tasks 1-2): `fix(data): append new backlog tasks per project`
+- **Commit 2** (after tasks 3-5): `test(queue): cover backlog ordering regressions`
+- **Commit 3** (after tasks 6-8): `docs: add backlog ordering remediation guidance`
+
+## Tasks
+
+### Phase 1: Correct Backlog Ordering Semantics
+
+- [x] Task 1: Fix ordinary backlog task creation to append within the owning project
+  - Files: `packages/data/src/index.ts`
+  - Deliverable: change the default `createTask()` position path from global `min(position) - 100` to project-scoped tail insertion (preferably `max(position) + 100` scoped by `projectId`), while preserving explicit `position` overrides for callers that intentionally manage queue placement.
+  - Include/consider: whether to add a symmetric helper such as `getMaxBacklogPosition(projectId)` alongside `getMinBacklogPosition(projectId)` to keep the intent explicit and testable.
+  - Logging requirements: do not add noisy per-task creation logs on the hot path; if helper-level diagnostics are needed, keep them `DEBUG` only and consistent with existing `@aif/data` logging patterns.
+
+- [x] Task 2: Make backlog selection deterministic without changing coordinator semantics
+  - Files: `packages/data/src/index.ts`
+  - Deliverable: update `nextBacklogTaskByPosition(projectId)` to order by `asc(position)`, `asc(createdAt)`, `asc(id)` so tied positions are stable and reproducible. Review adjacent position-sorted queries for semantic consistency, but do not broaden the change beyond backlog selection unless a concrete bug is found.
+  - Logging requirements: preserve existing coordinator/data logging behavior; any touched log lines must remain concise and at current levels unless a new failure mode requires explicit `WARN`/`ERROR`.
+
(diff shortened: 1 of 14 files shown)


lee-to:
## Code review

This PR fixes the backlog FIFO issue by appending ordinary `createTask()` rows per project and keeping roadmap import as the explicit front-of-queue path. The core queue direction looks correct and CI is green, but I found one blocking correctness issue in the new normalization utility and one scope issue in the lockfile refresh.

### Must fix

1. **Normalization apply can rewrite tasks that are no longer backlog** — `packages/data/src/normalizeBacklogPositions.ts:156` — `normalizeBacklogPositions()` builds a plan from current backlog rows, then later applies updates with `where(eq(tasks.id, task.id))` only. If the coordinator or a user advances a task after the plan read but before/during apply, the utility can still overwrite that non-backlog task's `position`, which violates the documented contract in `docs/configuration.md` that the rewrite only touches `status = backlog` rows. Keep the read and write in one transaction and/or guard each update with `and(eq(tasks.id, task.id), eq(tasks.projectId, task.projectId), eq(tasks.status, "backlog"))`; return the actual updated row count from `.run().changes` and add a regression test for a task that leaves backlog between planning and apply.

### Should fix

1. **Separate or narrow the unrelated lockfile refresh** — `package-lock.json:6927`, `package-lock.json:11636`, `package-lock.json:11765`, `package-lock.json:11983`, `package-lock.json:12292` — the PR's functional change only adds the `@aif/data` normalization script/dependency, but the lockfile also moves unrelated packages such as `hono`, `turbo`, `undici`, `vite`, and `ws`. That mixes dependency/audit risk into a queue-ordering bug fix and makes rollback/review harder. Please either split the audit/dependency refresh into its own PR or regenerate the lockfile so this PR only carries the `tsx`/data-script delta required by the new utility.

### Nits

- None.

### Context gates

- Architecture: pass — `api` and `agent` continue to use `@aif/data`; the direct DB work stays inside `@aif/data`.
- Rules: error — the apply path for the destructive normalization utility does not preserve its backlog-only invariant under concurrent state changes.
- Roadmap: pass — the PR references issue `#133`, and the included plan states this is a post-roadmap correctness fix.
- CHECKLIST compliance (touched packages: `data`, `api`, `agent`): warn — package tests/docs are present, but the `@aif/data` checklist's edge-case bar needs the stale-plan/non-backlog apply case covered.
- Docs sync: pass — API, architecture, and configuration docs were updated for the queue semantics and operator utility.
- PR size (1,554 changed lines / 14 files / 4 concerns): warn — after excluding the lockfile and `.ai-factory` plan, the reviewable change is still about 852 lines, mostly tests/docs. The blocking decomposition concern is the unrelated dependency refresh.
- Risk gating (feature flag): pass — no environment flag is needed for the targeted `#133` bug fix; the new destructive remediation behavior is operator-invoked and defaults to dry-run.

Verdict: `REQUEST_CHANGES` until the normalization apply path is guarded and the lockfile scope is handled.

ichinya:
Addressed in `c9ae77f`.

- Guarded normalization apply with `id`, `projectId`, and `status = backlog`, so a stale plan cannot rewrite a task that already left backlog.
- `updatedTaskCount` now reports actual DB changes from `.run().changes`.
- Added a regression test for a task leaving backlog between planning and apply.
- Narrowed `package-lock.json` so the net diff only carries the `@aif/data` `tsx` entry needed by the new utility.

One note: after removing the audit refresh from this PR, the `npm audit` check is expected to fail again on existing `hono` / `undici` / `ws` advisories. That needs the dependency refresh as a separate PR or merge before this PR can be fully green.

ichinya:
Follow-up on the failed `npm audit` job: https://github.com/lee-to/aif-handoff/actions/runs/28024212827/job/82947808198?pr=134

This is the expected result after narrowing the lockfile scope per review feedback. The job fails on existing runtime dependency advisories:

- `hono <=4.12.24` (`hono@4.12.18` in the current lockfile)
- `undici 7.0.0 - 7.27.2` (`undici@7.25.0` in the current lockfile)
- `ws 8.0.0 - 8.20.1` (`ws@8.20.0` in the current lockfile)

The earlier broader lockfile refresh made this audit pass, but it was removed from this PR to keep the queue-ordering change scoped. To make CI fully green, the dependency/audit refresh needs to be restored here or handled in a separate PR and merged before this one.

lee-to:
## Code review follow-up

Response pass for prior review: https://github.com/lee-to/aif-handoff/pull/134#issuecomment-4777462353

### Response pass

- Addressed: `packages/data/src/normalizeBacklogPositions.ts:152` now applies normalization inside a transaction, guards each update by `id`, `projectId`, and `status = backlog`, and returns the actual `.run().changes` count.
- Addressed: `packages/data/src/__tests__/normalizeBacklogPositions.test.ts:290` adds the stale-plan regression where a task leaves backlog after the plan is read.
- Addressed: `package-lock.json` is now narrowed to the single `@aif/data` `tsx` devDependency entry needed by the new utility; the unrelated `hono` / `turbo` / `undici` / `vite` / `ws` refresh is gone.
- New: the `npm audit` check is red, but the log shows the failures are the existing runtime dependency advisories for `hono`, `undici`, `ws`, and `qs` after the audit refresh was removed from this scoped PR. The net PR diff does not introduce those production dependency versions.
- Still unresolved: no code-review blockers from the prior pass.

### Must fix

None from this follow-up.

### Should fix

None from this follow-up. The audit failure still needs to be handled before merge if the branch requires green checks, but it should be a separate dependency/audit PR or a base-branch fix unless maintainers decide to re-scope this PR.

### Nits

- None.

### Context gates

- Architecture: pass — data access remains inside `@aif/data`; `api` and `agent` only consume the data package.
- Rules: pass — the destructive normalization path now preserves the backlog-only write invariant and has edge-case coverage.
- Roadmap: pass — this remains tied to issue `#133` and the included implementation plan.
- CHECKLIST compliance (touched packages: `data`, `api`, `agent`): pass — data edge coverage was added, API/agent changes are test-only regression coverage, and focused tests are reported in the PR body.
- Docs sync: pass — API, architecture, and configuration docs are still aligned with the queue semantics and operator utility.
- PR size (1,050 changed lines / 14 files / 3 concerns): warn — still above the soft threshold, but the remaining size is mostly tests/docs and the prior unrelated lockfile concern is resolved.
- Risk gating (feature flag): pass — no feature flag is needed for this targeted bug fix; the risky remediation utility remains operator-invoked and dry-run by default.
- CI: warn — build, lint, MCP unit, and tests are green; `npm audit` is red on existing advisories after the unrelated audit refresh was removed.

Verdict: `COMMENT` — code-review blockers are addressed; merge readiness depends on how maintainers want to handle the pre-existing audit failure.

ichinya:
Follow-up after #135 was merged.

I merged the updated `main` into this branch in `fb9597f`, so PR #134 now includes the separate audit/dependency refresh from #135 without carrying that change directly in the FIFO fix commits.

Local validation after the merge:

- `npm audit --omit=dev --audit-level=high` passes
- `npm run lint --workspace=@aif/data` passes

The focused DB test could not run locally because this Windows/Node 24 environment currently lacks the `better-sqlite3` native binding after `npm ci --ignore-scripts`; CI runs on Ubuntu/Node 22 and has been restarted for this branch.

lee-to:
## Codex PR Review

Verdict: APPROVE

Response pass for prior review: https://github.com/lee-to/aif-handoff/pull/134#issuecomment-4780549619

The remaining follow-up is clean. Since the last review, the branch merged current `main` in `fb9597f`; that follow-up delta only updates `package-lock.json` with the audit/dependency refresh already landed upstream. I do not see new source changes in that merge commit.

Prior findings:
- The stale normalization update blocker remains addressed: updates are applied with `id + projectId + status = backlog`, and the result counts actual `run().changes`.
- The regression coverage for a task leaving backlog between planning and apply is present.
- The broad audit lockfile refresh is now reconciled through `main` / #135, and CI audit is green.

Context gates:
- Architecture: PASS
- Project rules / DB boundary: PASS
- CHECKLIST coverage for touched packages: PASS
- Docs sync: PASS
- CI: PASS (`Build`, `ESLint`, `MCP Unit`, `Tests`, `mcpunit`, `npm audit`)
- PR size: WARN only; the diff is still large, but mostly focused implementation, tests, docs, and the upstream audit lockfile refresh.

No blocking findings remain from my side.

PR #136, opened 2026-06-25
fix(web): make randomUUID work outside secure contexts

Aleksey512:
## Fix: `crypto.randomUUID is not a function` when served over plain HTTP on a non-localhost host

  ### Problem
  Deploying the web UI to an internal server and opening it via a LAN address
  over plain HTTP — e.g. `http://10.10.100.29:5180` (not `localhost`, not HTTPS) —
  crashes when sending a chat message:

  Uncaught (in promise) TypeError: crypto.randomUUID is not a function
      at sendMessage (useChat.ts)

  ### Root cause
  `crypto.randomUUID()` is only exposed in a **secure context** — HTTPS or
  `http://localhost`. On a plain-HTTP, non-localhost origin the browser leaves
  `crypto.randomUUID` `undefined`, so the call in
  `packages/web/src/hooks/useChat.ts` throws. (Works on localhost / behind HTTPS,
  which hides it in typical dev.)
  See: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#secure_context

  ### Fix
  Add `packages/web/src/lib/uuid.ts` — a small dependency-free `randomUUID()` that
  uses the native `crypto.randomUUID()` when present and otherwise derives an
  RFC 4122 v4 UUID from `crypto.getRandomValues()`, which **is** available in
  insecure contexts (so randomness stays crypto-grade). Swap the single browser
  call site to use it.

  ### Tests
  `uuid.test.ts`: native path, insecure-context fallback (the case that used to
  throw), no-crypto last resort, and uniqueness. `vitest`, `eslint`, `tsc` all pass.

  ### Note
  Serving over HTTPS or via `localhost` also resolves it, but this makes the app
  work on a bare internal HTTP host out of the box.

  PR/коммит/ветку оставляю тебе (ты сказал, что сам сделаешь). Хочешь — закоммичу в новую ветку и оформлю PR через gh/glab, если репозиторий это поддерживает.

  ### Tests
  `uuid.test.ts`: native path, insecure-context fallback (the case that used to
  throw), no-crypto last resort, and uniqueness. `vitest`, `eslint`, `tsc` all pass.

  ### Note
  Serving over HTTPS or via `localhost` also resolves it, but this makes the app
  work on a bare internal HTTP host out of the box.

### DIFF
--- a/packages/web/src/hooks/useChat.ts
+++ b/packages/web/src/hooks/useChat.ts
@@ -9,6 +9,7 @@ import type {
   RuntimeLimitSnapshot,
 } from "@aif/shared/browser";
 import { api, ApiError } from "@/lib/api";
+import { randomUUID } from "@/lib/uuid";
 import { getWsClientId } from "./useWebSocket";
 
 interface SessionStreamState {
@@ -297,7 +298,7 @@ export function useChat(
         currentSessionIdRef.current = null;
       }
 
-      const newConversationId = crypto.randomUUID();
+      const newConversationId = randomUUID();
       const effectiveSessionId = forceNewSession
         ? null
         : (sessionId ?? currentSessionIdRef.current);
--- a/packages/web/src/lib/uuid.test.ts
+++ b/packages/web/src/lib/uuid.test.ts
@@ -1,0 +1,43 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { randomUUID } from "./uuid";
+
+const V4 = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+describe("randomUUID", () => {
+  it("returns a valid v4 UUID via native randomUUID (secure context)", () => {
+    expect(randomUUID()).toMatch(V4);
+  });
+
+  it("falls back to getRandomValues when randomUUID is missing (insecure http context)", () => {
+    // Simulate plain-HTTP, non-localhost origin: randomUUID is undefined,
+    // getRandomValues is still present. This is the case that used to throw
+    // `TypeError: crypto.randomUUID is not a function`.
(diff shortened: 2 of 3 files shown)


lee-to:
## Code review

This PR fixes the insecure-context chat crash by replacing the browser-only `crypto.randomUUID()` call with a small web helper that keeps the native path when available and falls back to `crypto.getRandomValues()` for plain-HTTP LAN hosts. The code change is narrow, covered by focused unit tests, and does not touch API contracts, schema, runtime adapters, Docker config, CSS, or UI component structure. Verdict: **COMMENT** only because the `Tests` check is still in progress; code review has no blocking findings.

### Must fix

None.

### Should fix

None from the code diff. Before merge, wait for the pending `Tests` check to complete successfully.

### Nits

None.

### Context gates

- Architecture: pass — the change stays inside `packages/web` and does not violate package dependency boundaries.
- Rules: pass — no expensive CSS, no new UI components, no DB/runtime adapter boundary changes, and no string-based error classification.
- Roadmap: warn — this is a `fix(web)` PR and it does not link a roadmap milestone. Non-blocking; the affected area maps to the completed AI Chat Sessions/chat workflow work.
- CHECKLIST compliance (touched packages: `packages/web`): warn — applicable UI/CSS/API/WS checklist items are either satisfied or not applicable; `npm run lint`/Build/ESLint are green in CI, while `Tests` is still in progress at review time.
- Docs sync: pass — no REST route, schema, env var, runtime adapter, Docker, or UI theme documentation obligation is triggered by this PR.
- PR size (88 changed lines / 3 files / 1 concern): pass — well below the soft size threshold and scoped to one browser compatibility fix.
- Risk gating (feature flag): n/a — low-risk compatibility fallback; secure-context behavior continues to use native `crypto.randomUUID()`, and the fallback only activates when the current code path would otherwise throw.

### Positive notes

The tests cover the native path, the insecure-context fallback that caused the bug, the no-crypto last resort, and basic uniqueness. That is the right coverage shape for this helper.

Final verdict: no code changes requested; merge should wait for the pending `Tests` job to finish green.

PR #139, opened 2026-06-28
feat(projects): add compact project task overview endpoint (additive)

ichinya:
## Summary

Per review, rescoped to **purely additive**: `GET /projects/overview` endpoint + shared types + data layer + docs + tests. **No UI migration, no flag.** Head: `916e0a9` (10 files, +378/-1).

The dashboard UI migration (App.tsx, ProjectsOverview → `/overview`) moved to a separate follow-up PR, per the review's option 1 ("keep this PR purely additive: endpoint, data layer, shared types, docs, tests").

## What's here

- **Shared types:** `ProjectTaskOverview`, `ProjectTaskPreview` (exported via `@aif/shared` + `@aif/shared/browser`).
- **Data layer:** `listProjectTaskOverviews(previewLimit)` — per-project status counts, metric totals (`total*`), retry/flag counts, small title previews bounded in SQL via `row_number()`.
- **API:** `GET /projects/overview` route + repository re-export.
- **Docs:** `### Project Task Overview` documents the actual wire shape (`totalToken*`, `blockedTasks`, `completedTasks`/…) — fixes the docs/contract mismatch raised in the #138 review (must-fix #1).

## Why no flag

The endpoint is **purely additive** — it adds a new route without changing or replacing any existing contract. Nothing user-facing switches behavior in this PR. The dashboard keeps its current behavior (legacy bare `GET /tasks`) until the separate UI-migration PR lands.

## Chain context

| PR | Scope | State |
|---|---|---|
| **#138** | `TaskListItem` + lite scoped `GET /tasks` + Board migration | ✅ approved |
| **#139 (this)** | `GET /projects/overview` endpoint (additive) | pure backend, no UI |
| **follow-up** | dashboard UI migration → `/overview` | separate PR |
| **#142** | remove dead `listTasks()` no-arg client overload (non-breaking) | stacked |
| **#140** | perf runner (independent) | ✅ approved |

## Verification

- `tsc --noEmit` clean across shared/data/api/web.
- Tests: data **214**, api **405**, web **666** — all green.
- `prettier` + `build` green.

## Fork-to-fork note

GitHub doesn't allow fork-to-fork PRs, so this targets `main` and its diff temporarily includes #138's commits **until #138 merges**, then auto-rebases to this single commit (`916e0a9`). Review #138 first.

PR #143, opened 2026-06-29
feat(web): migrate dashboard to GET /projects/overview

ichinya:
## Summary

Follow-up to the merged additive `/projects/overview` endpoint (#139). Moves the dashboard consumers off the legacy bare `GET /tasks` (full `Task[]`) onto the aggregate `/overview` endpoint.

This is the **UI-migration step** that was deliberately split out of #139 per the review ("keep that PR purely additive"). #139 landed the endpoint; this PR lands the consumer switch.

## What's here

**Dashboard migration** (consume `ProjectTaskOverview[]` from `/overview`):
- **`App.tsx`**: header metrics switch from `api.listTasks()` (bare) + `calculateTaskMetrics` → `useProjectTaskOverviews()` + `calculateOverviewMetrics`.
- **`ProjectsOverview.tsx`**: card view reads per-project `statusCounts`, `statusPreviews`, and metric totals from the aggregate endpoint instead of fanning out full task lists.
- **`useProjects.ts`**: add `useProjectTaskOverviews` hook + `invalidateProjectTaskOverviews` helper; invalidate the overview query on project mutations.
- **`useWebSocket.ts`**: invalidate the overview query on task/roadmap events.
- **`useTasks.ts`**: list path is scoped-only (`TaskListItem[]`); the no-arg bare caller path is removed (the dashboard no longer uses it).
- **`lib/api.ts`**: `listProjectTaskOverviews` method; `listTasks(projectId)` now requires `projectId` (no-arg overload removed).
- **`lib/taskMetrics.ts`**: `calculateOverviewMetrics` + `calculateProjectOverviewMetrics` reduce `ProjectTaskOverview[]` into the dashboard summary.

**Tests:** AppSmoke asserts overview hook wiring; `ProjectsOverview` regression test for the empty-project loading state; `api.test` covers the `/overview` endpoint.

## Merge-safety

Builds on the merged #139 endpoint. The server bare `GET /tasks` route stays available (legacy) — the web client just no longer calls it. After this lands, the bare route's only remaining caller is gone, and a future cleanup PR can remove the dead no-arg client overload / deprecate the server route.

## Verification

- `tsc --noEmit` clean (web).
- web tests **674** green (incl. the empty-project loading regression + overview endpoint test).
- `prettier` / `lint` / `build` green.
- 10 files, +347 / -118.

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -1,22 +1,21 @@
 import { useState, useCallback, useEffect, useMemo } from "react";
-import { QueryClient, QueryClientProvider, useQuery } from "@tanstack/react-query";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { Header } from "./components/layout/Header";
 import { Board } from "./components/kanban/Board";
 import { TaskDetail } from "./components/task/TaskDetail";
 import { CommandPalette } from "./components/layout/CommandPalette";
 import { useWebSocket } from "./hooks/useWebSocket";
 import { useCommitToasts } from "./hooks/useCommitToasts";
-import { useProjects } from "./hooks/useProjects";
+import { useProjectTaskOverviews, useProjects } from "./hooks/useProjects";
 import { useTasks } from "./hooks/useTasks";
 import { useTheme } from "./hooks/useTheme";
 import { useKeyboardShortcut } from "./hooks/useKeyboardShortcut";
 import { ChatBubble } from "./components/chat/ChatBubble";
 import { ChatPanel } from "./components/chat/ChatPanel";
-import { calculateTaskMetrics } from "./lib/taskMetrics";
+import { calculateOverviewMetrics, calculateTaskMetrics } from "./lib/taskMetrics";
 import { readStorage, writeStorage, removeStorage } from "./lib/storage";
 import { STORAGE_KEYS } from "./lib/storageKeys";
-import { api } from "./lib/api";
-import type { Project, Task } from "@aif/shared/browser";
+import type { Project } from "@aif/shared/browser";
 import { ProjectRuntimeSettings } from "./components/project/ProjectRuntimeSettings";
 import { ProjectsOverview } from "./components/project/ProjectsOverview";
 import { ToastProvider } from "./components/ui/toast";
@@ -30,13 +29,31 @@ const queryClient = new QueryClient({
   },
 });
 
+const PROJECT_ROUTE_PATTERN = /^\/project\/([^/]+)(?:\/task\/([^/]+))?/;
+
+function readInitialSelection(): { projectId: string | null; taskId: string | null } {
+  const match = window.location.pathname.match(PROJECT_ROUTE_PATTERN);
+  if (match) {
+    return { projectId: match[1] ?? null, taskId: match[2] ?? null };
+  }
+
+  return {
+    projectId: readStorage(STORAGE_KEYS.SELECTED_PROJECT),
+    taskId: null,
+  };
+}
+
 function AppContent() {
   useWebSocket();
   useCommitToasts();
   const { theme, toggleTheme } = useTheme();
   const { data: projects } = useProjects();
-  const [selectedProjectId, setSelectedProjectId] = useState<string | null>(null);
-  const [selectedTaskId, setSelectedTaskId] = useState<string | null>(null);
+  const [selectedProjectId, setSelectedProjectId] = useState<string | null>(
+    () => readInitialSelection().projectId,
+  );
+  const [selectedTaskId, setSelectedTaskId] = useState<string | null>(
+    () => readInitialSelection().taskId,
+  );
   const [commandOpen, setCommandOpen] = useState(false);
   const [chatOpen, setChatOpen] = useState(false);
   const [runtimeSettingsOpen, setRuntimeSettingsOpen] = useState(false);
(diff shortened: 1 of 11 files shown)


lee-to:
## Code review

Verdict: **REQUEST_CHANGES**.

This PR is a cohesive UI migration from dashboard-wide bare task fetches to the aggregate `/projects/overview` endpoint. The direction is good and the added regression coverage is useful, but two details need fixing before merge: the retained bare `api.listTasks()` path is now broken, and one documented WebSocket event path can leave the new overview metrics stale.

### Must fix

1. **Retained bare task helper now calls `/tasks/tasks`** — `packages/web/src/lib/api.ts:260` / `packages/web/src/lib/api.ts:265` — the no-arg overload is still exported as `api.listTasks()`, but its branch now does `request<Task[]>(`${API_BASE}/tasks`)` while `API_BASE` is already `"/tasks"`. Any remaining no-arg caller will hit `/tasks/tasks` instead of the legacy `GET /tasks` route. Either delete the no-arg overload and branch completely if this PR is meant to remove that client surface, or restore the legacy branch to `request<Task[]>(API_BASE)` and add a test that covers the retained path.

2. **Runtime usage WebSocket updates do not invalidate the new overview query** — `packages/web/src/hooks/useWebSocket.ts:224` / `packages/web/src/hooks/useWebSocket.ts:228` — `project:runtime_limit_updated` returns immediately after invalidating `["tasks"]`, so it skips `invalidateProjectTaskOverviews(queryClient)`. That event is documented as firing when persisted runtime-limit state or last usage changes (`docs/api.md:1298`), and the overview metrics now aggregate token/cost fields from `ProjectTaskOverview` (`packages/web/src/lib/taskMetrics.ts:130`). Add `invalidateProjectTaskOverviews(queryClient)` in this branch, ideally with a WebSocket hook regression test, so the dashboard header and project overview cards refresh after usage updates.

### Should fix

1. **PR size is just over the soft review limit** — 451 changed lines / 10 files / one concern. This is cohesive and test-heavy, so I am not asking for a split here, but keep the follow-up cleanup PR limited to removing the legacy task-list path rather than combining it with unrelated UI work.

### Nits

None.

### Context gates

- Architecture: pass — the web changes consume data through `lib/api.ts` and hooks; no package boundary or DB access violations found.
- Rules: error — the web checklist requires complete WebSocket invalidation for consumed events; the `project:runtime_limit_updated` branch misses the new overview query.
- Roadmap: warn — this is a `feat(web)` PR but does not link to a roadmap milestone. It does reference follow-up PR #139, which is enough context for review, but milestone traceability is still absent.
- CHECKLIST compliance (touched packages: `packages/web`): error — API calls are centralized and no new UI primitive was added, but the WebSocket invalidation checklist item is not fully satisfied.
- Docs sync: pass — this consumes an already documented endpoint/event; no new API contract is introduced in this PR.
- PR size (451 lines / 10 files / one concern): warn — over the soft line threshold, but cohesive.
- Risk gating (feature flag): n/a — this is a web consumer migration to an existing endpoint, with the old server route still available.

Positive note: the empty-project overview regression test is the right kind of coverage for this migration, and the aggregate metric reducer keeps the UI-side calculation small.

ichinya:
Both must-fix items addressed. Head: `9f75d18` (3 files, +33 / -10).

## Must-fix #1 — `/tasks/tasks` bare URL → overload removed entirely

You flagged that the no-arg overload hit `/tasks/tasks` (Bug #1, re-introduced when I migrated files off an older branch state). Since this PR migrates the dashboard off the bare path, the no-arg overload has **no remaining caller** — so rather than restore the legacy branch, I removed the overload completely:

- `listTasks(projectId: string)` is the only signature now; `projectId` is required at the type boundary.
- The bare branch + its `/tasks/tasks` URL are gone.
- `tsc --noEmit` confirms no remaining no-arg callers (the only list-path caller is `useTasks.ts`, always scoped).

The server bare route stays available for backward compatibility; the web client just no longer calls it.

## Must-fix #2 — `project:runtime_limit_updated` overview invalidation

Added `invalidateProjectTaskOverviews(queryClient)` to that branch (`useWebSocket.ts`), so the dashboard header + project cards refresh token/cost after usage updates:

```ts
if (data.type === "project:runtime_limit_updated" && hasRuntimeLimitPayload(data.payload)) {
  invalidateRuntimeLimitQueries(queryClient, data.payload);
  if (typeof data.payload.taskId === "string" && data.payload.taskId.length > 0) {
    pendingTaskIds.current.add(data.payload.taskId);
    queryClient.invalidateQueries({ queryKey: ["tasks"] });
  }
  // Overview aggregates token/cost fields, so refresh after usage updates.
  invalidateProjectTaskOverviews(queryClient);
  return;
}
```

## Regression test

`useWebSocketOverviewInvalidation.test.ts` pins that `invalidateProjectTaskOverviews` invalidates `["projectTaskOverviews"]`. A future change that drops or renames the overview query key surfaces immediately. (The handler isn't exported in isolation, so the test targets the public helper it calls — same surface the WS branch relies on.)

## Verification

- `tsc --noEmit` clean (web).
- web tests **675** green (+1 new regression test).
- `prettier` / `lint` / `build` green.

The should-fix (size) is left as-is per your note ("not asking for a split here").

lee-to:
## Code review

Follow-up review for the prior comment: https://github.com/lee-to/aif-handoff/pull/143#issuecomment-4830871253

Verdict: **APPROVE**.

This follow-up commit addresses both previously blocking issues without widening the PR scope. The bare no-arg task-list client path is removed, the runtime-limit WebSocket branch now refreshes the new overview query, and the current CI checks are green.

### Response pass

- Addressed: Must-fix #1, the retained bare `api.listTasks()` path. `packages/web/src/lib/api.ts:262` now exposes only `listTasks(projectId: string)`, and the `/tasks/tasks` branch is gone.
- Addressed: Must-fix #2, stale overview metrics after runtime usage updates. `packages/web/src/hooks/useWebSocket.ts:231` now calls `invalidateProjectTaskOverviews(queryClient)` in the `project:runtime_limit_updated` branch.
- New: none found in the follow-up delta.
- Still unresolved: none.

### Must fix

None.

### Should fix

1. **PR size remains over the soft review limit** - 492 changed lines / 11 files / one concern. This is still cohesive and test-heavy, so I am not blocking on decomposition. Keep the next cleanup PR narrow.

### Nits

None.

### Context gates

- Architecture: pass - the web package still consumes data through hooks and `lib/api.ts`; no package-boundary or DB-access violations found.
- Rules: pass - the previous WebSocket invalidation gap is fixed, and no new UI primitive or expensive CSS was introduced in the follow-up.
- Roadmap: warn - this `feat(web)` PR still does not link a roadmap milestone, though the PR body links the relevant follow-up context from #139.
- CHECKLIST compliance (touched packages: `packages/web`): pass - API calls remain centralized, the consumed WebSocket event now invalidates the affected overview query, and the follow-up adds regression coverage around the invalidation helper.
- Docs sync: pass - this consumes the already documented `/projects/overview` endpoint and `project:runtime_limit_updated` event; no new API contract is introduced.
- PR size (492 lines / 11 files / one concern): warn - over the soft line threshold, but still a single dashboard migration.
- Risk gating (feature flag): n/a - this is a web consumer migration to an existing endpoint, with the legacy server route still available.

LGTM now.

PR #146, opened 2026-07-06
fix(mcp): support concurrent clients on HTTP transport

dostoevskiy-spb:
## Summary

Fixes the Handoff MCP **HTTP transport** so multiple clients can connect concurrently. Previously only one client could ever `initialize`; any second client (a second Claude Code window, or a remote client) was rejected with JSON-RPC `-32600 "Server already initialized"`.

## The bug

**Symptom:** With the `handoff` MCP registered over HTTP (`http://localhost:<MCP_PORT>/mcp`), the first Claude Code window connects fine, but every other window fails to initialize the MCP — only one active session works at a time.

**Root cause:** `startHttp()` in `packages/mcp/src/index.ts` created **one** `McpServer` and **one** `StreamableHTTPServerTransport({ sessionIdGenerator: () => randomUUID() })` for the whole process, and routed every `/mcp` request into that single transport.

Passing a `sessionIdGenerator` puts the transport in **stateful mode**, which is built for a single session:

- Window 1 sends `initialize` → the transport marks `_initialized = true`.
- Window 2 sends `initialize` into the same transport → the SDK returns `-32600 "Server already initialized"` (`@modelcontextprotocol/sdk` `server/webStandardStreamableHttp.js`).

So only the first client ever completes the handshake.

## The fix

Switch the HTTP path to a **stateless per-request transport** (`sessionIdGenerator: undefined`): each `POST /mcp` builds a fresh `McpServer` + `StreamableHTTPServerTransport`, connects, handles the single request, and tears down on response close. Every client initializes independently. The SDK explicitly requires a fresh transport per request in stateless mode.

This is safe because server→client events do **not** travel through the MCP transport — task updates are pushed via the API broadcast endpoint (`/tasks/:id/broadcast`), so no session-scoped SSE or session tracking is needed (a `Map<sessionId, transport>` would be unnecessary).

**Key detail:** the shared `RateLimiter`/`ToolContext` is **hoisted to startup scope** and injected into every per-request server. `RateLimiter` holds stateful in-memory token buckets; rebuilding it per request would reset the buckets and silently disable rate limiting.

## Refactor

- **New** `packages/mcp/src/server.ts` — pure, import-safe factories (`createToolContext`, `createMcpServer(context)`, `createMcpHttpHandler(env, context)`).
- `packages/mcp/src/index.ts` — thin process bootstrap (`startStdio` / `startHttp` / `main` + graceful shutdown) delegating to `server.ts`; `import "./stdioEnv.js"` kept as the first line so stdio stdout stays clean for JSON-RPC.
- The `stdio` transport (default for local Claude Code) is unchanged.

## Tests

New `packages/mcp/src/__tests__/httpTransport.test.ts`:

- **two independent `initialize` calls both succeed with no `-32600`** (the core regression);
- `/health` returns `200 {"status":"ok"}` (Docker healthcheck contract);
- unknown path → `404`;
- shared, stateful `RateLimiter` (guards against a per-request limiter reset).

`@aif/mcp`: 85 tests pass, `tsc` clean, ESLint clean, coverage Functions 100% (≥ 70% threshold).

## Docs

`docs/mcp-sync.md` → "HTTP — Docker / remote" now documents concurrent multi-client support.

## Verification note

The automated multi-session test covers the core regression at the handler level. End-to-end confirmation additionally benefits from restarting the running MCP process (it was launched without `--watch`) and opening two Claude Code windows, both connecting without `-32600`.

### DIFF
--- a/.ai-factory/patches/2026-07-10-16.22.md
+++ b/.ai-factory/patches/2026-07-10-16.22.md
@@ -1,0 +1,75 @@
+# Gate MCP HTTP multi-session behind an off-by-default env flag
+
+**Date:** 2026-07-10 16:22
+**Files:** packages/mcp/src/env.ts, packages/mcp/src/server.ts, packages/mcp/src/__tests__/httpTransport.test.ts, docs/mcp-sync.md, .env.example
+**Severity:** medium
+
+## Problem
+
+PR #146 fixed the MCP HTTP transport so multiple clients (Claude Code windows)
+can connect, by replacing the single process-wide stateful transport with
+stateless per-request server/transport construction. Code review (REQUEST_CHANGES)
+flagged three issues:
+
+1. **Must fix:** the `/mcp` path unconditionally flipped every HTTP deployment to
+   the new stateless behavior — a non-trivial change to an external MCP transport
+   contract with no rollout guard.
+2. **Should fix:** the stateless path accepted any HTTP method, so the SDK
+   client's optional `GET /mcp` SSE stream held an idle server/transport open per
+   client for events this server never pushes through the transport.
+3. **Should fix:** the regression test only proved two `initialize` requests stop
+   colliding — it never proved a client can `initialize` then actually
+   `tools/list`/`tools/call` through the stateless path.
+
+## Root Cause
+
+1. The fix optimized for correctness but ignored the project rollout rule: a risky
+   behavior change on an external contract must default to the previous behavior.
+   Shipping the new transport as the unconditional default gives operators no safe
+   fallback if the stateless path misbehaves in their environment.
+2. Routing sent all methods into a fresh transport. In stateless request/response
+   usage there is no reason to serve `GET` SSE — events are delivered out-of-band
+   via the API broadcast endpoint (`utils/broadcast.ts`), so an accepted GET is a
+   pure resource leak.
+3. The test asserted the absence of `-32600` but not the presence of working tool
+   calls, so a transport change that fixed init while breaking real MCP usage
+   would pass.
+
+## Solution
+
+- Added `AIF_MCP_HTTP_MULTI_SESSION_ENABLED` (default **false**) parsed once in
+  `loadMcpEnv` (`env.ts`) using the shared `1/true/yes/on` boolean convention.
+- `createMcpHttpHandler` now selects the `/mcp` behavior once at build time:
+  - flag **true** → `createStatelessMcpDispatcher`: fresh server+transport per
+    `POST`; non-POST methods get `405 Allow: POST` (no idle SSE stream).
+  - flag **false** → `createSingleSessionMcpDispatcher`: one shared stateful
+    transport (`sessionIdGenerator: () => randomUUID()`) connected lazily once —
+    preserves the exact previous behavior (2nd client still gets `-32600`).
+- Test suite exercises both paths: the multi-session suite adds `initialize →
+  tools/list` (proves real usage) and `GET /mcp → 405`; a new legacy suite proves
+  the flag-off path still returns `-32600` for the 2nd client.
+- Documented the flag in `.env.example` and `docs/mcp-sync.md` (prose + env table).
+
+Verified: `@aif/mcp` tests (88 passing), lint, `format:check`, and monorepo
+`build` (typecheck) all green.
+
+## Prevention
+
+- **Gate external-contract behavior changes behind an off-by-default env flag.**
+  When a change alters how an outside client talks to us (transport, protocol,
+  API shape), keep the old behavior as the default and make the new behavior an
+  explicit opt-in. Flip the default only after the new path has soaked.
+- **Regression tests must assert real end-to-end usage, not just the absence of
+  the reported error.** For a protocol/transport fix, drive the full client
+  sequence (`initialize` → `tools/list`) so a fix that trades one break for
+  another is caught.
+- **SDK note (MCP `@modelcontextprotocol/sdk` 1.29.0):** stateless mode
+  (`sessionIdGenerator: undefined`) needs a fresh transport per request (reuse
+  causes message-ID collisions). `validateSession` early-returns when session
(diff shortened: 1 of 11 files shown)


lee-to:
## Code review

This PR fixes the MCP HTTP transport by replacing the single process-wide stateful transport with stateless per-request server/transport construction, extracts import-safe factories, and adds regression coverage. The core POST implementation matches the SDK stateless example and the reported CI checks are green, but this is still a behavior change on an external MCP transport path and needs the project-required rollout guard before I would merge it.

### Must fix

1. **Gate the new HTTP transport behavior** - `packages/mcp/src/server.ts:93` - the `/mcp` HTTP path now unconditionally switches every HTTP deployment to the new stateless per-request behavior. This is a non-trivial external contract change on the MCP hot path, and the review policy requires risky behavior changes to be behind an environment-driven feature flag that defaults off. Add a flag such as `AIF_MCP_HTTP_MULTI_SESSION_ENABLED=false`, load it once at startup through the MCP env config, document it in `.env.example` and `docs/mcp-sync.md` or `docs/configuration.md`, and branch the HTTP startup/handler so operators can enable the new stateless path intentionally while the false path preserves the previous behavior.

### Should fix

1. **Do not leave optional GET SSE streams open when the server says events are out-of-band** - `packages/mcp/src/server.ts:93` and `docs/mcp-sync.md:64` - all `/mcp` methods currently flow into a fresh SDK transport. The SDK client attempts an optional `GET /mcp` SSE stream after the initialized notification, and treats `405` as the expected "server does not offer SSE" response. Since the docs state server-to-client events use the API broadcast path and each MCP request should be short-lived, accepting GET creates an idle server/transport per connected client for no benefit. Restrict the stateless path to `POST /mcp` and return `405` for `GET`/`DELETE`/unsupported methods, or explicitly document and test why the long-lived GET stream is intentional.

2. **Cover a real post-initialize tool flow** - `packages/mcp/src/__tests__/httpTransport.test.ts:102` - the regression test proves two independent `initialize` requests no longer return `-32600`, but it does not prove that a client can initialize and then successfully call/list tools through the stateless path. Add a test for the normal sequence (`initialize`, initialized notification if needed, then `tools/list` or a read-only `tools/call`) so the suite would catch a transport change that fixes initialization but breaks actual MCP usage.

### Nits

None.

### Context gates

- Architecture: pass - MCP remains isolated in `@aif/mcp`; data access still flows through existing tool registrations and `@aif/data`.
- Rules: error - risky external transport behavior is not behind an off-by-default feature flag.
- Roadmap: pass - this is a bug fix under the completed Handoff/AIF sync area, and the PR explains that linkage.
- CHECKLIST compliance (touched packages: `packages/mcp`): pass with one test-depth caveat above - no tool contract drift, no new direct DB access, tool validation unchanged, and MCP lint/test checks are green in CI.
- Docs sync: pass - `docs/mcp-sync.md` was updated for concurrent HTTP clients.
- PR size (926 total changed lines / 8 files; effective review size 377 changed lines / 4 files after excluding generated `.ai-factory` plan/QA artifacts): pass - no decomposition needed.
- Risk gating (feature flag): error - add an off-by-default `AIF_MCP_HTTP_MULTI_SESSION_ENABLED` style flag and document the rollout.

### Positive notes

The extraction into `server.ts` makes the process entry import-safe for tests, and the shared `RateLimiter` assertion is a good guard against accidentally resetting buckets per request.

Verdict: REQUEST_CHANGES until the rollout flag is added.

lee-to:
## Code review — follow-up

This is a response pass over the commits added since the [previous review](https://github.com/lee-to/aif-handoff/pull/146#issuecomment-4894470212). The rollout concern and both follow-up requests are addressed, and I found no new blocking issues in the delta.

### Response pass

- Addressed: ✔ **Off-by-default rollout guard** — commit `9b11317` adds `AIF_MCP_HTTP_MULTI_SESSION_ENABLED`, parses it once during MCP environment loading, defaults it to `false`, and preserves the legacy stateful path when disabled.
- Addressed: ✔ **No idle GET SSE transport** — `packages/mcp/src/server.ts:138` restricts the stateless path to `POST` and returns `405 Allow: POST` for unsupported methods, with regression coverage at `packages/mcp/src/__tests__/httpTransport.test.ts:150`.
- Addressed: ✔ **Post-initialize tool flow coverage** — `packages/mcp/src/__tests__/httpTransport.test.ts:126` now verifies `initialize` followed by a successful `tools/list`.
- New: none.
- Still unresolved: none.

### Must fix

None.

### Should fix

None.

### Nits

None.

### Context gates

- Architecture: pass — the MCP package remains within its existing boundaries, and the delta introduces no direct database access.
- Rules: pass — the external transport behavior is now guarded by an opt-in flag whose disabled path preserves existing behavior.
- Roadmap: pass — this remains a focused bug fix within the completed Handoff ↔ AIF sync milestone.
- CHECKLIST compliance (touched package: `packages/mcp`): pass — tool contracts and validation are unchanged; transport regression tests were added; MCP unit tests and ESLint pass in CI.
- Docs sync: pass — the flag, default, rollout intent, and POST-only stateless behavior are documented in both `.env.example` and `docs/mcp-sync.md`.
- PR size (1,197 total changed lines / 11 files; about 573 effective changed lines / 6 implementation, test, and documentation files after excluding AI Factory plan, QA, and patch artifacts): pass — this is one cohesive transport fix, and splitting the flag, compatibility path, tests, and docs would reduce reviewability.
- Risk gating: pass — `AIF_MCP_HTTP_MULTI_SESSION_ENABLED` defaults to `false`, is read once at startup, and leaves the existing single-session behavior on the off path.
- CI: pass — Build, ESLint, MCP Unit, Tests, mcpunit, and npm audit are green.

### Positive notes

The legacy-path regression test makes the rollback behavior explicit, while the new `initialize → tools/list` assertion covers actual protocol use rather than only the original error signature.

Verdict: **APPROVE — LGTM now.**

PR #149, opened 2026-07-11
feat(agent): enable cross-project coordinator lanes

### LINKED ISSUE #145: [feature] Coordinator serializes tasks across different projects — no cross-project parallelism
### Summary

The agent coordinator runs at most `COORDINATOR_MAX_CONCURRENT_TASKS` (default **3**) tasks at a time **across all projects combined**, and processes pipeline stages **sequentially within each poll cycle**. As a result, tasks from different, fully independent projects block/wait on each other even though each project has its own working directory and could run safely in parallel.

Verified on `main` @ `5160dd6`.

### Impact

- With N projects each having a ready task, only 3 run at once; the rest wait — regardless of project isolation.
- Worse (cross-stage head-of-line blocking): a long-running task in an **early** stage of project A blocks a task in a **later** stage of project B for the whole poll cycle — up to `AGENT_STAGE_RUN_TIMEOUT_MS` (**default 1 hour**). E.g. project A's `planner` stalls project B's `reviewer`.
- Because the poll scheduler is single-flight, while a cycle is draining a long task no new cycle starts, so throughput collapses to roughly one busy stage at a time.

### Root cause (`packages/agent/src/coordinator.ts`)

Four mechanisms combine; all are **global**, none are per-project:

1. **Single global concurrency pool.** `StageSemaphore` is one module-level instance (`coordinator.ts:111-141`, instance at `:141`), keyed only by stage label; `totalActive()` sums across everything with **no project dimension**. `globalMax = env.COORDINATOR_MAX_CONCURRENT_TASKS` (`coordinator.ts:789`; default `3`, `min 1 max 10` at `packages/shared/src/env.ts:111`). Enforced at `coordinator.ts:827-834` (skip stage when `totalActive >= globalMax`) and `coordinator.ts:921-928` (break candidate loop after claim when full). All projects share one pool of 3 — two projects don't each get 3, together they get 3.

2. **Sequential per-stage loop = cross-project head-of-line blocking (the severe part).** `for (const stage of PIPELINE)` (`coordinator.ts:824`) processes stages `planner → plan-checker → implementer → reviewer` strictly in order, and `await Promise.allSettled(spawned)` (`coordinator.ts:958`) fully drains every task spawned in a stage before advancing. `processOneTask` (`coordinator.ts:329`) → `runStageWithTimeout` → `withTimeout(runner(...), STAGE_RUN_TIMEOUT_MS)` (`coordinator.ts:169-173`) **awaits the full agent run** (default 1h, `AGENT_STAGE_RUN_TIMEOUT_MS` at `env.ts:53`). So a task in an earlier stage of one project blocks later-stage tasks of every other project within the cycle.

3. **Single-flight poll scheduler.** `packages/agent/src/pollScheduler.ts:21-31` — `if (isRunning) return;` drops overlapping timer ticks, so while a cycle is draining a long task, no new cycle runs (`POLL_INTERVAL_MS` default 30s). Only the WS "wake" path (`packages/agent/src/index.ts`) calls `pollAndProcess` outside this guard.

4. **Serial-by-default project config.** `projects.parallelEnabled` defaults `false` (`packages/shared/src/schema.ts:15`); `AIF_TASK_WORKTREES_ENABLED` defaults `false` (`env.ts:165`). `resolveProjectConcurrency` (`coordinator.ts:798-822`) only ever limits a project **downward** (`max: 1` unless parallel+worktrees). The per-project spawn counter and `hasActiveLockedTaskForProject` (`coordinator.ts:885`; `packages/data/src/index.ts:1536-1548`) restrict **within** a project — they never isolate one project from another's contention for the shared global pool.

Additionally, candidate selection is globally ordered (`findCoordinatorTaskCandidates`, `packages/data/src/index.ts:1346-1375`: `orderBy(position, createdAt)` with no `projectId` partition, small window ≈15), so one busy project can crowd others out of the candidate window.

### Why cross-project parallelism is safe (and worth fixing)

Within a single project, concurrent tasks mutate one shared working dir (`rootPath`) and would corrupt each other — that's why within-project parallelism is correctly gated behind worktrees (`executionRoot = task.worktreePath ?? project.rootPath`, `coordinator.ts:378`). But **different projects have different `rootPath`s**, so cross-project parallelism needs no worktrees. The current design penalizes safe cross-project parallelism as if it were unsafe within-project parallelism.

### Not the cause (to preempt)

The QA "serialize concurrent runs" logic from PR #130 (`tryStartQaRun`, `packages/data/src/index.ts`) is **per-task** (`WHERE id = ? AND qa_status != 'running'`) and never enters the coordinator concurrency path — it is unrelated to this bug.

### Reproduction

Default config (`COORDINATOR_MAX_CONCURRENT_TASKS=3`, `parallelEnabled=false`, worktrees off):

- **Global cap:** create 5 projects, one ready task each → at most 3 run simultaneously; the other 2 wait despite full isolation.
- **Cross-stage head-of-line:** project A with a slow task in an early stage (`planning`/`implementing`) + project B with a task in `review` → B's `reviewer` does not start until A's earlier stage fully drains (observe the single poll cycle awaiting stage-by-stage).

### Suggested fix (for maintainers to decide)

Make each project effectively its own **lane**:

1. **Two-level concurrency governor** instead of one global stage-keyed semaphore: a **per-project** cap plus a global safety ceiling. Suggested env: `COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT` (per-project, default 3), add `COORDINATOR_MAX_CONCURRENT_PROJECTS` (lane cap, default e.g. 4), and repurpose `COORDINATOR_MAX_CONCURRENT_TASKS` as a raised global ceiling. Provide `COORDINATOR_MAX_CONCURRENT_PROJECTS=1` as an escape hatch to restore current behavior.
2. **Run project lanes concurrently:** split `pollAndProcess` so each actionable project runs its own PIPELINE pass via `Promise.allSettled(lanes)`; keep the per-stage `await` **inside** a lane only (preserves within-project serial-unless-worktrees safety) so projects no longer head-of-line-block each other.
3. **Per-project candidate query** (`findCoordinatorTaskCandidatesForProject`) + a `listActionableProjectIds(limit)` selector ordered by oldest waiting task, so no project starves another's candidate window.

Keep `parallelEnabled` as an independent **within-project** flag (requires worktrees); cross-project lanes should work without it.

**Config-only partial mitigation (not a real fix):** raising `COORDINATOR_MAX_CONCURRENT_TASKS` toward 10 lets more same-stage tasks run together, but does **not** fix cross-*stage* head-of-line blocking and increases resource/API usage.

### Environment

- Repo `lee-to/aif-handoff`, branch `main` @ `5160dd6`
- Node/TypeScript monorepo (Turborepo); coordinator in `packages/agent`

lee-to:
## Summary

- process independent project lanes concurrently while preserving stage order inside each project
- keep separate global task, project-lane, and per-project task limits
- select actionable projects by the oldest waiting task and query stage candidates per project
- document the new concurrency settings and add regression coverage for cross-stage blocking, lane capacity, global safety, fairness, and empty results

## Configuration

- `COORDINATOR_MAX_CONCURRENT_TASKS=12` sets the global safety ceiling
- `COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT=3` limits parallel work within one project stage
- `COORDINATOR_MAX_CONCURRENT_PROJECTS=4` limits independent project lanes per poll cycle

## Validation

- `npm run ai:validate`
- all package coverage thresholds remain above 70%
- all three k6 load scenarios passed
- Codex app-server protocol artifacts are in sync

Related issue: [#145](https://github.com/lee-to/aif-handoff/issues/145)

Closes #145


ichinya:
## PR #149 - review summary

**Verdict: `REQUEST_CHANGES`.**

The current head 4b0b95f7adcd62bceba20c79b95c1637b0b1868c has been reviewed.

I attempted to submit a GitHub review with line-by-line comments and then leave a regular comment on the PR. GitHub rejected both actions with a `403 Resource not accessible by integration`, so **nothing was published in the PR itself**. Below is the completed review text.

---

## Overall assessment

The architectural direction is correct:

* project-scoped selection eliminates the displacement of some projects by others from the global candidate window;
* moving `PIPELINE` inside the project lane removes cross-project head-of-line blocking;
* separating global, per-project, and project-lane limits makes the configuration model clearer;
* the data layer, documentation, and env-mocks have been modified in a consistent manner.

Before merge, two scheduler invariants must be closed.

### [P1] Global task slots are allocated to lanes greedily

The problem area is the calculation of `available` and the immediate `continue` if the global slots are occupied:

```ts
const available = stageSemaphore.available(stageKey, projectMax, globalMaxTasks);
if (available <= 0) {
  log.debug({ stage: stage.label, projectId }, "Project stage at capacity, skipping");
  continue;
}
```

`projectIds.map(processProjectLane)` calls async-functions from left to right. Each lane before the first `await` manages to synchronously:

1. get candidates;
2. take up to `projectMax` slots of semaphore;
3. start tasks;
4. only then stop at `await Promise.allSettled(spawned)`.

 Therefore, the first lane can take the entire global capacity. Subsequent lanes will see `available === 0`, go through `continue` at each stage, and complete the entire poll cycle.

 The specific configuration is as follows:

```text
COORDINATOR_MAX_CONCURRENT_TASKS=2
COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT=2
COORDINATOR_MAX_CONCURRENT_PROJECTS=2
```

With two parallel projects with two planning-task each, both planner-tasks will be taken from the first project. The second selected project will not start at all.

 The current test does not detect this: it only checks the total number of `runPlanner` calls, but not the project IDs to which these calls belong. Therefore, two calls from the same project pass the test.

This returns cross-project starvation — the same class of problem that a PR should fix. With a large or constantly growing backlog of one old project, other projects may be repeatedly skipped.

**What to change:** the global governor should be work-conserving and fair. Suitable options include:

* FIFO async semaphore, where the lane waits for a slot to become available instead of completing the current pass;
* round-robin distribution: first, no more than one task per selected lane, then issuing additional slots up to `projectMax`;
* a central queue of runnable tasks that maintains project-level fairness.

 A minimum regression test should use deferred planner mocks and verify that before the promises are resolved:

```ts
expect(startedProjectIds).toEqual(new Set(["project-1", "project-2"]));
expect(peakActive).toBeLessThanOrEqual(2);
```

### [P1] Project lane is not reserved between overlapping poll cycles

`COORDINATOR_MAX_CONCURRENT_PROJECTS` only limits the list selected by a single call to `pollAndProcess()`. There is no shared set of active project IDs or a shared single-flight queue.

 However, the cron scheduler only protects its own ticks through the local `isRunning`. The wake-path bypasses this scheduler and calls `pollAndProcess()` directly.  The debounce wake channel is set to two seconds and only limits the frequency of events, but does not prevent long cycles from crossing.

Possible scenario:

1. cycle A starts the slow planner of project P;
2. after more than two seconds, a wake is received or a cron tick starts;
3. cycle B also enters `pollAndProcess()`;
4. the locked planner is not included in the sample, but project P has an unlocked review-task;
5. cycle B reaches the reviewer;
6. The `P:reviewer` key is independent of the `P:planner`, so the reviewer runs before the planner completes.

 For a parallel project, the `hasActiveLockedTaskForProject()` check is intentionally not applied, so it does not stop this scenario.

 This is inconsistent with the documented guarantee:

> The project planner completes before the reviewer of the same project.

For the same reason, the actual number of concurrently active project lanes may exceed `COORDINATOR_MAX_CONCURRENT_PROJECTS`: each overlapping call selects its own full set of projects.

**What to change:** The preferred option is to run cron and wake through a single coordinator scheduler:

* one active `pollAndProcess`;
* wake during an active cycle sets `pending=true`;
* another cycle is executed after completion if pending was set.

An alternative is to reserve active project IDs at the module level and release them in `finally`, but this requires a global project lanes counter.

 A test is needed with two concurrently running `pollAndProcess()`:

```text
cycle A: slow planner, project P
cycle B: ready reviewer, project P
```

The reviewer should not start until the planner is complete, provided that the declared stage-ordering semantics are maintained.

If the cross-stage overlap of a single parallel project is indeed a deliberate contract, the documentation should be rewritten: it currently promises a stronger guarantee than the implementation provides.

### [P2] Tests do not prove the actual parallelism of lanes

The test with four projects does `await pollAndProcess()`, and then only checks that four calls have ever occurred. Since all runners immediately return resolved promises, a fully sequential implementation would also pass this test.

The test should be made barrier-based:

1. each planner writes the project ID to `started`;
2. the planner remains pending;
3. the test waits for four planners to be called;
4. it asserts that all four have started before at least one is resolved;
5. it then releases promises and completes `pollAndProcess()`.

 Similarly, the global cap test should check the following simultaneously:

* the peak active count;
* the distribution of launches between project IDs;
* the absence of missed selected lanes.

### [P3] `StageSemaphore` accumulates zero project-stage keys

 After switching from a fixed stage key to `${projectId}:${stage}`, the `release()` method continues to store the value `0`:

```ts
release(key: string): void {
  const current = this.counts.get(key) ?? 0;
  this.counts.set(key, Math.max(0, current - 1));
}
```

Now `counts` turns into a history of all projects and stages ever processed by the process. When creating and deleting a large number of projects:

* The map grows indefinitely;
* `totalActive()` scans long-inactive keys on each check.

Fix:

```ts
release(key: string): void {
  const current = this.counts.get(key) ?? 0;

  if (current <= 1) {
    this.counts.delete(key);
    return;
  }

  this.counts.set(key, current - 1);
}
```

For `totalActive()` it is also reasonable to maintain a separate numeric counter, so as not to perform a full Map traversal on every `available()` and `tryAcquire()`.

## Configuration compatibility

Need an explicit migration note.

 Previously documented parameter:

```text
COORDINATOR_MAX_CONCURRENT_TASKS=3
```

meant per-project/stage concurrency. After the update, the same existing setting becomes a global ceiling for all projects. Therefore, a deployment with a saved `.env` will remain with a total capacity of `3`, despite the new default of `12`, and the declared cross-project parallelism will be significantly limited.

 You should explicitly write what the operators need:

```text
COORDINATOR_MAX_CONCURRENT_TASKS=12
COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT=3
COORDINATOR_MAX_CONCURRENT_PROJECTS=4
```

## CI

`Build`, `Lint`, `Tests`, `Security Audit`, and `MCP Unit` completed successfully.

 The failure of `AI Review` is not related to the PR changes. Webhook was not sent due to TLS hostname mismatch for `ai-review-runner.cutcode.dev`:

```text
SSL: no alternative certificate subject name matches target host name
```

**Result:** The solution should be improved before merge. I consider fair distribution of global capacity and general coordination of overlapping poll cycles to be blocking.

ichinya (changes requested):
The solution should be improved before merge. I consider fair distribution of global capacity and general coordination of overlapping poll cycles to be blocking.

lee-to:
@ichinya Thank you for the detailed review. I addressed the findings in `0daf2b2`.

- **Global capacity fairness:** `StageSemaphore` now uses an async FIFO permit queue. Selected project lanes wait for global capacity instead of completing the cycle without work, and the first wave is distributed across runnable projects while respecting the global cap.
- **Overlapping poll cycles:** `pollAndProcess()` is now single-flight across both cron and WebSocket wake paths. A trigger received during an active cycle is coalesced into one follow-up cycle, so a later cycle cannot overtake an earlier stage in the same project.
- **Concurrency regressions:** the tests now use deferred/barrier runners and verify project distribution, peak active count, real lane overlap, and same-project stage ordering across concurrent poll requests.
- **Semaphore cleanup:** inactive project-stage keys are deleted on the final release, and the total active count is maintained in O(1) time.
- **Configuration compatibility:** the upgrade note now explains that an existing `COORDINATOR_MAX_CONCURRENT_TASKS=3` becomes a global ceiling and recommends explicitly setting all three coordinator limits.

Validation completed successfully with `npm run ai:validate`. The new GitHub checks are also green: Build, ESLint, MCP Unit, Tests, mcpunit, and npm audit.

Please take another look when you have a chance.

ichinya:
## Overall update assessment

The follow-up commit has significantly improved the implementation and addressed all the comments from the previous review.

 The following issues have been resolved:

* Instead of immediately skipping the lane when the global cap is occupied, a waiting permit governor has been added;
* The first wave of permits is now distributed among independent lanes, and the test checks not only the total number of launches but also the project IDs along with the peak concurrency;
* cron and WebSocket wake now use a shared single-flight drain, and repeated triggers are rolled back in the follow-up cycle;
* The test of four project lanes has become barrier-based and actually confirms that all four planners are running before the promises are released;
* `StageSemaphore.release()` removes inactive keys, and `totalActive()` no longer scans the entire Map;
* A correct migration note has been added for the changed semantics of `COORDINATOR_MAX_CONCURRENT_TASKS`.

 The architecture has become significantly more stable. However, the new order of `permit → claim` creates two blocking defects.

---

## [P1] Pending permit keeps stale candidate snapshot

### Problem

Now coordinator performs the following actions in this order:

1. gets `TaskRow` from candidate query;

2. calculates project concurrency;

3. selects runtime profile;

4. evaluates proactive runtime gate;

5. waits for `stageSemaphore.acquire(...)`;

6. calls `claimTask(...)`.

Waiting for a permit can now take as long as another agent stage takes, potentially for dozens of minutes. During this time, the original `TaskRow` may no longer be actionable:

* The user may pause the task.
* The task may be moved to `done`, `backlog`, or another status.
* The `autoMode` may change for `plan_ready`.
* The runtime-limit snapshot may change.
* Another coordinator process may start a task for the same non-parallel project.

Human state transitions are performed through the API independently of the active coordinator poll and additionally send a wake signal.

After waiting, `claimTask()` only checks `task.id` and the availability of the lock. It **does not check** the following:

* the expected `projectId`;
* the expected `status`;
* `paused = false`;
* the expected `autoMode`.

 After a successful stale claim, `processOneTask()` uses the old `TaskRow` and sets `stage.inProgress` without CAS. This allows the coordinator to cancel a later user action and launch a runner for an outdated stage.

### Reproducible scenario

```text
COORDINATOR_MAX_CONCURRENT_TASKS=1
COORDINATOR_MAX_CONCURRENT_PROJECTS=2
```

1. Project A's planner receives a single global permit and runs for a long time.
2. Project B's review-task has already selected a candidate query and is waiting for a permit.
3. The user either completes project B's review-task or pauses it.
4. Project A's planner ends.
5. Lane B receives a permit.
6. `claimTask()` successfully sets the lock because the lock is free.
7. `processOneTask()` returns the task to `review` and starts the reviewer with an outdated snapshot.

 For `paused=true`, the situation is even worse: the paused flag will remain set, but the runner will still be started.

### Required fix

After receiving a permit, it is necessary to re-verify the eligibility of the task. The most reliable option is a coordinator-specific CAS claim with the following conditions:

```ts
task.id === expectedTaskId
task.projectId === expectedProjectId
task.status === expectedStatus
task.paused === false
task.autoMode === expectedAutoMode // для plan_ready
lock отсутствует или истёк
```

Also after waiting, you should:

* re-evaluate the runtime gate;
* repeat the non-parallel project lock guard;
* if the CAS miss, release the permit without status update, broadcast, or other side effects.

 The code already has a correct example of this approach: `blockTaskForRuntimeGateIfEligible()` checks the expected status, project, paused state, and `autoMode`, and if the candidate changes, simply cancels the operation.

### Required regression test

The test should:

1. set global cap to `1`;
2. hold project A planner via deferred promise;
3. wait for project B task to be queued for permit;
4. change status or `paused` of task B directly in DB;
5. release planner A;
6. confirm that runner B was not called, new state is not overwritten, permit is released.

---

## [P1] Exiting from `claimTask()` permanently loses global permit

### Problem

`StageSemaphore.acquire()` increases the `activeCount` before returning control to the calling code.

 After that, `claimTask()` is called outside of the `try/catch/finally`:

```ts
await stageSemaphore.acquire(stageKey, projectMax, globalMaxTasks);

if (!claimTask(task.id, COORDINATOR_ID, CLAIM_LOCK_DURATION_MS)) {
  stageSemaphore.release(stageKey);
  continue;
}
```

The `false` branch releases the permit, and normal execution passes cleanup to `taskPromise.finally()`. But the synchronous exception from SQLite `.run()` is between these two cleanup paths.

 In this exception:

* `activeCount` remains increased;
* project lane reject-ится;
* permit is not released;
* task lock cleanup is also not performed if the exception occurred after a partial claim;
* queued semaphore waiters do not receive a slot.

 If another lane is already waiting for a permit, the external `Promise.allSettled(projectIds.map(...))` will never complete: the waiter will remain pending, and there will be no one to release the permit.

 As a result:

* `activePollPromise` remains pending forever;
* cron and wake triggers continue to return the same frozen promise;
* `followUpPollRequested` is set, but the follow-up cycle never starts.

 Even if there are no waiting lanes, the leaked `activeCount` will remain after the cycle ends, and one of the next polls will freeze when trying to acquire a global permit.

### Required fix

The entire section after a successful `acquire()` should be exception-safe. For example:

```ts
await stageSemaphore.acquire(stageKey, projectMax, globalMaxTasks);

let cleanupOwnedByTaskPromise = false;
let claimed = false;

try {
  claimed = claimTaskIfEligible(/* expected candidate fields */);

  if (!claimed) {
    continue;
  }

  const taskPromise = processOneTask(task, stage).finally(() => {
    stageSemaphore.release(stageKey);
    releaseTaskClaim(task.id);
  });

  cleanupOwnedByTaskPromise = true;
  spawned.push(taskPromise);
} finally {
  if (!cleanupOwnedByTaskPromise) {
    stageSemaphore.release(stageKey);

    if (claimed) {
      releaseTaskClaim(task.id);
    }
  }
}
```

It's better to move this to a small helper that explicitly passes the ownership permit from the setup phase to the task promise.

### Required regression test

You should mock `claimTask()` so that the first call throws an exception, and check:

```ts
expect(getStageSemaphore().totalActive()).toBe(0);
expect(getStageSemaphore().trackedKeyCount()).toBe(0);
```

After that, the second `pollAndProcess()` should successfully receive the permit and start the task. A test with two lanes should also confirm that the queued lane does not remain pending after the first lane fails.

---

## [P2] Rejected project lanes are diagnostically lost

External call:

```ts
await Promise.allSettled(
  projectIds.map((projectId) => processProjectLane(projectId)),
);

log.debug("Poll cycle complete");
```

does not analyze the returned results. Therefore, exceptions from:

* candidate query;
* project concurrency resolution;
* runtime profile resolution;
* runtime gate;
* claim/setup path

either completely lost or look like a normal successful `Poll cycle complete`.

`Promise.allSettled` is suitable here because an error in one project lane should not cancel the others. But the results should be processed:

```ts
const results = await Promise.allSettled(
  projectIds.map((projectId) => processProjectLane(projectId)),
);

results.forEach((result, index) => {
  if (result.status === "rejected") {
    log.error(
      { projectId: projectIds[index], err: result.reason },
      "Project lane failed",
    );
  }
});
```

This does not replace exception-safe permit cleanup, but it makes the remaining lane-level failures observable.

---

## [P2] Overlap test does not prove that the coalesced follow-up cycle is executed

The current test creates both the planning-task and the review-task **before** the first call to `pollAndProcess()`. Then, a second call is made while the planner is pending.

The problem is that the first `runPollCycle()` will see the `overlap-review-task` itself when it reaches the reviewer stage after the planner. Therefore, the test will pass even with a broken implementation in which the second `pollAndProcess()`:

* returns an active promise;
* does not set `followUpPollRequested`;
* does not run the follow-up pass at all.

 The test actually checks for the absence of overlap, but it does not check for coalescing.

 A more accurate version is:

1. run the first poll with the deferred planner of project A;
2. wait for the planner to start — by this point, the list of `projectIds` of the first cycle is already fixed;
3. create an actionable task in a new project B;
4. call the second `pollAndProcess()`;
5. make sure that B does not run until A is completed;
6. release A;
7. make sure that B has been processed by the follow-up cycle.

This way, the test will capture both guarantees: single-flight and the mandatory execution of one coalesced follow-up.

---

## CI

The following have been successfully completed on the current head:

* `Build`;
* `Lint`;
* `Tests`;
* `Security Audit`;
* `MCP Unit`.

## Summary

This update correctly addresses the previous issues related to fairness, overlapping poll cycles, real test parallelism, cleanup semaphore keys, and configuration migration.

Before the merge, two P1 invariants need to be fixed:

1. **Do not run the task on an outdated candidate snapshot after waiting for the permit;**
2. **Guarantee that the permit is released when any exception occurs between `acquire()` and the cleanup is passed to the task promise.**

After this, the solution will be close to merge-ready.

ichinya (changes requested):
do not run the task on an outdated candidate snapshot after waiting for a permit;
guaranteed to release the permit on any exception between acquire() and passing cleanup to the task promise.

lee-to:
@ichinya Thank you for the follow-up review. I addressed the new findings in `45e66e6`.

- **Stale candidates after permit waits:** added `claimCoordinatorTaskIfEligible`, an atomic CAS that validates project, status, paused state, conditional `autoMode`, and lock availability, then returns the fresh claimed row. The coordinator now repeats the non-parallel guard and runtime-limit gate after receiving the permit.
- **Exception-safe permit ownership:** all setup after `acquire()` is protected by `try/finally`. Permit and owner-scoped claim cleanup remain local until ownership is transferred to the task promise, so claim/setup exceptions release capacity and queued lanes continue.
- **Lane diagnostics:** rejected `Promise.allSettled()` project lanes are now logged with project context.
- **Follow-up-cycle coverage:** the overlap test now creates a new project only after the first cycle project snapshot is fixed, proving that the coalesced follow-up actually runs.
- **Regression coverage:** added stale pause, runtime-gate refresh, throwing claim with a queued lane, later-poll recovery, CAS conflict/fresh-row, and owner-scoped release tests.

Validation: `npm run ai:validate` completed successfully, including full tests, coverage, build, browser performance checks, k6 load checks, protocol sync, and checklist validation. GitHub checks have started for the pushed commit.

Please take another look when you have a chance.

ichinya:
The current head 45e66e6e16a1ac73596bfbecfafca74722008641 has been verified. The PR is open, not draft, and mergeable; it currently has three commits and 17 modified files.

## Overall update assessment

The latest commit correctly addresses almost all of the previous review's concerns.

 Now, the coordinator is:

* after waiting for a permit, performs a CAS-claim with validation of `projectId`, `status`, `paused`, conditional `autoMode`, and lock availability;
* retrieves a fresh `TaskRow` from `UPDATE ... RETURNING`, so downstream execution no longer uses the old candidate snapshot;
* re-checks the non-parallel guard and runtime gate after waiting for a permit;
* keeps the permit cleanup under `try/finally` until ownership is transferred to the task promise;
* logs rejected project lanes instead of silently ignoring them;
* actually checks the coalesced follow-up cycle on a task that was created after the first poll;
* contains regression coverage for paused candidate, repeated runtime-gate evaluation, and permit release after claim exception.

 However, there is still one blocking scheduler-invariant in the lane-level exception handling.

---

## [P1] Project lane completes before the already running tasks if the next candidate setup throws an exception

In each stage, the `spawned` array is created first, and then the coordinator sequentially processes the candidates. The running `processOneTask()` is added to this array, and `Promise.allSettled(spawned)` is called only **after the successful completion of the entire candidate loop**.

 However, the following operations within the loop may throw an exception:

* `hasActiveLockedTaskForProject`;
* `blockCandidateIfRuntimeLimited`;
* `claimCoordinatorTaskIfEligible`;
* other synchronous DB/runtime-selection operations.

The inner `finally` correctly frees the resources of the **current candidate**, but does not catch the error. Therefore, the exception exits from `processProjectLane()` and control never reaches:

```ts
await Promise.allSettled(spawned);
```

After that, the outer `Promise.allSettled(projectIds.map(...))` considers the lane to be completed with `rejected`, logs the error, and allows `runPollCycle()` to end.

At the same time, task promises that were added to `spawned` before the exception continue to run in the background.

### Why is this dangerous

As a result, `pollAndProcess()` can resolve and clear `activePollPromise` while the agent stage is still running.

 The next cron tick, WebSocket wake, or coalesced follow-up cycle can start a new poll. For a parallel project, a new poll can select a different task from the same project at a later stage because:

* the current task has been excluded from the candidate query by its lock;
* guard `hasActiveLockedTaskForProject()` applies only to non-parallel projects;
* the semaphore key is divided by stage: `projectId:planner`, `projectId:reviewer`, and so on.

 Thus, a reviewer or implementer from a new poll can start while a previously running planner is still running. This violates the stated guarantee that stages within a project lane are drained sequentially.

 The global task cap will likely remain intact due to the remaining semaphore permit, but the following will be violated:

* project-local stage ordering;
* single-flight semantics of the actual work being performed;
* the meaning of `COORDINATOR_MAX_CONCURRENT_PROJECTS`;
* the assumption that `pollAndProcess()` only completes after the stages it started have completed.

### Reproducible scenario

```text
parallelEnabled=true
COORDINATOR_MAX_CONCURRENT_TASKS=2
COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT=2
```

1. There are two planning tasks in the same project.
2. The first task successfully claims and starts a long-running `runPlanner`.
3. During the preparation of the second task, `claimCoordinatorTaskIfEligible()` throws an exception.
4. The second task's permit is released.
5. `processProjectLane()` rejects without waiting for the first planner.
6. `runPollCycle()` logs the lane failure and terminates.
7. A new poll starts while the first planner is still pending.

The current claim-exception test does not cover this scenario: the exception occurs on the first claim failed lane, when there are no active tasks in its `spawned` array. It confirms the progress of another project lane, but does not check the drain of the already running work in the failed lane.

### Required fix

Stage-level drain should be placed in a `finally` block, so that the initial error is propagated only after all previously launched task promises have been completed:

```ts
const spawned: Promise<void>[] = [];

try {
  for (const task of candidates) {
    // acquire, revalidate, claim, start task
  }
} finally {
  if (spawned.length > 0) {
    await Promise.allSettled(spawned);
  }
}
```

`finally` will retain the original exception and automatically re-throw it after `allSettled`. Then:

* active tasks are correctly drained;
* permits and claims are released;
* lane error will still be logged in the existing `laneResults` logging;
* `runPollCycle()` will not finish before the work it actually started.

### Required regression test

A test is needed with one parallel lane and two planning candidates:

1. the first `claimCoordinatorTaskIfEligible` returns a task;
2. the first planner remains pending;
3. the second claim throws an exception;
4. it is checked that `pollAndProcess()` is still pending;
5. a review task of the same project is created or added in advance;
6. a second `pollAndProcess()` should not trigger a reviewer;
7. after the planner is released, the first poll completes and subsequent work can continue.

 Minimum validation:

```ts
let pollSettled = false;
void pollPromise.finally(() => {
  pollSettled = true;
});

await new Promise((resolve) => setTimeout(resolve, 20));
expect(pollSettled).toBe(false);
expect(runReviewer).not.toHaveBeenCalled();

resolvePlanner();
await pollPromise;
```

---

## [P2] Post-write exception from claim can still leave task lock

Cleanup closure releases task claim only if `claimedTask` is already assigned:

```ts
if (claimedTask) {
  releaseTaskClaim(claimedTask.id, COORDINATOR_ID);
}
```

Normal SQLite statement errors should roll back the statement, so the main path is safe. But there is still a defensive gap: if `UPDATE ... RETURNING` has already applied a lock, and an exception occurs during subsequent processing or conversion of the result before assigning `claimedTask`, the permit will be released, but the DB lock will remain.

 The new owner-scoped variant `releaseTaskClaim(taskId, coordinatorId)` allows you to safely close this case: it will not clear the lock of another coordinator.

 It is recommended to wrap the claim itself:

```ts
try {
  claimedTask = claimCoordinatorTaskIfEligible(...);
} catch (err) {
  releaseTaskClaim(task.id, COORDINATOR_ID);
  throw err;
}
```

The current test makes a simple `throw` before executing the actual claim, so it only confirms the semaphore cleanup. For full coverage, mock can first call the actual implementation and then throw an exception; after poll, `lockedBy` should be `null`.

This is a less likely path than P1, but the fix is small and consistent with the introduced owner-scoped cleanup model.

---

## What I consider closed

Previous blocking notes are no longer relevant:

* global permits do not discard late selected lanes;
* stale task state is rechecked atomically after waiting;
* runtime gate is reevaluated;
* permit is released on a regular claim exception;
* overlapping polls are combined into a single-flight drain;
* the follow-up cycle is confirmed by a separate test;
* rejected lanes are now observable;
* semaphore keys are cleared.

## CI

On the current head, the following were successfully completed:

* `Build`;
* `Lint`;
* `Tests`;
* `Security Audit`;
* `MCP Unit`.

## Summary

The last commit fixes the previous issues. There is one blocking error-path before approval: **lane is required to wait for all already running tasks, even when the preparation of the next candidate resulted in an exception**.

 After moving the stage drain to `finally` and adding the corresponding regression test, the PR can be approved.

lee-to:
@ichinya Thank you for the follow-up review. I addressed the remaining scheduler invariant and the defensive lock-cleanup gap in `b94c271`.

- **Lane drain on setup failure:** the candidate loop is now wrapped in `try/finally`, and the stage always awaits `Promise.allSettled(spawned)` before the lane exits. The original setup error is still propagated to the existing lane-level diagnostics, but only after every task already started by that stage has completed and released its resources.
- **Ambiguous claim outcomes:** added a narrow `claimOutcomeUncertain` ownership flag around the CAS call. If the claim path throws after writing the lock but before assigning the returned row, cleanup performs an owner-scoped release. The flag is cleared after every normal return, including a CAS miss, so cleanup cannot clear another active claim.
- **Regression coverage:** added a parallel-lane test where the first planner remains pending and the second candidate claim throws. It verifies that the poll stays pending, an overlapping poll remains coalesced, no later review starts early, and subsequent work proceeds after the planner is released. A second test performs the real claim write and then throws, verifying that `lockedBy` and `lockedUntil` are cleared.

Regression-first result: both tests failed on the previous head with an early poll settlement and a persisted lock, then passed after the fix. The complete coordinator suite passes with 66 tests, and the full agent suite passes with 366 tests.

Validation: `npm run ai:validate` completed successfully, including format, lint, all tests, coverage, build, browser performance checks, all three k6 load scenarios, protocol sync, and checklist validation. Agent coverage is 88.21% statements and 89.02% lines. GitHub checks are running for the pushed commit.

Please take another look when you have a chance.

ichinya:
Re-reviewed at b94c2711. The latest update resolves the remaining blocking scheduler issues.

The candidate loop now drains every previously spawned stage task from a finally block before a project lane can exit on a later setup failure. This preserves project-local stage ordering and keeps the shared poll promise active until all work actually started by the cycle has completed.

The ambiguous post-write claim path is also handled correctly: cleanup tracks uncertain claim ownership and performs an owner-scoped lock release, while normal CAS misses cannot clear another coordinator’s claim. The new regression tests exercise both the pending-task drain and a real claim write followed by an exception.

Build, lint, tests, security audit, and MCP unit checks are green. I found no remaining blocking issues in the PR and approve it.

Non-blocking follow-up: project-lane isolation currently assumes project roots are disjoint, but that invariant is not enforced by the project schema or create/update validation. Consider tracking canonical root overlap validation separately.

PR #150, opened 2026-07-11
feat: organize project picker

### LINKED ISSUE #147: [feature] Project picker: deterministic ordering, search, pinning, and grouping
## Problem

I run 15+ projects in a single AIF instance. The project picker renders every project as a flat, unbounded, unordered list, which breaks down badly at that scale. Three distinct problems are tangled together here:

**1. There is no ordering at all.** `listProjects()` (`packages/data/src/index.ts:1297`) is `select().from(projects).all()` with no `ORDER BY`, so the result comes back in SQLite rowid order. Both consumers render it as-is: the picker dropdown (`ProjectSelector.tsx:233`) and the overview grid (`ProjectsOverview.tsx:80,86`). A newly created project lands wherever, and there is no stable mental map of the list.

**2. There is no search in the dropdown.** The Cmd+K palette does include projects (`CommandPalette.tsx:73`), but the dropdown is the primary entry point — it's what you click when you're browsing rather than recalling a name, and it's the surface that visually breaks.

**3. The dropdown has no height bound.** The container (`ProjectSelector.tsx:232`) is `min-w-[280px]` with no `max-h` or `overflow-y`. With 15 projects the tail of the list simply runs off the bottom of the viewport (see screenshot).

Beyond those, there's no way to mark a project as important (pin), and no way to group related projects. My projects are platform slices of the same product — `AI_TRANSLATOR_IOS` / `AI_TRANSLATOR_ANDROID` / `aitranslator.chama.pro`, `Modbus_IOS` / `modbus.chama.pro`, `StorageRadar mac` / `StorageRadar.chama.pro` — so the ones I want adjacent are scattered.

## Proposed Solution

Three stages, independently mergeable. Stage 1 fixes what is arguably a bug and can land on its own.

### Stage 1 — ordering, search, scroll (no migration)

- Give `listProjects()` a deterministic order: ``.orderBy(sql`${projects.name} collate nocase`)``. This fixes the dropdown and the overview grid in one place.
- Add a filter input to the dropdown, autofocused on open, matching `name` and `rootPath` (case-insensitive substring). Reuses the existing `Input` primitive.
- Bound the dropdown: `max-h-[60vh] overflow-y-auto`.
- Add keyboard navigation (↑/↓/Enter/Esc); there is currently none.

### Stage 2 — sort by last activity (no migration)

`projects.updatedAt` is not a proxy for activity — it only moves when project settings are edited, not when agents actually work on tasks. Real activity is `MAX(tasks.updated_at)` per project.

Conveniently, `GET /projects/overview` (`routes/projects.ts:199`) already aggregates tasks per project via `listProjectTaskOverviews()` (`data/src/index.ts:1349`), grouped by `(project_id, status)`. Adding `lastActivityAt: string | null` to `ProjectTaskOverview` (`shared/src/types.ts:201`) is one extra `MAX(tasks.updated_at)` in that existing aggregate — no new column, no migration.

That unlocks sort options: **Name** / **Last activity** / **Active tasks**. The overview screen gets them for free, since it already consumes this endpoint.

### Stage 3 — pin and group (migration required)

The `projects` table (`shared/src/schema.ts:5`) has no pin/archive/group column. This needs an append-only migration at the next free version in `MIGRATIONS`, per the project's migration rule:

- `pinned_at TEXT NULL` — pinned projects float to the top. A nullable timestamp rather than a boolean, so ordering *among* pinned projects is stable and costs nothing extra.
- `group_name TEXT NULL` — flat grouping, rendered as section headers in the dropdown. Deliberately flat rather than a tree.
- `archived_at TEXT NULL` (optional) — hides a finished project from the picker without destroying it and its task history. Right now `Delete` is the only way to get a project out of the list, and it's a destructive action behind a bare `confirm()` (`ProjectSelector.tsx:110`).

The sort key then becomes `pinned_at IS NULL, pinned_at, <selected sort>`.

## Alternatives Considered

**Lean on Cmd+K and do nothing else.** The palette already searches projects, so this is tempting. But it doesn't address ordering, pinning, or the overflowing dropdown, and it only helps when you already know the project's name. Not either/or — the palette entry is worth keeping regardless.

**Store pins and groups in `localStorage` instead of the DB.** Ships fast, no migration. Rejected: AIF commonly runs in Docker with several browsers or machines pointed at one API, so project organization that lives in one browser is more confusing than no organization at all. The picker and the overview screen would also have to share it.

**A boolean `pinned` column** instead of `pinned_at`. Simpler, but provides no ordering among pinned projects, for the same storage cost.

**A virtualized list (react-window).** Rejected — a new dependency to render a list in the tens. `max-h` plus scroll covers it.

**Nested groups / folder tree.** Rejected for now. It needs a separate groups table, drag-and-drop, and a real migration story. A flat `group_name` already covers the observed case: one product with N platform targets.

## Additional Context

Screenshot below — 15 projects, the list runs past the bottom of the viewport, no search field, no discernible order.

lee-to:
## Summary

- return projects in a deterministic case-insensitive order with pinned projects first
- add project search, bounded scrolling, keyboard navigation, pin controls, and flat group sections to the picker
- expose real task activity and add Name, Last activity, and Active tasks sorting to the picker and overview
- persist pin and group metadata through append-only migration v25 and a validated organization endpoint
- broadcast organization updates so other connected clients refresh their project list

## Scope note

The optional archived_at proposal is intentionally deferred. listProjects is also consumed by the coordinator and indexers, so archive semantics need a separate operational design to avoid silently disabling active projects.

## Validation

- npm run ai:validate
- regression test reproduced unordered SQLite results before the fix
- targeted data, shared migration, API, and web component tests
- full monorepo coverage gate, production build, Playwright performance checks, k6 load thresholds, and Codex protocol sync

Closes #147

### DIFF
--- a/.ai-factory/patches/2026-07-11-14.45.md
+++ b/.ai-factory/patches/2026-07-11-14.45.md
@@ -1,0 +1,44 @@
+# Project picker organization did not scale beyond a short project list
+
+**Date:** 2026-07-11 14:45
+**Files:** `packages/shared/src/schema.ts`, `packages/shared/src/db.ts`, `packages/shared/src/types.ts`, `packages/data/src/index.ts`, `packages/api/src/routes/projects.ts`, `packages/web/src/components/project/ProjectSelector.tsx`, `packages/web/src/components/project/ProjectsOverview.tsx`
+**Severity:** medium
+
+## Problem
+
+Projects were returned in SQLite insertion order and rendered in an unbounded,
+mouse-only picker. Larger installations could not reliably find a project,
+navigate the list with a keyboard, pin important projects, group related
+projects, or sort by real task activity.
+
+## Root Cause
+
+The data boundary exposed SQLite's implicit row order instead of defining a
+stable project ordering contract. The picker was originally designed as a
+small CRUD list, so it had no search state, active option, bounded scrolling,
+or persisted organization metadata. Project activity was also unavailable to
+the UI because the overview aggregate did not retain `MAX(tasks.updated_at)`.
+
+## Solution
+
+Added deterministic case-insensitive ordering with an ID tie-breaker, a v25
+append-only migration for `pinned_at` and `group_name`, and a dedicated
+validated organization PATCH endpoint. The existing overview aggregate now
+returns `lastActivityAt`. The picker reuses existing UI primitives for search,
+bounded scrolling, keyboard navigation, pinning, and flat group sections;
+both picker and overview can sort by name, last activity, or active task count.
+
+## Prevention
+
+- Database list functions must define a total order when consumers depend on a
+  stable visual sequence.
+- Large dropdown regressions should cover search, focus, keyboard selection,
+  empty results, and a bounded scroll region.
+- Persist cross-browser organization metadata at the data/API boundary rather
+  than in browser-local storage.
+- Activity sorting must use task activity aggregates, not project settings
+  timestamps.
+
+## Tags
+
+`#projects` `#ordering` `#react` `#accessibility` `#sqlite` `#migration` `#api`
--- a/.ai-factory/patches/2026-07-15-12.09.md
+++ b/.ai-factory/patches/2026-07-15-12.09.md
@@ -1,0 +1,29 @@
+# Project picker navigation and accessibility review fixes
+
+**Date:** 2026-07-15 12:09
+**Files:** packages/web/src/components/project/ProjectSelector.tsx, packages/web/src/__tests__/ProjectSelector.test.tsx
+**Severity:** medium
+
+## Problem
+
+The grouped project picker navigated the pre-grouping project array instead of the rendered section order. Saving a changed group for the selected project invoked `onSelect` twice and produced duplicate browser-history entries. The listbox lacked a complete active-descendant focus model, and user groups named `Pinned` or `Other` collided with built-in React section keys.
+
+## Root Cause
+
+The picker derived rendering and interaction from two different collections: `projectSections` for the DOM and `filteredProjects` for keyboard indices. Section identity was coupled to display labels, even though user labels share the same namespace. Edit success paths independently invoked the navigation callback before and after the organization mutation. Accessibility state also conflated the temporary keyboard highlight with the persisted project selection.
+
+## Solution
+
+Flattened rendered sections into a single `visibleProjects` interaction source, introduced namespaced structural keys, and linked the search combobox to stable listbox option IDs through `aria-activedescendant`. `aria-selected` now reflects `selectedId`, and Escape restores trigger focus from every picker control. The edit callbacks are mutually exclusive when a group changes, so only the final organized project is selected once. Added regression tests for grouped navigation, reserved labels, browser history, ARIA state, and Escape focus restoration.
+
+## Prevention
+
+- Derive keyboard order and DOM rendering from the same final collection.
+- Keep React structural identity separate from user-facing labels.
+- Treat selection callbacks as navigation events and make asynchronous success paths mutually exclusive.
+- Test composite widgets through their focus ownership, active descendant, persisted selection, and Escape behavior.
+- Include user-provided values that match built-in labels in key-collision regression fixtures.
+
+## Tags
+
+`#react` `#accessibility` `#keyboard-navigation` `#browser-history` `#testing`
--- a/.ai-factory/patches/2026-07-16-12.53.md
+++ b/.ai-factory/patches/2026-07-16-12.53.md
@@ -1,0 +1,29 @@
+# Project picker edit callbacks navigated and active indexes drifted
+
+**Date:** 2026-07-16 12:53
+**Files:** packages/web/src/components/project/ProjectSelector.tsx, packages/web/src/__tests__/ProjectSelector.test.tsx
+**Severity:** medium
+
+## Problem
+
+Saving edits to the currently selected project still invoked `onSelect`, which always pushes a browser-history entry. A delayed organization update could also navigate back to the edited project after the user had selected another project. Separately, the combobox stored its active option as an array index, so an asynchronous list reorder could silently change `aria-activedescendant` and the project selected by Enter.
+
+## Root Cause
+
+The edit flow retained a selection callback from an older state model even though the application now derives the selected project from `selectedProjectId` and the invalidated projects query. The delayed callback captured stale selection props, turning a data-update completion into navigation. The combobox also treated a position as option identity even though pinning, grouping, sorting, filtering, and overview refreshes can all reorder the visible array.
+
+## Solution
+
+Removed navigation callbacks from both edit-success paths while retaining query invalidation and contextual `[FIX:pr-150]` organization logs. Replaced the numeric active index state with `activeProjectId`, deriving the current index from the rendered project order. Reordering now preserves option identity, while a missing active ID resets to the first visible project before commit. Added regression tests for zero-history edits, delayed organization responses, unchanged-group edits, and active-option preservation across reorder.
+
+## Prevention
+
+- Keep data-update callbacks separate from navigation callbacks.
+- Never let delayed mutation completion navigate based on captured selection state.
+- Store identity, not position, for active items in asynchronously reordered collections.
+- Test both immediate history effects and late-response races.
+- Reorder a rendered list after keyboard activation when testing `aria-activedescendant` and Enter behavior.
+
+## Tags
+
+`#react` `#combobox` `#async-race` `#browser-history` `#keyboard-navigation` `#testing`
--- a/docs/api.md
+++ b/docs/api.md
@@ -133,6 +133,8 @@ GET /projects
     "planCheckerMaxBudgetUsd": 2,
     "implementerMaxBudgetUsd": 15,
     "reviewSidecarMaxBudgetUsd": 2,
+    "pinnedAt": "2026-01-02T00:00:00.000Z",
+    "groupName": "Platform",
     "createdAt": "2026-01-01T00:00:00.000Z",
     "updatedAt": "2026-01-01T00:00:00.000Z"
   }
@@ -154,6 +156,7 @@ This endpoint does not return full task rows.
 [
   {
     "projectId": "uuid",
+    "lastActivityAt": "2026-01-03T12:00:00.000Z",
     "totalTasks": 12,
     "completedTasks": 2,
     "verifiedTasks": 0,
@@ -195,7 +198,8 @@ This endpoint does not return full task rows.
 that is not `backlog`, `done`, or `verified`. `blockedTasks` counts
 `blocked_external`. `statusPreviews` lists are small (bounded in SQL) and
 include only task id/title pairs — never plan text, logs, or other detail-only
-fields.
+fields. `lastActivityAt` is the latest task `updatedAt` timestamp for the
+project, or `null` when the project has no tasks.
 
 ### Create Project
 
@@ -228,6 +232,30 @@ PUT /projects/:id
 **Body:** Same as Create Project.
 
 **Response:** `200 OK` — the updated project object.
+
+### Update Project Organization
+
+```
+PATCH /projects/:id/organization
+```
+
+Updates picker-only organization metadata without requiring the project's name
+or root path.
+
+**Body:**
+
+| Field       | Type         | Required | Description                                                             |
+| ----------- | ------------ | -------- | ----------------------------------------------------------------------- |
+| `pinned`    | boolean      | no       | Pin or unpin the project. Pinning preserves the original pin timestamp. |
+| `groupName` | string\|null | no       | Flat picker group (max 100 chars); `null` or an empty string clears it. |
+
+At least one field is required.
+
+**Response:** `200 OK` — the updated project object. Returns `404` when the
+project does not exist.
+
+**WebSocket event:** `project:organization_updated` with the full project
+object.
 
 Parallel auto-queue with `git.create_branches=true` requires
 `AIF_TASK_WORKTREES_ENABLED=true`. With the default `false`, the API rejects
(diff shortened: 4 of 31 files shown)


ichinya:
# PR review: `feat: organize project picker` — #150

**Verdict: Request changes.**

The backend part is generally implemented neatly: a total SQL order with `pinnedAt`, case-insensitive name, and ID tie-breaker has been added; `lastActivityAt` is calculated based on real tasks; migration v25 follows the append-only approach; and the PATCH endpoint is validated and sends a WebSocket event.

 However, there are several functional errors in the picker. The first one directly breaks the declared keyboard navigation when using groups, so the PR should not be approved yet.

## 1. P1 — keyboard navigation uses a different order than the visual one

**File:** `packages/web/src/components/project/ProjectSelector.tsx:89–122, 327–350, 400–425`

`selectedProjectIndex`, `projectIndexById`, and arrow and Enter handling work with `filteredProjects`. But the UI is rendered through `projectSections`, which additionally changes the order: first pinned, then groups by name, then Other.

 A reproducible example:

```text
Alpha — group "Z"
Beta — group "A"
Sort — Name
```

This results in:

```text
filteredProjects: Alpha, Beta
DOM: Beta, Alpha
```

When opened, the active element will be `Alpha`, which is the second element in the DOM. After pressing `ArrowDown`, the active element will be `Beta`, which is higher in the DOM. This means that pressing the down arrow will visually move the selection up. Pressing `Enter` will also select an element based on its logical order, rather than its visual order.

 The code explicitly builds indexes based on `filteredProjects` and renders `projectSections`. The arrow and Enter handlers also refer to `filteredProjects`. Although the DOM is built based on the sections, the `projectIndex` is still taken from the original array.

 After building the sections, you need to get a single array in the actual rendering order:

```ts
const orderedProjects = useMemo(
  () => projectSections.flatMap((section) => section.projects),
  [projectSections],
);
```

And use it for:

* ArrowUp / ArrowDown / Enter;
* `projectIndexById`;
* the refs array;
* calculating the active element.

It's more reliable to store the `activeProjectId` instead of a numeric index, so that asynchronous loading of the overview or pin/unpin does not silently switch the activity to another project with the same index.

**Ready-made inline comment:**

> **[P1] Keyboard navigation according to the order of section display**
>
> `selectedProjectIndex` indexes `filteredProjects`, but the DOM is displayed based on `projectSections`, which changes the order of non-page-fixed projects by group. For example, if group `Z` has an element `Alpha` and group `A` has an element `Beta`, then the DOM order will be `Beta, Alpha`, while pressing the down arrow key/Enter will move from `Alpha` to `Beta`; thus, pressing the down arrow key will move the selection up, and pressing the Enter key will select another visual adjacent element. Please combine `projectSections` into a single array and use it for links, navigation, and the Enter key. It is preferable to track the active project by its ID, so that its ID does not change when the review is updated asynchronously. This could be verified using the keyboard-based regression test for grouped navigation.

## 2. P2 - Extra entries appear in the browser history when the group is saved
**Files:**

* `packages/web/src/components/project/ProjectSelector.tsx:267–292`
* `packages/web/src/App.tsx:154–159`

When saving the selected project:

1. the main method `updateProject` completes successfully;
2. `updateOrganization` is triggered;
3. the external `onSuccess` calls `onSelect(project)`;
4. after PATCH, the internal `onSuccess` re-calls `onSelect(organizedProject)`.

Both calls are visible in the same edit flow.

 However, `App.handleSelectProject` is not a harmless object update: it always calls `window.history.pushState`, even if the project ID is already selected.

 As a result, changing the current project's group adds two identical `/project/:id` entries. After saving, the user needs to press Back multiple times to actually leave the project.

 The additional `onSelect(organizedProject)` is unnecessary here. The application stores only `selectedProjectId`, and the actual object is calculated from `projects`. The organization mutation itself already invalidates the query `["projects"]`.

 The minimum fix is to remove the nested `onSuccess` from `onSelect`. A more systematic fix is to make `handleSelectProject` a no-op for navigation when the active ID is selected, or to separate the callbacks `onSelectProject` and `onProjectUpdated`.

**Ready inline-comment:**

> **[P2] Do not select/navigate again after the organization update**
>
> The outer update success already calls `onSelect(project)`, and the new organization success calls `onSelect(organizedProject)` a second time. In `App`, every `onSelect` invokes `history.pushState`, so changing the group of the current project adds duplicate same-URL history entries and makes Back require extra presses. The organization mutation already invalidates `["projects"]`, while the selected project is derived from `selectedProjectId`, so this nested `onSuccess`/`onSelect` is unnecessary. Please remove it or make same-project selection avoid pushing a new history entry.

## 3. P2 — listbox does not have correct focus/active semantics

**File:** `packages/web/src/components/project/ProjectSelector.tsx:374–468`

Focus remains in the regular textbox. The arrows change the React state, but:

* the input does not have `role="combobox"`;
* there are no `aria-controls` or `aria-activedescendant`;
* the listbox and options do not have associated IDs;
* `aria-selected` reflects `isKeyboardSelected`, not the actual selected `selectedId`;
* `active` is set both for the current project and for the keyboard-highlighted project.

 Because of this, when the picker is opened, two projects may be visually highlighted at once: the actually selected one and the element with index `0`. For screen reader, the situation is even worse: DOM-focus does not leave textbox, so changing `aria-selected` on an unbound option does not tell the user which element became active.

In addition, `Escape` is only handled on search input. After Tab on sort, Pin, Edit, or Delete, it no longer closes the picker. Action buttons are also inside the subtree with `role="listbox"`, although they are not options.

 You need one of two standard options:

1. **Combobox pattern:** input с `role="combobox"`, `aria-expanded`, `aria-controls`, `aria-activedescendant`; options с постоянными ID.
2. **Roving focus:** DOM-фокус переводится между option-кнопками, у неактивных options `tabIndex={-1}`.

Также следует разделить:

```ts
const isCurrentProject = project.id === selectedId;
const isActiveOption = project.id === activeProjectId;
```

и не использовать один `active` style для двух разных состояний. Escape лучше обрабатывать на корневом контейнере popup.

**Готовый inline-comment:**

> **[P2] Implement an accessible combobox/listbox focus model**
>
> Arrow navigation currently changes an index while DOM focus remains in a plain textbox. The input has no `aria-activedescendant`/`aria-controls`, and `aria-selected` is assigned to the keyboard-highlighted option rather than the actually selected project. This means assistive technology is not informed when the active option changes; visually, the current project and the keyboard option can also both appear active. Please use either the ARIA combobox pattern with stable option IDs and `aria-activedescendant`, or move focus with a roving tabindex. Escape should also be handled at the popup level so it still works after tabbing to sort or row actions.

## 4. P2 — React keys of sections conflict with valid group names

**File:** `packages/web/src/components/project/ProjectSelector.tsx:105–113, 400–402`

Labels are used for service sections:

```ts
"Pinned"
"Other"
```

And the React key is set like this:

```tsx
key={section.label ?? "projects"}
```

At the same time, the API allows an arbitrary string `groupName` up to 100 characters.

 Therefore, the following scenarios are fully valid:

```text
there are pinned projects + a custom group "Pinned"
there are ungrouped projects + a custom group "Other"
```

In both cases, the sibling sections have the same React keys. The service labels themselves are created next to the custom groups and then used as identities.

 It is necessary to separate the identity of the section from the displayed label:

```ts
{ key: "system:pinned", label: "Pinned", projects: pinned }
{ key: `group:${label}`, label, projects: groupProjects }
{ key: "system:ungrouped", label: "Other", projects: ungrouped }
```

**Ready inline-comment:**

> **[P2] Use collision-free keys for system and user-defined sections**
>
> `section.label` is used as the React key, but `groupName` accepts any string. A user group named `Pinned` collides with the built-in pinned section, and a group named `Other` collides with the ungrouped fallback. React reconciliation with duplicate sibling keys is unsupported and can produce stale or misplaced rows when projects move between sections. Please give each section an internal key such as `system:pinned`, `group:${label}`, and `system:ungrouped`, independent of its display label.

## Non-blocking notes

### Shared `Select` changes globally

PR removes `shadow-md` from the shared `Select`, so not only the new sort controls will change visually, but also the existing Selects in Board, TaskSettings, AddTaskForm, ConfigEditor, RuntimeProfileForm, and other screens.

If this is not an intentional change to the design system, it is better to return shadow to primitive and pass a separate prop/class only to new dropdowns.

### The first debug log is incorrect for a partial PATCH

When updating a group, `patch.pinnedAt` remains `undefined`, so the log shows `pinned: false`, even if the existing project is pinned. Similarly, `groupName` is logged as `undefined` when updating a pin. The endpoint response itself is correct, but the diagnostic log is misleading.

For the log, you should use the final values:

```ts
pinned: input.pinned === undefined
  ? existing.pinnedAt != null
  : patch.pinnedAt != null,

groupName: input.groupName === undefined
  ? existing.groupName
  : patch.groupName,
```

## Tests to add before merge

1. Keyboard navigation with two groups whose order differs from the global sort order.
2. Groups with the names `Pinned` and `Other`.
3. Saving the current project's group without additional `history.pushState`.
4. Testing the combobox/listbox semantics: active descendant, selected state, and Escape after Tab.
5. Component test for sorting switching in `ProjectsOverview`; the comparator is currently tested separately, but the interaction with the new Select is not.

The current keyboard test uses only two ungrouped projects, `Alpha` and `Beta`, so the DOM and index order discrepancy is not detected.

## CI

On the head commit, **Build, Lint, Tests, Security Audit, and MCP Unit** were successfully completed. The only red workflow is **AI Review**; inside it, the `Send signed webhook` step failed, so this is not a product tests or build failure.

ichinya (changes requested):
The backend portion of this PR looks solid: the SQL ordering is total and
deterministic, the v25 migration is append-only, lastActivityAt is based on
real task updates, and the organization endpoint is validated and broadcast
to connected clients.

I found four issues in the picker implementation that should be addressed
before merge:

1. Keyboard navigation indexes filteredProjects, while the DOM is rendered
   from projectSections. Grouping changes the visual order, so ArrowDown can
   move the highlight upward and Enter can select a different visual neighbor.
2. Saving a group for the currently selected project calls onSelect twice.
   Since App.onSelect always calls history.pushState, this creates duplicate
   same-URL browser history entries.
3. The listbox does not implement a complete focus model. Focus remains in a
   plain textbox without aria-activedescendant, aria-selected represents the
   temporary keyboard index rather than the actual project selection, and
   Escape only works while the search input has focus.
4. Built-in sections use their display labels as React keys. Valid user groups
   named "Pinned" or "Other" therefore create duplicate sibling keys.

Please also add grouped keyboard-navigation, reserved-label, browser-history,
and accessibility regression tests. Once these are fixed, the remaining
backend and data-layer changes look ready.

lee-to:
Thanks for the thorough review, @ichinya. I addressed all four project picker issues in commit 3bb4b8c:

- Keyboard navigation and Enter now follow the rendered section order.
- Saving a group for the selected project invokes the selection/navigation callback only once.
- The search now uses a combobox/listbox active-descendant model; aria-selected tracks the actual project, and Escape closes the picker and restores trigger focus from any internal control.
- Built-in and user-group section keys are namespaced, including user groups named Pinned or Other.

I also added regression coverage for grouped navigation, reserved labels, browser history, and accessibility. The full npm run ai:validate suite passes, including lint, tests, coverage, build, Playwright performance checks, k6 load thresholds, protocol sync, and the checklist gate.

Please take another look when you have a chance.

ichinya:
Request changes

The follow-up commit correctly fixes most of the previous review findings:

- keyboard navigation now follows the rendered section order;
- built-in and user-defined sections have collision-free keys;
- the search input implements an active-descendant combobox model;
- aria-selected reflects the actual selected project;
- Escape works from all controls and restores trigger focus.

Two issues remain before this is ready to merge.

1. Editing the current project still invokes the navigation callback.

The change reduces the group-edit path from two onSelect calls to one, but App.handleSelectProject always calls history.pushState. Saving a project that is already selected must add zero navigation entries, not one. The new test currently codifies the incorrect expectation by asserting one pushState call.

The delayed organization callback also creates a race: after the main update closes the dialog, the user can select another project while the PATCH request is still running. When the old request resolves, its captured selectedId === editingId condition calls onSelect and navigates the user back to the edited project.

Both update hooks already invalidate ["projects"], and App derives the selected project object from that query. Please remove onSelect from the edit-success paths, or introduce a separate non-navigation update callback. Add regression tests expecting zero history pushes and ensuring a late organization response cannot restore an old selection.

2. The active combobox option is still stored as an array index.

visibleProjects can reorder when overview data arrives, task activity changes, or pin/group updates refetch the project list. Keeping the same numeric index therefore changes the identity of the active option without a keyboard action; aria-activedescendant and Enter can suddenly reference another project.

Please store activeProjectId and derive its current index from visibleProjects. Only reset it when that ID is no longer visible. Add a test that reorders the list after an option becomes active and verifies that the active descendant and Enter continue to reference the same project.

All current Build, Lint, Tests, Security Audit, and MCP Unit workflows pass.

lee-to:
Thanks for the follow-up review, @ichinya. I addressed both remaining issues in commit de67333:

- Edit success paths no longer call onSelect. Saving either a regular edit or a group change now creates zero navigation/history entries, and a delayed organization response cannot navigate back to the edited project after the user selects another one.
- The combobox now stores activeProjectId and derives its current index from the rendered visibleProjects order. Async pin/group/overview reordering preserves the active option identity; the ID resets only when that project is no longer visible.

I added regression coverage for zero-history edits, the delayed organization race, unchanged-group edits, and preserving aria-activedescendant plus Enter selection across a list reorder.

Validation: npm run ai:validate passes, including 689 web tests, package coverage thresholds, build, Playwright performance checks, all three k6 load suites, protocol sync, and the checklist gate.

Please take another look when convenient.

ichinya:
Approve

Re-reviewed the latest head (de67333).

The follow-up commit correctly resolves the two remaining blocking findings:

- Edit mutations are now data-only and no longer invoke the project-selection  callback. Saving the current project adds no browser-history entries, and a  delayed organization response cannot navigate back to a stale selection.
- The combobox active option is now stored by project ID rather than array  position. Its identity is preserved when pinning, grouping, sorting, or   overview updates reorder the visible list.

The earlier fixes also remain intact: keyboard navigation follows the rendered group order, section keys do not collide with user labels, the combobox is connected to stable option IDs through aria-activedescendant, aria-selected represents the persisted project selection, and Escape restores focus from all picker controls.

The new regression tests cover zero-history edits, delayed organization responses, unchanged-group edits, grouped ordering, reserved section labels, ARIA state, Escape behavior, and active-option preservation across reordering.

All current Build, Lint, Tests, Security Audit, and MCP Unit workflows pass.

No blocking findings remain. This is ready to merge.

Non-blocking follow-ups:
- consider visually distinguishing the persisted selection from the temporary  keyboard-active option and adding project-specific accessible labels to the  Pin/Edit/Delete controls;
- consider logging final project organization values rather than partial patch  fields in updateProjectOrganization;
- confirm that removing the shared Select shadow is an intentional global  design-system change.

ichinya (approved):

PR #151, opened 2026-07-11
fix(web): bound Kanban columns and prioritize recent terminal tasks

### LINKED ISSUE #148: [feature] Terminal columns (done/verified) don't scale — the board grows unbounded, and deleting tasks erases cost history
## Problem

On a long-running project the `verified` column accumulates hundreds of tasks, and the board becomes unusable. Right now the only way to get relief is to delete those tasks — which, as it turns out, destroys data.

**The board has no vertical bound.** The card list inside a column (`Column.tsx:140`) is `min-h-[100px]` with no `max-h` and no `overflow-y`. The board itself (`Board.tsx:173`) is `overflow-x-auto` only. So the page height equals the height of the tallest column. `verified` hits this first, but it isn't specific to `verified` — 30 tasks in `implementing` produce the same problem.

**Terminal columns are sorted by `position`.** `Board.tsx:130` sorts *every* column with `a.position - b.position`. For `backlog` that's correct — it's the manual queue order. For `done` / `verified` it's meaningless: a freshly verified task lands somewhere in the middle of two hundred old ones, so even scrolling to the column doesn't show you what just completed.

**The existing `recent` filter can't be used for this.** It exists (`Board.tsx:95`, 24-hour cutoff), but it applies globally across all columns. Turning it on to tame `verified` also empties your `backlog`.

**Everything is fetched and rendered, always.** `GET /tasks?projectId=…` (`routes/tasks.ts:179`, `api.ts:262`) takes no `status`, `limit`, or `offset`. Every task in the project crosses the wire and gets a DOM node.

**And the workaround is destructive.** `deleteTask` (`data/src/index.ts:1094`) hard-deletes the task row and its comments; nothing is archived. Meanwhile `GET /projects/overview` computes `totalCostUsd`, `totalTokenTotal` and friends by summing over the `tasks` rows (`listProjectTaskOverviews`, `data/src/index.ts:1349`). So every time a user deletes a batch of verified tasks to clean up the board, the project's historical spend silently drops out of the overview. Tidying the board currently costs you your cost history.

## Proposed Solution

Several options, cheapest first. They compose; option 1 on its own already fixes the reported symptom.

### Option 1 — bound the column height (≈1 line, no API change)

Give the card list `max-h-[calc(100vh-…)] overflow-y-auto` so each column scrolls independently and the page stops growing. This is what Trello, Linear and Jira all do. Fixes every column at once, not just `verified`.

### Option 2 — collapsible terminal columns (frontend only)

Let `done` and `verified` collapse to a narrow strip showing just the label and the count, persisted in `localStorage` per project. This is the "hide verified" ask, but non-destructive — and keeping the counter visible means you don't lose the signal that a task reached the end of the pipeline.

### Option 3 — cap what is rendered (frontend only)

In terminal columns, render the most recent N (say 25) and add a `Show 175 more` button. Removes hundreds of DOM nodes that currently render unconditionally on every board paint.

### Option 4 — sort terminal columns by `updatedAt desc` (frontend only)

`position` ordering only makes sense for `backlog`. In `done` / `verified` the newest task should be on top. Pairs naturally with option 3, since "most recent 25" then means something.

### Option 5 — server-side filtering and archival (API + migration)

- Add `status`, `limit`, `offset` to `GET /tasks`, so the client stops pulling every task of the project on every invalidation.
- Add `archived_at TEXT NULL` to `tasks` and an "Archive" action, so users have a non-destructive way to clear the board. This is the one that removes the reason people reach for `Delete` in the first place — and with it, the silent corruption of `/projects/overview` metrics.

A reasonable split: **1 + 4** as a bugfix that can land immediately, **2 + 3** as the feature, **5** once projects reach thousands of tasks.

## Alternatives Considered

**Hide the `verified` column entirely behind a setting.** This was the first thing I reached for, and I don't think it should be the only fix. It treats the symptom rather than the cause (any column can grow), and a fully hidden column means losing the signal that a task actually completed. Option 2 gets the same relief while keeping the counter.

**Make `recent` a per-column filter.** Awkward UI, and it doesn't reduce the DOM in the general case — a column with 200 tasks all touched this week is still 200 cards.

**Virtualize the columns (react-window / tanstack-virtual).** The right answer at thousands of tasks, but it adds a dependency and complicates the reorder controls in `backlog`. Option 3 captures most of the benefit at a fraction of the cost, and can be swapped for virtualization later without changing the API.

**Auto-delete verified tasks older than N days.** Rejected — the same data loss as manual deletion, just automated and harder to notice.

**Keep deleting manually.** The status quo. It works, and it quietly rewrites the project's cost history every time.

## Additional Context

Encountered on a project with hundreds of tasks in `verified`: the board's page height became unmanageable, and deleting the tasks was the only available relief. The cost-history side effect was not obvious until reading `listProjectTaskOverviews`.

lee-to:
## Summary

- bound every Kanban card list to the viewport and give each column independent vertical scrolling
- sort done and verified tasks by updatedAt descending while preserving position ordering for active columns
- add regression coverage for all scrollable columns and terminal/non-terminal ordering

## Problem confirmation

The report was reproducible on main. Card lists only had a minimum height, so the tallest column controlled the page height. Board grouping also sorted every status by position, causing newly completed tasks to appear among older terminal tasks.

The new regression tests failed before the implementation and pass with this change.

## Scope

This PR implements the immediate bugfix split proposed in the issue: options 1 and 4. Collapsible columns, render caps, server-side pagination, and archival remain separate feature work.

## Validation

- npm test --workspace @aif/web -- Board.test.tsx: 11 passed
- npm test --workspace @aif/web: 677 passed
- npm run lint --workspace @aif/web: passed
- web coverage: 90.53% statements, 92.62% lines
- npm run ai:validate: passed
- existing Playwright performance suite: 3 passed

## Playwright MCP verification

A dedicated Playwright MCP scenario expanded the Verified column from 2 to 102 cards in the browser DOM and verified real layout behavior:

- computed overflow-y: auto
- computed max-height: 580px at an 868px viewport
- list client height: 580px
- list scroll height: 12,345px
- independent list scrolling reached scrollTop 11,765px
- document height stayed unchanged at 961px before and after expansion
- rendered Verified order matched updatedAt descending from the API
- no browser console errors

Fixes #148

### DIFF
--- a/.ai-factory/patches/2026-07-16-17.46.md
+++ b/.ai-factory/patches/2026-07-16-17.46.md
@@ -1,0 +1,36 @@
+# Kanban scroll chaining and terminal ordering review fixes
+
+**Date:** 2026-07-16 17:46
+**Files:** packages/web/src/components/kanban/Board.tsx, packages/web/src/components/kanban/Column.tsx, packages/web/src/__tests__/Board.test.tsx, packages/web/e2e/kanban-horizontal-scroll.spec.ts, packages/web/e2e/README.md
+**Severity:** medium
+
+## Problem
+
+PR #151 introduced two user-visible/correctness regressions and a weak test assertion:
+
+- `overscroll-contain` on each vertical card list blocked horizontal wheel and trackpad gestures from chaining to the Kanban board.
+- Terminal tasks with malformed `updatedAt` values could participate in a non-transitive comparator, producing input-order-dependent results.
+- Ordering tests used `textContent.indexOf()`, which could pass when the expected first card was missing.
+
+## Root Cause
+
+The card list applied overscroll containment to both axes even though it owns only vertical scrolling and the parent board owns horizontal scrolling. The terminal comparator also switched between timestamp and position ordering on a pair-by-pair basis whenever either timestamp was malformed, so it did not define a consistent total order.
+
+## Solution
+
+- Replaced `overscroll-contain` with `overscroll-y-contain` so vertical boundary containment remains while horizontal gestures chain to the board.
+- Added a deterministic terminal comparator ordered by timestamp validity, timestamp descending, position ascending, then ID.
+- Replaced string-index assertions with element lookup plus DOM-position comparisons.
+- Added a database-independent Playwright regression using native Chromium wheel input over a genuinely scrollable card list.
+- Removed temporary sorting debug output from the render-time memoization path.
+
+## Prevention
+
+- Scope `overscroll-*` containment to the axis owned by the component when a parent owns the other scroll axis.
+- Define sort comparators lexicographically so every input pair follows the same ordering dimensions.
+- Resolve both expected DOM elements before comparing their order.
+- Use trusted browser input (`page.mouse.wheel`) for scroll-chaining regressions; synthetic DOM wheel events do not exercise native browser scrolling.
+
+## Tags
+
+`#kanban` `#scroll-chaining` `#playwright` `#sorting` `#typescript` `#code-review`
--- a/packages/web/e2e/README.md
+++ b/packages/web/e2e/README.md
@@ -1,8 +1,8 @@
-# Browser performance suite
+# Browser E2E suite
 
-Playwright-driven synthetic perf tests. These exercise the real web stack
-(API + web bundle) against a running dev server and enforce latency budgets
-so regressions fail the suite instead of shipping.
+Playwright-driven browser tests. The suite exercises the real web stack
+(API + web bundle) against a running dev server, covers browser-only interaction
+regressions, and enforces latency budgets.
 
 ## Run
 
@@ -21,6 +21,8 @@ dev shell). Set `AIF_SKIP_DEV_SERVER=1` to bypass the auto-launch entirely and
 
 ## What each spec measures
 
+- `kanban-horizontal-scroll.spec.ts` — verifies horizontal wheel gestures over
+  a vertically scrollable card list still scroll the Kanban board.
 - `perf/dashboard-load.spec.ts` — cold kanban render. Asserts DOM-ready and
   LCP budgets after the first column paints.
 - `perf/runtime-profiles-endpoint.spec.ts` — cold + warm `/runtime-profiles`
--- a/packages/web/e2e/kanban-horizontal-scroll.spec.ts
+++ b/packages/web/e2e/kanban-horizontal-scroll.spec.ts
@@ -1,0 +1,118 @@
+import { expect, test } from "@playwright/test";
+
+const PROJECT_ID = "00000000-0000-4000-8000-000000000151";
+
+function makeTask(index: number) {
+  return {
+    id: `scroll-task-${index}`,
+    projectId: PROJECT_ID,
+    title: `Horizontal wheel target ${index}`,
+    description: "Horizontal gestures should reach the board",
+    autoMode: true,
+    isFix: false,
+    status: "backlog",
+    priority: 1,
+    position: index * 1000,
+    blockedReason: null,
+    blockedFromStatus: null,
+    retryAfter: null,
+    retryCount: 0,
+    tokenInput: 0,
+    tokenOutput: 0,
+    tokenTotal: 0,
+    costUsd: 0,
+    roadmapAlias: null,
+    tags: [],
+    reworkRequested: false,
+    reviewIterationCount: 0,
+    maxReviewIterations: 3,
+    manualReviewRequired: false,
+    paused: false,
+    lastSyncedAt: null,
+    runtimeProfileId: null,
+    modelOverride: null,
+    runtimeLimitSnapshot: null,
+    runtimeLimitUpdatedAt: null,
+    scheduledAt: null,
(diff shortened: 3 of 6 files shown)


ichinya:
Request changes

The PR has the right scope and the core approach addresses #148: every column gets a bounded vertical list, while done/verified are ordered by updatedAt descending.

I found one user-visible regression that should be fixed before merge:

1. `overscroll-contain` is applied to the inner vertical lists on both   axes. The board itself is the horizontal scroll container, so horizontal   trackpad/wheel gestures started over a card list no longer chain to the   board. I reproduced this in Chromium: with two-axis containment the   parent scrollLeft remained 0; with Y-only containment it moved as   expected. Please remove this class or change it to   `overscroll-y-contain`, and add a Playwright regression for horizontal   scrolling over a card list.

There are also two correctness/test robustness issues:

2. The terminal-task comparator falls back to position whenever either   timestamp is invalid. Mixing date-based and position-based pairwise   comparisons makes the comparator non-transitive, so Array.sort can   produce input-order-dependent results. Parse both timestamps first,   consistently place malformed timestamps after valid ones, and then use   position/id as tie-breakers.

3. The ordering tests use `textContent.indexOf()`. When the expected first    card is missing, indexOf returns -1 and the assertion can still pass if    the second card is present. Please resolve both elements with getByText()    and compare their DOM positions.

Non-blocking: the `[FIX:148]` console.debug is a side effect inside useMemoand looks like temporary instrumentation; it should be removed before merge.

Build, lint, tests, security audit, and MCP Unit are green. The failing AI Review workflow is failing in its webhook-dispatch step rather than in the application validation jobs.

lee-to:
@ichinya Thank you for the detailed review. I addressed all of the points in db2cf34:

- changed the card lists to overscroll-y-contain and added a database-independent Playwright regression that uses native Chromium wheel input over a genuinely scrollable card list;
- replaced the terminal-task comparator with a transitive total order: valid timestamps first, updatedAt descending, then position and ID tie-breakers;
- replaced textContent.indexOf assertions with getByText element resolution and DOM-position comparisons;
- removed the temporary [FIX:148] console.debug side effect from useMemo.

Validation: npm run ai:validate passed, including format, lint, unit tests, coverage, build, all 4 Playwright tests, k6 load checks, and the Codex protocol check.

ichinya:
I reviewed the delta from `94622cb` to `db2cf34`. All findings from the previous review have been addressed, and I found no new merge-blocking issues.

### Response pass

- ✔ **Horizontal scroll chaining:** `overscroll-contain` was replaced with  `overscroll-y-contain`, preserving vertical boundary containment while  allowing horizontal wheel/trackpad gestures to reach the Kanban board.

- ✔ **Browser regression coverage:** the new Playwright spec uses native  Chromium wheel input over a genuinely vertically scrollable card list.  It verifies both horizontal board overflow and vertical list overflow  before asserting that the board's `scrollLeft` increases.

- ✔ **Terminal comparator:** terminal ordering now defines a deterministic  total order: valid timestamps first, `updatedAt` descending, then  position and ID tie-breakers. Malformed timestamps can no longer produce  pairwise comparison cycles.

- ✔ **Ordering assertions:** the tests now resolve both task elements with  `getByText()` and compare their DOM positions, so a missing first element  cannot accidentally satisfy the assertion.

- ✔ **Render purity:** the temporary `[FIX:148]` `console.debug` side effect   was removed from the `useMemo` path.

### Should fix — non-blocking

The new Playwright regression is not currently part of the standard PR quality gate. The GitHub `Tests` job runs `npm test` and coverage, while the browser suite remains behind the separate `perf` script.

In addition, `packages/web/tsconfig.json` and the lint script cover only `src/**`, so `e2e/**` is neither type-checked nor linted. The new fixtures already drift from the declared contracts: the mocked project omits required budget/queue fields, and `runtimeDefaults.app` is `{}` even though the response type requires the default and resolved profile IDs.

I recommend adding:

1. `tsconfig.e2e.json` with `e2e/**/*.ts` and `playwright.config.ts`;
2. explicit `Project`, `TaskListItem`, and `SettingsResponse` fixture types;
3. a fast PR job for `kanban-horizontal-scroll.spec.ts`, separate from the   performance-budget specs.

This does not block the current merge because the implementation is correct and the full local validation was reported as passing, but the regression test should eventually become an enforced CI check.

### Nits

- Update the PR description from `3 passed` to `4 passed` for the  Playwright suite and mention the new automated horizontal-scroll test.
- The README now calls this the Browser E2E suite, while scripts/config  comments still use `perf`; either naming is fine, but it would be useful  to document that the naming is intentionally retained or split smoke  E2E from performance tests later.

### Context gates

- Architecture: PASS
- Project rules: PASS
- Web checklist: PASS
- Docs sync: PASS, with the stale PR validation count noted above
- Roadmap: PASS
- PR size: PASS — 318 changed lines across 6 files
- Risk gating: PASS — localized client-side presentation fix with no
  persistence or API-contract changes; a feature flag would be
  disproportionate

The previously reported regressions are resolved. LGTM now.

ichinya (approved):

Issue #164, opened 2026-07-27
[bug] Stale watchdog quarantines tasks that were only queued, not hung


chamav:
## Summary

`listStaleInProgressTasks()` selects in-progress tasks that have **no active lock**, and `recoverStaleInProgressTasks()` then blocks anything whose `lastHeartbeatAt`/`updatedAt` is older than `AGENT_STAGE_STALE_TIMEOUT_MS`. But an unlocked in-progress task is not necessarily hung — it is also what a task queued behind a busy slot looks like:

- a project with `parallelEnabled=false` whose single slot is taken by another stage,
- a project at `COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT`,
- any task waiting while the global `COORDINATOR_MAX_CONCURRENT_TASKS` ceiling is saturated.

Such a task sits unlocked with a frozen `updatedAt` for as long as the queue ahead of it runs. Once that exceeds the stale timeout, the watchdog moves it to `blocked_external`, and after `AGENT_STAGE_STALE_MAX_RETRY` rounds it is quarantined with "auto-retry limit reached" — even though nothing ever went wrong with it.

## Reproduction

Observed on a live instance right after the coordinator poll-cycle fix (before that, the watchdog never ran at all, which is why this stayed hidden):

```
status              = blocked_external
blocked_reason      = Watchdog: task stale in review for 96m; auto-recover scheduled
blocked_from_status = review
retry_count         = 1
```

The task was in `review` in a serial project (`parallelEnabled=false`) whose only slot was held by an `implementing` task. It was never claimed, never started, never hung — it simply waited its turn for 96 minutes with `AGENT_STAGE_STALE_TIMEOUT_MS=5400000` (90 min).

The exposure scales with stage duration: with `AGENT_STAGE_RUN_TIMEOUT_MS` allowed to be hours, any queue behind a long implementer stage will routinely cross a 90-minute stale threshold.

## Why the current signal cannot distinguish the two

Both cases look identical in the schema — unlocked, in-progress, stale timestamp. `lockedBy` is cleared on graceful shutdown and on claim release, so "had a lock once" is not recoverable after the fact either.

## Possible fixes

1. **Record when a stage actually started** (e.g. `stage_started_at`, set at claim time, cleared on stage exit) and let the watchdog consider only tasks that have one. A queued task never gets the field, so it can never look stale.
[...]

PR #165, opened 2026-07-27
fix(agent): stop the poll cycle from stalling behind long stages

### LINKED ISSUE #163: [bug] Coordinator stops claiming work until the longest stage finishes
## Summary

A poll cycle only settles once every stage it started has finished, and `pollAndProcess()` serialises every trigger on that promise. Because an implementer stage legitimately runs for hours, the coordinator stops claiming work entirely for that whole time — free stage slots stay empty, and housekeeping never re-runs.

## Observed behaviour

Live instance on `main` @ 5060210, `COORDINATOR_MAX_CONCURRENT_TASKS=6`, `POLL_INTERVAL_MS=30000`, wake channel enabled. Two tasks were genuinely executing (heartbeats advancing every ~2.5 min). Everything else sat idle with **2 of 6 slots used** while ~28 poll ticks fired:

| task | project | status | lock | idle |
|---|---|---|---|---|
| A | CacheDeckIos | `implementing` (rework bounced back) | none | 14 min |
| B, C, D, E | MQTT Debugger | `planning` | none | 30 min |

Task A is the clearest case: its project had **no other active task**, four global slots were free, and it was still never claimed.

Neither project was in forced-serial mode (`parallelEnabled=1`, no branch-bound tasks, git isolation off), so `COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT=3` applied — three planners could have run alongside the implementer.

Throughput on this instance dropped from 6–9 tasks/day to 2–3/day over the week.

## Root cause

1. `runPollCycle()` awaits every stage it started — `await Promise.allSettled(spawned)` per stage and `await Promise.allSettled(laneResults)` per cycle. Cycle duration == duration of the longest stage in it.
2. `pollAndProcess()` returns the active cycle promise and only sets `followUpPollRequested`, so both the cron tick and every WebSocket wake are absorbed by the running cycle.
3. Housekeeping — `releaseStaleTaskClaims()`, `recoverStaleInProgressTasks()`, `processDueScheduledTasks()`, `processAutoQueueAdvance()` — lives at the top of `runPollCycle`. The stale-task watchdog, the mechanism meant to recover a wedged coordinator, is itself blocked by the wedge.
4. A task landing in an already-passed stage (a `planning` task created after the lane moved on, or a rework bounced back to `implementing`) cannot be picked up until the whole cycle ends.

The coalescing was introduced deliberately in 0daf2b2 to preserve project-local stage order across trigger sources — a planner and an implementer of one project must not share a working tree. That guarantee is worth keeping; it just should not require freezing every *other* project too.

## Fix

Allow a bounded number of overlapping cycles while global stage slots are free, and have an overlapping cycle skip every project that already holds a stage:

- `stageSemaphore` remains the real concurrency ceiling, so overlap can never exceed `COORDINATOR_MAX_CONCURRENT_TASKS`.
- `stageSemaphore.hasActiveProject(projectId)` filters busy projects out of an overlapping cycle → project-local stage order is preserved exactly as before.
- When the ceiling is saturated, triggers still coalesce into one follow-up cycle (old behaviour).
- Claims are atomic and lanes revalidate candidates after acquiring a permit, so overlapping cycles cannot double-claim.

Verified on the same instance: after the change the coordinator immediately filled 5 of 6 slots — including the 14-minute-stuck rework task and three planners that had waited 30 minutes. Remaining waits were all explained by configured limits (per-project cap, serial project), not by the barrier.

PR follows with tests covering all three paths: overlapping cycle for an idle project, skipping a busy project, and coalescing at a saturated ceiling.

## Environment

- `main` @ 5060210, Node 22, macOS
- `COORDINATOR_MAX_CONCURRENT_TASKS=6`, `POLL_INTERVAL_MS=30000`, `AGENT_WAKE_ENABLED=true`

chamav:
Closes #163.

## Problem

A poll cycle only settles once every stage it started has finished, and both the coordinator and the scheduler above it serialised on that promise. An implementer stage legitimately runs for hours, so during that time the coordinator claimed nothing at all: free stage slots stayed empty and housekeeping (stale-lock release, stale-task watchdog, scheduled tasks, auto-queue) never re-ran, because all of it lives at the top of a cycle.

Measured on a live instance (`COORDINATOR_MAX_CONCURRENT_TASKS=6`, 30s poll, wake channel on): 2 of 6 slots in use, one task sat in `implementing` with no lock for 14 minutes **in a project with no other active task**, and four `planning` tasks waited 30 minutes — across ~28 poll ticks that had capacity and did nothing.

There are two independent barriers, and removing only the first is not enough.

## Change 1 — bounded overlapping cycles

Allow a bounded number of overlapping poll cycles while global stage slots are free:

- `canStartOverlappingPollCycle()` gates overlap on `stageSemaphore.totalActive() < COORDINATOR_MAX_CONCURRENT_TASKS` and a cap of 8 in-flight cycles, so the real concurrency ceiling is unchanged.
- An overlapping cycle skips every project that already holds a stage (`stageSemaphore.hasActiveProject()`), so **project-local stage order from #149 / 0daf2b2 is preserved** — a planner still never shares a working tree with an implementer of the same project.
- When the ceiling is saturated, triggers coalesce into one follow-up cycle exactly as before.

Claims are atomic and lanes already revalidate candidates after acquiring a permit, so overlapping cycles cannot double-claim a task.

## Change 2 — the interval tick must not wait for the cycle

`startPollScheduler` guards against re-entrancy by awaiting its callback. Because `pollAndProcess()` settles only after every stage the cycle started has finished, `isRunning` stayed true for hours and every tick in between was dropped *before* `pollAndProcess()` was called — so the bounded-overlap logic above never ran from the cron path at all.

That reduced the coordinator to a wake-only system: work appearing while a stage runs waits for an unrelated external event rather than the next tick. Concretely, a task that auto-queue advances into an already-passed stage, or a status change written by an out-of-process caller whose WS broadcast does not land, is invisible until something else happens to fire a wake.

`startPollScheduler` gains an explicit `awaitCallback` option (default `true`, existing behaviour and its test unchanged). The coordinator opts out — concurrency is already bounded inside `pollAndProcess()`.

## Tests

`packages/agent/src/__tests__/coordinator.test.ts`:

- `should run an overlapping poll cycle for a project with no active stage`
- `should keep project-local stage order by skipping busy projects in an overlapping cycle`
- `should coalesce a follow-up cycle when the global ceiling is saturated`

`packages/agent/src/__tests__/pollScheduler.test.ts`:

- `keeps ticking while a previous callback is pending when awaitCallback is false`

The pre-existing `should drain started lane tasks before propagating a later candidate setup failure` was updated: its assertion that the second trigger returns the *same* promise encoded the old contract; the behavioural assertion (no reviewer for a busy project while a stage drains) is unchanged and still passes.

Full suite: `npm run lint`, `npm test` (379 agent tests, 10/10 turbo tasks), `npm run build`, `npm run format:check` all green. Agent coverage 88.2% statements.

`npm run ai:load` / `ai:perf` were not run here — they contend with the live dev stack on port 3009 on this machine.

## Verification

After change 1 the same instance immediately filled 5 of 6 slots, including the 14-minute-stuck rework task and three planners that had waited 30 minutes.

Change 2 was verified separately against the running instance by inserting a task straight into SQLite — emitting no WebSocket event whatsoever, the exact case that used to hang forever:

```
08:04:42  inserted (status=planning, via SQL, no WS event)
08:05:09  claimed by the coordinator   ← ~27s, i.e. the POLL_INTERVAL_MS=30000 tick
```

Before this change that task would have waited indefinitely.

## Follow-ups spotted while verifying

- With the coordinator unblocked, the stale watchdog started running again for the first time — and immediately quarantined a task that had merely been queued behind a busy serial project. Filed separately as #164.
- The internal broadcast endpoints reject the agent's and the MCP server's calls with `401` under a default `npm run dev` setup, which is why the wake path could not compensate for the dead tick. Filed and fixed separately; it is orthogonal to this PR.

## Docs

`docs/architecture.md`, `docs/configuration.md` and `packages/agent/CHECKLIST.md` updated — each documented the old "one coalesced follow-up cycle" contract, and none described the scheduler-level barrier.

https://claude.ai/code/session_01CwzHqmF1TKfMhmRqjQ3o5G


lee-to:
## Code review

This fixes a real coordinator-starvation failure and the focused tests, documentation updates, and green CI give the change a solid base. However, the new coordinator hot-path behavior is enabled unconditionally, and the project scan can still hide an idle project under supported concurrency settings. **Verdict: REQUEST_CHANGES.**

### Must fix

1. **Gate the overlapping-cycle rollout off by default** — `packages/agent/src/coordinator.ts:1015` and `packages/agent/src/index.ts:61` — this PR unconditionally replaces the coordinator single-flight algorithm and changes interval dispatch on a request/polling hot path. Add `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED` to `packages/shared/src/env.ts` as a boolean defaulting to `false`, document it in `.env.example` and `docs/configuration.md`, and gate both overlapping `pollAndProcess()` behavior and `{ awaitCallback: false }` behind the single cached flag read. With the flag off, the exact existing single-flight behavior must remain reachable; add tests for both off and on paths.

2. **Do not cap idle-project discovery at `3 × maxProjectLanes`** — `packages/agent/src/coordinator.ts:1086` — the supported settings allow `COORDINATOR_MAX_CONCURRENT_PROJECTS=1` and `COORDINATOR_MAX_CONCURRENT_TASKS=4` (or much higher). If three active projects also have queued actionable work, they can occupy all three rows returned by `laneWindow`; the filter removes them and never sees a fourth idle project, so a free global slot remains unused until a long stage finishes. This contradicts the updated agent checklist guarantee that no cycle waits on a running stage before picking up an idle project. Exclude active project IDs in the data query, page until enough idle IDs are found, or size the query to at least `maxProjectLanes + distinctActiveProjectCount`; add a regression test with a low lane limit and a higher global task limit.

### Should fix

1. **Make `awaitCallback: false` genuinely dispatch-only** — `packages/agent/src/pollScheduler.ts:44-46` — this branch still executes `await callback()`. The missing `isRunning` guard allows later ticks, but every tick leaves another pending wrapper attached to the same multi-hour poll promise. Dispatch with `void callback(); return;` (the coordinator callback already handles its own error), or rename the option to describe concurrent ticks and document the pending-promise behavior.

### Nits

- None.

### Context gates

- Architecture: pass — package boundaries remain intact; agent DB operations still go through `@aif/data`.
- Rules: pass — no clear `.ai-factory/RULES.md` violation beyond the separately reported risk/checklist gates; CI confirms lint, tests, build, and audit are green.
- Roadmap: warn — this `fix` closes #163 but does not link the work to the roadmap; the closest milestone is **Safe Parallel Agent Processing**.
- CHECKLIST compliance (touched packages: `agent`): error — the new poll-scheduler item is not met for supported configurations because the fixed lane window can hide an idle project. Runtime/subagent, state-machine-transition, first-activity-watchdog, and auto-queue-commit items are not changed; permit revalidation and lane draining remain intact.
- Docs sync: pass — `docs/architecture.md`, `docs/configuration.md`, and `packages/agent/CHECKLIST.md` were updated; no API, runtime-adapter, UI/Pencil, DB-migration, or Docker sync obligation was triggered. The required rollout flag documentation remains part of must-fix #1.
- PR size (296 changed lines / 8 files / 2 tightly related concerns): pass — below the soft threshold and cohesively scoped.
- Risk gating (feature flag): error — `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED=false` is missing.

### Positive notes

- The tests cover idle-project overlap, project-local ordering, saturated coalescing, and continued interval ticks.
- The PR updates both architecture/configuration docs and the package checklist, and all current GitHub checks pass.

lee-to:
@chamav ping

chamav:
Thanks for the review — all three points are addressed in c2f38e5, and 3576cf0 merges current `main` (it had added its own flags at the same append points, so the branch had started to conflict).

### Must fix 1 — rollout flag

`AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED` added to `packages/shared/src/env.ts` (boolean, default `false`, using main's new `booleanEnvSchema` helper), documented in `.env.example` and `docs/configuration.md`, and added to `turbo.json` passthrough.

Both behaviours are gated by one read site, `isOverlappingPollCyclesEnabled()`:

- `canStartOverlappingPollCycle()` returns false when the flag is off, so at most one cycle is ever in flight and `pollAndProcess()` is exactly the original single-flight guard — the active promise is returned and one follow-up cycle is coalesced.
- `index.ts` passes `awaitCallback: !enabled`, so with the flag off the scheduler awaits the callback again.

Tests cover both states: the new `should keep the single-flight poll loop while the rollout flag is off`, and the three overlap tests, which now enable the flag. `should drain started lane tasks before propagating a later candidate setup failure` is back to its original same-promise assertion, since that is the default rollout again.

One note on "single cached flag read": the read is a property access on the module-level `env` object returned by `getEnv()` (already validated and memoised) rather than a `const` captured at import time, so tests and any future config reload still observe the value. It is still one site, and `index.ts` reads it once at startup.

### Must fix 2 — idle-project discovery

The `3 × maxProjectLanes` window is gone. Busy projects are now excluded inside the query: `listCoordinatorActionableProjectIds(limit, { excludeProjectIds })` adds a `NOT IN` predicate to the `WHERE`, so `limit` is only ever spent on projects the cycle can actually use. `StageSemaphore.hasActiveProject()` is replaced by `activeProjectIds()`; the snapshot is exact, because reading it and running the query are synchronous with no await in between.

Regression test `should reach an idle project behind more busy projects than there are lanes` uses exactly your combination — `COORDINATOR_MAX_CONCURRENT_PROJECTS=1`, `COORDINATOR_MAX_CONCURRENT_TASKS=4`, three busy projects that still hold queued work so they stay in the actionable set, and one idle project ranked behind them. I verified it in both directions: on the old windowing it fails (`expected "vi.fn()" to be called with [ 'idle-lane-review-task', … ]`), with the query-level exclusion it passes. The `@aif/data` side is covered separately by `spends the limit on projects the caller can use when excluding ids`.

### Should fix — dispatch-only ticks

`awaitCallback: false` no longer awaits inside the tick: `setInterval` calls `void callback()` and returns, so no per-tick wrapper stays attached to the multi-hour cycle promise. The option's docs now state that such a callback owns its own error handling, since the scheduler cannot observe a rejection it no longer holds — the coordinator's callback already try/catches.

### Roadmap

Linking this to **Safe Parallel Agent Processing**, as suggested.

### Merge with `main`

Resolved in 3576cf0: both entries kept in `env.ts`, `turbo.json`, `docs/configuration.md` and the `makeEnv` literal in `hooks.test.ts`; in `listCoordinatorActionableProjectIds` both changes are taken — main's `executionOwner = "ai"` filter joins the condition list alongside the exclusion.

### Verification

Local run is green: `npm run format:check`, `npm run lint`, `npm run build`, and the full suite — agent 391, data 260, shared 235, api 472, web 726, mcp 103. Coverage: agent 88.3%, data 94.3%, shared 80.8% statements. `npm run ai:load` / `ai:perf` were again not run — they contend with the live dev stack on port 3009 on this machine.

The workflow runs for 3576cf0 are sitting in `action_required` and need an "Approve and run" before CI can report — could you approve them?

lee-to:
## Code review

Follow-up on the prior review: the new commits address the blocking coordinator rollout issues and keep the risky behavior behind an off-by-default flag. I reviewed the delta in `c2f38e5` and `3576cf0`, plus the resulting PR head. CI is green. **Verdict: APPROVE.**

### Response pass

- Addressed: rollout flag is now off by default in `packages/shared/src/env.ts:276`, documented in `.env.example:145` and `docs/configuration.md:63`, passed through `turbo.json:11`, and read through `isOverlappingPollCyclesEnabled()` in `packages/agent/src/coordinator.ts:1091`.
- Addressed: idle-project discovery now excludes active projects in the data query before applying the lane limit (`packages/agent/src/coordinator.ts:1169`, `packages/data/src/index.ts:2163`), with regression coverage for the hidden-idle-project case in `packages/agent/src/__tests__/coordinator.test.ts:2251` and the SQL exclusion path in `packages/data/src/__tests__/index.test.ts:1029`.
- Addressed: `awaitCallback: false` is now dispatch-only; the scheduler drops the callback promise instead of awaiting it (`packages/agent/src/pollScheduler.ts:57`), and the coordinator callback owns error handling (`packages/agent/src/index.ts:63`).
- Prior comment: https://github.com/lee-to/aif-handoff/pull/165#issuecomment-5101947822

### Must fix

None.

### Should fix

None.

### Nits

None.

### Context gates

- Architecture: pass — package boundaries remain intact; `agent` still reaches DB state through `@aif/data`, and the new data query stays inside the data package.
- Rules: pass — no string-based error classification, no direct DB access from `agent`, no runtime-adapter/Docker/schema sync obligation triggered.
- Roadmap: pass — the PR closes #163 and fits the completed **Safe Parallel Agent Processing** milestone.
- CHECKLIST compliance (touched packages: `agent`, `data`, `shared`): pass — agent poll-scheduler flag states, idle-project query exclusion, dispatch-only ticks, data query coverage, and shared env/turbo coverage are present.
- Docs sync: pass — `.env.example`, `docs/configuration.md`, `docs/architecture.md`, and `packages/agent/CHECKLIST.md` now describe the rollout behavior and scheduler contract.
- PR size (696 changed lines / 15 files / 2 related concerns): pass — above the soft line guideline, but the change is cohesive and much of the size is targeted regression coverage/docs for the same coordinator fix.
- Risk gating (feature flag): pass — `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED=false` gates both overlapping poll cycles and dispatch-only scheduler ticks; the off path preserves the single-flight loop.

LGTM now.

lee-to:
## Code review

Status re-check after the prior follow-up review: there are no new commits after https://github.com/lee-to/aif-handoff/pull/165#issuecomment-5652243089, so I re-checked the current PR head and CI state rather than re-reviewing a new delta. The head remains `3576cf0cbadbfd67f6487b9ecd9f7c7cf6666938`, CI is green, and the previously reviewed fixes are still present. **Verdict: APPROVE.**

### Response pass

- Addressed: `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED` is still off by default in `packages/shared/src/env.ts`, documented in `.env.example` and `docs/configuration.md`, and passed through `turbo.json`.
- Addressed: overlapping cycles still exclude active projects inside `listCoordinatorActionableProjectIds(...)` before applying the limit, with data-layer regression coverage.
- Addressed: `awaitCallback: false` remains dispatch-only in `packages/agent/src/pollScheduler.ts`, and the coordinator callback owns its error handling.

### Must fix

None.

### Should fix

None.

### Context gates

- Architecture: pass - package boundaries remain intact; agent access to task/project state still goes through `@aif/data`.
- Rules: pass - no string-based error classification or direct DB access violation in the changed code; the rollout flag is off by default.
- Roadmap: pass - the PR closes #163 and fits the Safe Parallel Agent Processing work.
- CHECKLIST compliance (touched packages: `agent`, `data`, `shared`): pass - the relevant package checklist concerns have targeted tests or docs.
- Docs sync: pass - `.env.example`, `docs/configuration.md`, `docs/architecture.md`, and `packages/agent/CHECKLIST.md` remain in sync with the behavior.
- PR size (696 changed lines / 15 files / 2 related concerns): pass - above the soft line guideline, but cohesive and largely regression coverage/docs.
- Risk gating (feature flag): pass - `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED=false` preserves the legacy single-flight path until explicitly enabled.

LGTM still.

lee-to:
@chamav pls resolve merge conflicts

chamav:
Merge conflicts resolved in f3bb966 — the branch merges `main` cleanly again.

The conflict was purely an append-point collision: `main` added `AIF_GITHUB_PROJECT_CLONE_ENABLED` at the same three places this branch appends `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED`. Both flags are kept, in the same order, in:

- `packages/shared/src/env.ts`
- `docs/configuration.md`
- `packages/agent/src/__tests__/hooks.test.ts` (env fixture)

`.env.example` and `turbo.json` auto-merged and carry both flags.

No behavior changed — the approved diff is untouched. `AIF_AGENT_OVERLAPPING_POLL_CYCLES_ENABLED` is still off by default, idle-project discovery still excludes active projects in the data query, and `awaitCallback: false` is still dispatch-only.

Verified locally on the merge result: `turbo lint`, `turbo build` and `turbo test` all pass (14/14 build+lint tasks, 10/10 test tasks).

PR #167, opened 2026-07-28
fix(mcp): authorize internal broadcasts so WS events are not dropped

### LINKED ISSUE #166: Internal broadcasts are rejected with 401 by default: no live UI updates, coordinator wake channel never fires
## Summary

`POST /tasks/:id/broadcast` and `POST /projects/:id/broadcast` are the only way the agent coordinator and the MCP server can emit WebSocket events — both write to SQLite through `@aif/data` and then ask the API to broadcast. Both endpoints are guarded by `internalBroadcastAuth`, which trusts a caller in exactly two ways:

1. the configured `INTERNAL_BROADCAST_TOKEN`, or
2. a loopback caller header (`X-Forwarded-For` / `X-Real-IP`) while `NODE_ENV=development`.

Two separate problems make those calls fail:

**1. MCP never sends any credential.** `packages/mcp/src/utils/broadcast.ts` sends only `{ "Content-Type": "application/json" }`. The agent's notifier has the correct logic (`internalBroadcastHeaders()` in `packages/agent/src/notifier.ts`), but MCP's copy of the same call was never updated when the guard was introduced. So every MCP-driven task change is rejected with `401`, regardless of configuration.

**2. The default local setup closes both accepted paths.** `npm run dev` (`scripts/dev.mjs` → turbo) does not set `NODE_ENV`, and `.env.example` presents `INTERNAL_BROADCAST_TOKEN` as an optional production knob:

```
# Optional internal token used to authorize /projects/:id/broadcast calls.
# INTERNAL_BROADCAST_TOKEN=change-me
```

With neither set, `resolveBroadcastAuthDecision` falls through to `{ trusted: false, mode: "rejected" }`. So on a stock local install **every** out-of-process broadcast — agent included — is answered with `401`.

## Impact

- The Kanban board does not live-update on any agent-driven transition. Only mutations the web app performs itself refresh, because those broadcast in-process.
- The agent's wake channel never fires on agent-side transitions, since the wake channel is a WS subscriber and the event never goes out.

The second point is the damaging one. The coordinator is documented as a dual-trigger system (periodic tick + WS wake). With broadcasts silently rejected, the wake trigger only fires for changes made through the web UI. Combined with #163 — where the interval tick was also suppressed while any stage ran — the coordinator had no live trigger at all, and tasks sat in an actionable status indefinitely with free stage slots. #163 fixes the tick; this issue is why the wake path could not compensate.

## Reproduction

Stock local dev (`npm run dev`, no `NODE_ENV`, no token). The exact request MCP sends:

```
$ curl -s -o /dev/null -w "%{http_code}\n" -X POST \
    http://localhost:3009/tasks/<id>/broadcast \
    -H "Content-Type: application/json" -d '{"type":"task:updated"}'
401
```

The agent's development fallback (`X-Real-IP: 127.0.0.1`) is also rejected, because `NODE_ENV` is empty rather than `development`:

```
$ curl -s -o /dev/null -w "%{http_code}\n" -X POST \
    http://localhost:3009/tasks/<id>/broadcast \
    -H "Content-Type: application/json" -H "X-Real-IP: 127.0.0.1" \
    -d '{"type":"task:updated"}'
401
```

After setting `INTERNAL_BROADCAST_TOKEN` in `.env` and restarting the stack, the same call with the token returns `200` and the board updates live again.

## Fix

PR follows:

- extract the header logic into `@aif/shared` as `internalBroadcastHeaders()` so the agent and MCP share one implementation instead of two divergent copies (the token is passed in rather than read inside, so each caller resolves it through its own `getEnv()` binding);
- use it in `broadcastTaskChange`;
- set `INTERNAL_BROADCAST_TOKEN` by default in `.env.example` and document in `docs/configuration.md` and both package checklists that leaving it unset locally silently disables live updates and the wake channel.

Setting `NODE_ENV=development` in `scripts/dev.mjs` would also close the second half, but an explicit token is robust regardless of `NODE_ENV` and behind a proxy. Happy to do it that way instead if maintainers prefer.

## Environment

- `main` @ 5060210, Node 22, macOS
- `npm run dev`, no `NODE_ENV`, no `INTERNAL_BROADCAST_TOKEN`

chamav:
Closes #166.

## Problem

`POST /tasks/:id/broadcast` and `POST /projects/:id/broadcast` are how the agent coordinator and the MCP server emit WebSocket events — both write to SQLite through `@aif/data` and then ask the API to broadcast. `internalBroadcastAuth` trusts a caller in exactly two ways: the configured `INTERNAL_BROADCAST_TOKEN`, or a loopback caller header while `NODE_ENV=development`.

**MCP sends neither.** `packages/mcp/src/utils/broadcast.ts` sets only `{ "Content-Type": "application/json" }`, so every MCP-driven task change is rejected with `401` regardless of configuration. The agent's notifier already had the correct logic; MCP's copy of the same call was never updated when the guard was introduced.

**The default local setup closes both paths.** `npm run dev` does not set `NODE_ENV`, and `.env.example` presented the token as an optional production knob — so on a stock install *every* out-of-process broadcast, agent included, is rejected.

Consequence: the board does not live-update on agent-driven transitions, and the coordinator's wake channel never fires for them. Combined with #163 (where the interval tick was suppressed while any stage ran) the coordinator had no live trigger at all.

## Change

- `internalBroadcastHeaders()` moves to `@aif/shared`, so the agent and MCP share one implementation instead of two divergent copies. The token is passed in rather than read inside, so each caller resolves it through its own `getEnv()` binding — which also keeps the existing package-local env mocks working in tests.
- `broadcastTaskChange` uses it.
- `.env.example` sets `INTERNAL_BROADCAST_TOKEN` by default and explains that leaving it unset locally silently disables live updates and the wake channel. Same note added to `docs/configuration.md` (the variable was previously undocumented there) and to the `@aif/agent` / `@aif/mcp` checklists.

No behaviour change for a correctly configured deployment: with a token set, the headers are byte-for-byte what the agent notifier already sent.

Setting `NODE_ENV=development` in `scripts/dev.mjs` would also close the second half. I went with the explicit token because it is robust regardless of `NODE_ENV` and behind a proxy — happy to switch if you prefer.

## Tests

- `packages/shared/src/__tests__/internalBroadcast.test.ts` (new) — token in both accepted header forms, whitespace trimming, the `NODE_ENV=development` loopback fallback, and the misconfigured case (no token outside development) asserting the helper does not paper over it.
- `packages/mcp/src/__tests__/broadcast.test.ts` — new `broadcastTaskChange internal auth` case asserting the headers reach `fetch`.

Full suite: `npm run lint`, `npm test` (10/10 turbo tasks), `npm run build`, `npm run format:check` all green.

`npm run ai:load` / `ai:perf` were not run here — they contend with the live dev stack on port 3009 on this machine.

## Verification

Against a running instance, the exact request MCP sends:

```
before: 401
after (token configured, same request + headers): 200
```

With the token in place the board resumed live-updating and agent-side transitions started firing the wake channel again — a task approved in the UI was claimed immediately instead of waiting for the next tick.

https://claude.ai/code/session_01CwzHqmF1TKfMhmRqjQ3o5G


lee-to:
## Code review

This PR correctly centralizes the internal-broadcast request headers and wires the MCP mutation path into the existing auth contract. The helper and caller tests are focused, package boundaries remain intact, and all CI checks are green. One deployment-security issue is blocking approval, though: the example environment now installs a publicly known credential into every copy-based setup, including the documented production Docker flow.

### Must fix

1. **Do not ship a live, predictable internal credential** — `.env.example:222` — `INTERNAL_BROADCAST_TOKEN=change-me` is copied verbatim by the README/getting-started setup and injected by `docker-compose.production.yml`. The production web proxy exposes API routes under `/api/`, so a deployment that follows the documented copy step without replacing this value has a known token for the trusted broadcast endpoints; an external caller can forge task/project broadcasts and repeatedly trigger coordinator wake work. Keep the token commented and make local development use the existing development fallback, or generate/require a strong per-install value. If an active placeholder remains, production startup must reject it and the setup docs must require rotation before launch.

### Should fix

1. **Remove the second, stale MCP broadcast implementation** — `packages/mcp/src/notifier.ts:21` (unauthenticated headers at line 31) — `broadcastTaskEvent()` has no production import and is exercised only by its own test, but it duplicates `broadcastTaskChange()` and still sends only `Content-Type`. That contradicts the new `packages/mcp/CHECKLIST.md:9` rule and leaves the same regression trap in the tree. Delete `notifier.ts` plus its isolated test, or make it delegate to the shared authenticated path.

### Nits

None.

### Context gates

- Architecture: pass — shared remains a foundation utility and agent/MCP consume it through the public `@aif/shared` export.
- Rules: warn — the active path is DRY, but the stale MCP notifier remains duplicated dead code.
- Roadmap: warn — the PR links issue #166 but does not identify a roadmap milestone.
- CHECKLIST compliance (touched packages: `agent`, `mcp`, `shared`): warn — applicable build/test and boundary checks pass; the stale MCP notifier does not satisfy the new authenticated-broadcast item.
- Docs sync: pass — the environment variable and operational failure mode are documented; the placeholder-value security issue above still needs correction.
- PR size (157 changed lines / 10 files / 2 concerns): pass — focused and comfortably below the soft limit.
- Risk gating (feature flag): n/a — this restores the existing internal broadcast contract and adds no new API shape, transport, external call, or persistence behavior.

### Positive notes

- `internalBroadcastHeaders()` preserves package-local `getEnv()` mocking while removing the live agent/MCP divergence.
- Tests cover token trimming, both accepted token headers, the development fallback, and the MCP fetch boundary.
- Build, ESLint, MCP Unit, Tests, npm audit, and the MCP status check all pass.

**Verdict: REQUEST_CHANGES** — blocked on replacing the known default credential; the stale duplicate should be cleaned up in the same pass.

TheCoffeeToDev:
I took a look at this after the recent changes on `main` and put together an updated version based on the current branch.

It addresses the review feedback by keeping the example token commented, generating an ephemeral token for local `npm run dev`, and removing the stale MCP notifier implementation/test.

Compare:
https://github.com/lee-to/aif-handoff/compare/main...TheCoffeeToDev:aif-handoff:fix/internal-broadcast-auth-final

Commit: `7724e7c`

Happy to open this as a separate PR against current `main` if that would be easier to review/merge.

lee-to:
@chamav ping

chamav:
Rebased onto current `main` and addressed both review points. Two commits: `1835f6f` merges `main`, `d5f5b40` carries the fixes.

### Must fix 1 — live placeholder credential

`.env.example:239` keeps `INTERNAL_BROADCAST_TOKEN` commented out, so a copy-based setup installs no known token.

Local development no longer needs one: `scripts/dev.mjs:108` mints a random 32-byte per-run token when the variable is unset, and the dev processes inherit it through the existing `spawnDev({ env: process.env })`. Nothing is written to disk, and the value differs every run.

One honest limitation, now stated in the docs: only processes this script spawns inherit it. An MCP server started by an MCP client outside `npm run dev` still needs an explicit shared token.

### Should fix 1 — stale second MCP broadcast implementation

`packages/mcp/src/notifier.ts` and its test are deleted. `broadcastTaskEvent()` had no production importer and sent unauthenticated headers.

### While merging: `main` grew its own copy of this helper

`main` added `internalApiHeaders()` inside `packages/agent/src/notifier.ts`, which is behaviourally identical to the `internalBroadcastHeaders()` this PR extracts into `@aif/shared`. Rather than ship both, the shared helper wins and main's two call sites in `packages/agent/src/githubWorkflow.ts:52,114` now use it. `packages/mcp/src/utils/broadcast.ts` also drops main's partial inline headers (token only, no `Authorization`) along with the then-dead `internalToken` const.

Happy to split that fold-in into its own PR if you would rather review it separately.

### Also: the loopback fallback no longer exists

While resolving `.env.example` I noticed `main` rewrote `internalBroadcastAuth` — it now accepts a matching token, or any caller while `NODE_ENV=test`, and there is no `X-Real-IP` loopback path anymore.

That made the helper's `X-Real-IP: 127.0.0.1` branch a dead header on every tokenless call, and its doc comment, the `docs/configuration.md` row and one unit test all described a contract the server had stopped honouring. All four now follow the real one. This is also why the ephemeral dev token matters: without it, `npm run dev` gets a 401 on every broadcast, which is the bug this PR set out to fix.

### Verification

`turbo lint`, `turbo build` and `turbo test` all pass locally on the merge result (14/14 build+lint tasks, 10/10 test tasks).

Note for CI: `packages/agent/src/__tests__/subagentQuery.test.ts > forces new session and skips session persistence for isolated codex subagent workflows` flaked three times for me under parallel load and passes in isolation — it runs ~7s, so it looks like a timeout rather than anything this branch touches. Unrelated to these changes, but it may show up in a run.

lee-to:
## Code review

Follow-up pass on the commits after my July 28 review: the internal broadcast fix is now in the right shape. The two prior findings are addressed, the active agent/MCP call sites share `internalBroadcastHeaders()`, and the docs now describe the real `internalBroadcastAuth` contract after the loopback fallback was removed.

### Response pass

- Addressed: ✔ **Live placeholder credential** — prior comment: https://github.com/lee-to/aif-handoff/pull/167#issuecomment-5101923383. `.env.example:239` is commented again, and `scripts/dev.mjs:108` mints a per-run token for the `npm run dev` process tree when none is configured.
- Addressed: ✔ **Stale MCP notifier implementation** — `packages/mcp/src/notifier.ts` and `packages/mcp/src/__tests__/notifier.test.ts` are deleted, so the unauthenticated duplicate path is gone.
- New: ⚠ **CI status** — `npm audit` is currently failing with high/moderate dependency advisories (`fast-uri`, `hono`, `qs` via `@modelcontextprotocol/sdk`/`express`, `smol-toml`). This PR does not change dependency manifests, so I am treating that as repository dependency state rather than a regression from this patch.

### Must fix

None.

### Should fix

None for this PR. The red `npm audit` check still needs a dependency-refresh follow-up before merge policy can consider the branch fully green.

### Nits

None.

### Context gates

- Architecture: pass — shared remains a foundation utility and agent/MCP consume it through the public `@aif/shared` export; no DB-boundary violations found.
- Rules: pass — the duplicate broadcast implementation was removed and the remaining call sites use the shared helper.
- Roadmap: warn — the PR links issue #166, but no roadmap milestone is referenced.
- CHECKLIST compliance (touched packages: `agent`, `mcp`, `shared`): pass — the new checklist items match the implementation, and touched code keeps broadcast auth, package boundaries, and tests in place.
- Docs sync: pass — `.env.example` and `docs/configuration.md` document the token behavior and the external-MCP limitation.
- PR size (274 changed lines / 14 files / 1 concern): pass — focused and below the soft limit.
- Risk gating (feature flag): n/a — this restores an existing internal auth contract and dev setup path; it does not introduce a new external behavior that needs a rollout flag.

**Verdict: COMMENT** — code review LGTM now, with the merge caveat that the current `npm audit` check is red for dependency advisories outside this PR’s changed files.

PR #169, opened 2026-07-31
feat: add Participants Mode and human/AI task ownership

### LINKED ISSUE #159: [feature] Add Participants Mode with local auth and human/AI task ownership
## Summary

Add an optional multi-user **Participants Mode** with lightweight local username/password authentication and explicit human/AI task ownership.

The goal is to let one AIF Handoff instance coordinate a mixed team:

- some tasks are assigned to AI and continue through the automated workflow;
- some tasks are assigned to human participants and are completed manually;
- everyone can see ownership, progress, comments, and handoffs on the same Kanban board.

The current anonymous single-user behavior should remain the default when Participants Mode is disabled.

## Problem

Today, AIF Handoff has no application-user model or task assignees:

- the web UI and most API routes assume a trusted single-user environment;
- tasks do not identify whether a human or AI is responsible for execution;
- `human` only identifies the author type of a comment, not a real participant;
- the coordinator and auto-queue treat actionable tasks as candidates for AI execution;
- there is no "My tasks" view or accountability/audit trail for individual people.

This makes it difficult to use Handoff as a shared delivery board where people and AI agents work side by side.

### 1. Optional Participants Mode

Add an instance-level setting such as `PARTICIPANTS_MODE_ENABLED`.

When disabled:

- preserve the current anonymous/single-user experience;
- do not require login;
- preserve existing task and API behavior.

When enabled:

- require authentication for the web UI, REST API, and WebSocket connections;
- show the current participant and provide login/logout flows;
- allow administrators to create, deactivate, and reset credentials for participants;
- do not enable public self-registration by default.

### 2. Lightweight Local Authentication

The initial scope can use instance-local accounts:

- username;
- display name;
- password;
- active/inactive status;
- role: `admin` or `member`.

Security requirements:

- never store or log plaintext passwords;
- use an established password-hashing algorithm with unique salts;
- use secure, expiring sessions in `HttpOnly`, `SameSite` cookies;
- protect state-changing requests against CSRF;
- rate-limit login attempts;
- invalidate sessions when an account is deactivated or credentials are reset;
- provide a documented, non-interactive way to bootstrap the first administrator;
- keep runtime-provider credentials completely separate from participant credentials.

External identity providers, SSO, OAuth, and complex organization management can remain out of scope for the first version.

### 3. Human/AI Task Ownership

Add explicit execution ownership to every task:

- `ai` — handled by the existing coordinator and configured runtime/workflow;
- `human` — handled manually by one or more assigned participants.

The task create/edit UI should allow users to:

- choose Human or AI ownership;
- assign active participants to human-owned tasks;
- reassign a task between participants;
- hand a task from a human to AI or from AI to a human at any supported point during its lifecycle;
- clearly see ownership and assignees on task cards and in task details.

Existing tasks should migrate to `ai` ownership to preserve current behavior.

Ownership is not fixed for the lifetime of a task. A typical task may move through several handoffs, for example:

```text
AI plans -> Alice implements -> AI reviews -> Bob addresses feedback -> AI verifies
```

Each handoff must preserve all existing task context and make the next executor's responsibility explicit.

### 4. Coordinator and State-Machine Rules

- Human-owned tasks must never be claimed by the AI coordinator, scheduler, or project auto-queue.
- Human-owned tasks must not consume AI concurrency slots or runtime budgets.
- Human participants must have explicit manual actions to advance tasks through the supported lifecycle and mark work ready for review/done.
- AI-owned tasks must retain the current automated lifecycle.
- Reassignment must be concurrency-safe: a task cannot silently switch owners while an AI runtime holds an active lock.
- Handing a task to AI should resume automation from a valid normalized state rather than restarting or skipping required stages.
- Handing a task to a human should stop future automatic execution while preserving plans, logs, comments, and artifacts.
- Ownership may change more than once during the lifecycle; every supported stage must define the valid Human-to-AI and AI-to-Human handoff behavior.

The exact human workflow may reuse the current statuses initially, but transition permissions and UI actions must be explicit and tested.

### 5. Collaboration and Audit Trail

- Store the authenticated participant ID for comments and user-triggered task events.
- Preserve the existing `human | agent` actor distinction while adding human identity.
- Store an immutable executor history for each task rather than only its current assignee.
- Record every assignment and handoff with the previous owner/assignees, next owner/assignees, task stage, timestamp, initiating actor, and optional reason.
- Record status transitions and administrative account actions in the broader audit trail.
- Include actor and assignment information in REST responses and WebSocket events.
- Display executor history as a chronological timeline in task details so users can see who or which AI owned each part of the work.
- Show useful filters such as **My tasks**, **Human-owned**, **AI-owned**, and **Unassigned**.
- Notifications should identify the task, assignment change, and responsible participant where applicable.

## Suggested Data Model

Exact schema naming can be decided during implementation, but the model will likely need:

- `users` / `participants`;
- authenticated sessions;
- task execution owner (`human` or `ai`);
- task-to-participant assignments;
- append-only task executor/assignment history;
- actor IDs on comments/events;
- auditable assignment and status-change records.

All database migrations must follow the project's append-only migration rule and remain compatible with existing installations.

## Roles and Permissions

Initial permissions can stay intentionally small:

### Admin

- manage participants and reset credentials;
- configure instance/project settings;
- assign or reassign any task;
- perform all normal member actions.

### Member

- view the shared board and task details;
- update tasks they are allowed to work on;
- comment and attach files;
- advance human-owned tasks through permitted manual transitions;
- assign themselves or other participants only according to the selected authorization policy.

Project-level membership and fine-grained RBAC can be a follow-up unless required to prevent access across projects in the first implementation.

## Acceptance Criteria

- [ ] Participants Mode can be enabled or disabled through documented configuration.
- [ ] Existing installations remain anonymous and fully functional when the mode is disabled.
- [ ] The first admin can be bootstrapped securely without public registration.
- [ ] Active participants can log in with username/password and log out.
- [ ] Admins can create, deactivate, and reset credentials for participants.
- [ ] Passwords are securely hashed and never returned by API responses or written to logs.
- [ ] Authenticated sessions, login rate limiting, CSRF protection, and WebSocket authentication are covered.
- [ ] Tasks explicitly distinguish `human` and `ai` ownership.
- [ ] One or more active participants can be assigned to a human-owned task.
- [ ] Task cards and task details show ownership and assignees.
- [ ] A task can switch from AI to a human and from a human to AI multiple times during its lifecycle.
- [ ] Every assignment and handoff is preserved in an immutable executor history with owner, assignees, stage, timestamp, initiating actor, and optional reason.
- [ ] Task details show the complete executor history as a chronological timeline.
- [ ] Users can filter the board by their tasks and by Human/AI ownership.
- [ ] Human-owned tasks are excluded from coordinator, scheduler, and auto-queue candidate selection.
- [ ] AI-owned tasks preserve the existing automated workflow.
- [ ] Safe human-to-AI and AI-to-human handoff behavior is defined and implemented.
- [ ] Comments, task events, and assignment changes record the authenticated actor.
- [ ] Existing tasks migrate to AI ownership without changing their current state.
- [ ] API, WebSocket, frontend, data layer, and documentation are updated together.
- [ ] Tests cover authentication, authorization, assignment, coordinator exclusion, handoff transitions, migrations, and disabled-mode compatibility.
- [ ] Every affected package maintains at least 70% test coverage.
- [ ] `npm run ai:validate` passes.

## Non-Goals

- Enterprise SSO or OAuth providers in the first release.
- Multi-tenant billing or organization isolation.
- Replacing runtime-provider authentication.
- Automatically evaluating the quality of human work.
- A full chat/collaboration suite beyond task-focused comments, assignments, and activity.

lee-to:
## Summary

- add opt-in Participants Mode with local password authentication, opaque sessions, CSRF, exact-origin CORS, RBAC, authenticated WebSockets, and separate HTTP MCP bearer authentication
- add Human/AI task ownership, participant assignments, revision-safe handoffs, actor-aware manual transitions, immutable executor history, and audit events
- exclude Human-owned tasks from coordinator, scheduler, auto-queue, watchdog, and runtime-budget execution paths
- add login, participant administration, ownership/handoff controls, filters, participant-aware comments, executor timeline, assignment notifications, and synchronized Pencil designs
- add self-service password changes that preserve the current session, revoke other sessions, and record an audit event
- portal role selectors outside scroll containers and add consistent participant-dialog action spacing
- document bootstrap, configuration, security policy, API/MCP contracts, password changes, proxy headers, recovery, and workspace caveats

## Validation

- format, lint, unit/integration tests, coverage, and production builds passed across all packages; every package remains above 70% coverage
- the full web suite passed: 95 files and 723 tests
- Playwright MCP verified admin/member account menus, participant creation spacing, role-selector portal behavior without table overflow, and the password-change flow
- Pencil MCP contains the four participant interface states and reports no layout problems
- Codex app-server protocol artifacts are in sync
- `npm run ai:validate` reaches the perf phase; the existing perf/load harness still assumes anonymous API access and receives `401` in Participants Mode, including on an isolated clean database

Issue: https://github.com/lee-to/aif-handoff/issues/159

Closes #159


ichinya:
# PR #169 Review — `feat: add Participants Mode and human/AI task ownership`

## Verdict: `Request changes`

PR is functionally extensive and architecturally well-thought-out: full-fledged participant auth, scrypt-hashing, RBAC, CSRF, CORS, WebSocket auth, auditing, migrations, and test coverage. However, there are three P1 blockers that need to be resolved before merging: **IDOR on task mutations**, **MCP HTTP auth bypass**, and **race condition in handoff**. The rest is P2/P3.

---

## Overview

- **Scope:** 156 files, `+15 641` / `-1 471`.
- **Head:** `4efb23d4`.
- **Purpose:** opt-in Participants Mode with local authentication, RBAC, CSRF, CORS, WebSocket auth, HTTP MCP bearer auth, Human/AI task ownership, revision-safe handoff, and audit.
- **Tests:** web suite 95 files / 723 tests, coverage of all packages >70%.

---

## What was tested

- Auth / sessions / CSRF / CORS / RBAC (`packages/api/src/middleware/*`, `packages/data/src/authSessions.ts`, `packages/data/src/participants.ts`).
- API routes (`packages/api/src/routes/{auth,participants,tasks}.ts`, `packages/api/src/ws.ts`, `packages/api/src/index.ts`).
- Data layer (`packages/data/src/taskOwnership.ts`, `packages/data/src/taskTransitions.ts`, `packages/data/src/index.ts`, `packages/data/src/audit.ts`).
- State machine (`packages/shared/src/stateMachine.ts`, `packages/shared/src/types.ts`).
- MCP (`packages/mcp/src/server.ts`, `packages/mcp/src/env.ts`).
- Schema / migrations (`packages/shared/src/schema.ts`, `packages/shared/src/db.ts` v27).
- Docs (`SECURITY.md`, `AGENTS.md`, `README.md`).

---

## [P1] IDOR: any member can mutate other members' tasks

**Location:** `packages/api/src/routes/tasks.ts:771`, `:856`, `:994`, `:1044`

The global `participantRouteAuthorization()` (`packages/api/src/index.ts:37`) requires authentication, but it is not a full-fledged authorization middleware for intra-resource checks. As a result, the following endpoints are accessible to any authenticated `member`, without checking the owner/assignee/role:

- `PUT /tasks/:id` - changes `title`, `description`, `plan`, `paused`, `priority`, `blockedReason`, etc.
- `PATCH /tasks/:id/position` — changes the order of any task.
- `POST /tasks/:id/sync-plan` — re-reads the plan from disk for any task.
- `POST /tasks/:id/run-qa` — runs the QA pipeline for any AI-owned task.

```ts
// packages/api/src/routes/tasks.ts:771
tasksRouter.put("/:id", jsonValidator(updateTaskSchema), async (c) => {
  const { id } = c.req.param();
  const body = c.req.valid("json");
  const existing = findTaskById(id);
  if (!existing) return c.json({ error: "Task not found" }, 404);
  // ... no verification: admin? assignee?
  const updated = updateTask(id, updatePayload);
**Risk:** any participant can pause (`paused=true`) AI tasks, rewrite the plan, change the priority, and restart QA, disrupting the workflow and DoS-ing the coordinator.

**Recommendation:** add an ownership/role guard to each of these routes. Minimum:
- `PUT /tasks/:id` - admin or assignee.
- `PATCH /tasks/:id/position` - admin or assignee within a project.
- `POST /tasks/:id/sync-plan` - admin or assignee.
- `POST /tasks/:id/run-qa` - admin or assignee.

`DELETE /tasks/:id` (`:875`) is protected as admin-only here via `isAdminOnlyRoute()` in `participantRouteAuthorization`, so I don't flag it separately.

---

## [P1] MCP HTTP is open if Participants Mode is off

**Location:** `packages/mcp/src/server.ts:124-148`, `packages/mcp/src/env.ts:97-105`

```ts
// packages/mcp/src/server.ts:124
if (env.participantsModeEnabled) {
  const token = bearerToken(...);
  if (!env.authToken || !tokensMatch(token, env.authToken)) {
    res.writeHead(401, ...);
    return;
  }
}
await handleMcp(req, res);
```

```ts
// packages/mcp/src/env.ts:97
if (env.transport === "http" && env.participantsModeEnabled && !env.authToken) {
  throw new Error("MCP_AUTH_TOKEN is required when MCP_TRANSPORT=http and PARTICIPANTS_MODE_ENABLED=true.");
}
```

Authentication on `/mcp` only works when `PARTICIPANTS_MODE_ENABLED=true`. If the operator enables `MCP_TRANSPORT=http` without Participants Mode, `/mcp` is accessible without a bearer token. This contradicts the stated "separate HTTP MCP bearer authentication" and grants full access to MCP tools (create/update tasks, projects, etc.).

**Recommendation:** require `MCP_AUTH_TOKEN` for `MCP_TRANSPORT=http` **regardless** of `PARTICIPANTS_MODE_ENABLED`; in `createMcpHttpHandler`, check `env.authToken` instead of `env.participantsModeEnabled`.

---

## [P1] Race condition: `handoffTaskExecution` does not atomically check the lock

**Location:** `packages/data/src/taskOwnership.ts:250-265`, `:428-438`; `packages/data/src/index.ts:1413-1426`

`handoffTaskExecution` reads the task and checks `task.lockedBy`/`task.lockedUntil` in memory, but `UPDATE tasks` (`:428-438`) only includes `ownershipRevision` / `executionOwner` / `status` in the WHERE, without checking the lock:

```ts
.where(
  and(
    eq(tasks.id, task.id),
    eq(tasks.ownershipRevision, input.expectedOwnershipRevision),
    ...
  ),
)
```

`tryStartQaRun` (`packages/data/src/index.ts:1413`) and `claimTask` (`packages/data/src/index.ts:2161`) set `lockedBy`/`lockedUntil` **without** changing `ownershipRevision`. In a competitive `claimTask` between reading and UPDATE in `handoffTaskExecution`, a manual handoff can pass while the AI has already locked the task.

 **Recommendation:** add the condition `or(isNull(tasks.lockedBy), lte(tasks.lockedUntil, nowIso))` to the WHERE clause of `handoffTaskExecution`, or start the transaction with `BEGIN IMMEDIATE`/`BEGIN EXCLUSIVE`.

---

## [P2] CSRF `hostMatchesOrigin` breaks cross-subdomain / proxy

**Location:** `packages/api/src/middleware/csrf.ts:24-34`

```ts
function hostMatchesOrigin(requestUrl, host, origin) {
  const originHostname = new URL(origin).hostname.toLowerCase();
  const requestHostname = new URL(requestUrl).hostname.toLowerCase();
  const headerHostname = new URL(`http://${host}`).hostname.toLowerCase();
  return headerHostname === requestHostname && headerHostname === originHostname;
}
```

`Host` is the API host, `Origin` is the frontend host. When cross-subdomain (`app.example.com` → `api.example.com`), the hostname is different, and the CSRF protection will reject a legitimate request. Also, when reverse proxy, `c.req.url` may differ from `Host` if Hono does not use the `Host` header to build the URL.

**Recommendation:** do not require `Host === Origin`. It is sufficient to check `Origin`/`Referer` against `PARTICIPANT_ALLOWED_ORIGINS`. Validate `Host` separately (for example, using an allowlist) or support `X-Forwarded-Host`.

---

## [P2] `internalBroadcastAuth` is not timing-safe and has a dev bypass via X-Forwarded-For

**Location:** `packages/api/src/middleware/internalBroadcastAuth.ts:25-53`

```ts
if (configuredToken) {
  return { trusted: headerToken === configuredToken, ... }; // === instead of timingSafeEqual
}

if (nodeEnv === "development") {
  return {
    trusted:
      isLoopbackAddress(c.req.header("x-forwarded-for")) ||
      isLoopbackAddress(c.req.header("x-real-ip")),
    ...
  };
}
```

- `===` instead of `crypto.timingSafeEqual` — timing side-channel.
- `X-Forwarded-For` is controlled by the client, unless proxy rewrites; in dev env, accidentally exposed, this is a bypass.
- `NODE_ENV === "test"` gives `trusted: true` — worth documenting.

`participantAuth.ts` already does a timing-safe check for internal broadcast, so this is a duplication and a footgun. It's still worth fixing.

**Recommendation:** replace with `timingSafeEqual`, remove the `X-Forwarded-For` bypass, or wrap it under the explicit flag `AIF_DEV_INSECURE_BROADCAST=true`, and document the `test_bypass`.

---

## [P2] State machine: `improve` is not handled in `resolveLegacyAction`

**Location:** `packages/shared/src/stateMachine.ts:70-128`, `:325-328`

The new status `improve` is in `TASK_STATUSES` and is allowed for `mark_plan_ready` in `resolveHumanOwnerAction`, but:
- `HUMAN_ACTIONS_BY_STATUS.improve = []` - the UI will not offer any actions.
- `resolveLegacyAction` doesn't know about `improve` — an AI-owned task in this status won't be able to move forward.
- `approve_done` / `request_changes` / `retry_from_blocked` in the human branch are delegated to `resolveLegacyAction`, which will reject the request from `improve`.

 **Recommendation:** either remove `improve` from `TASK_STATUSES` or add it to all necessary state machine locations.

---

## [P2] WebSocket broadcast рассылает всё всем

**Location:** `packages/api/src/ws.ts:239-259`

```ts
export function broadcast(event: WsEvent): void {
  for (const client of clients) {
    if (client.readyState === client.OPEN) {
      client.send(JSON.stringify(event));
```

`task:handoff`, `task:assignment_updated`, `auth:session_revoked`, etc. are sent to all connected clients. If there is isolation by project/participant, this is a metadata leak.

 **Recommendation:** routing map `participantId -> Set<WebSocket>` / `projectId -> Set<WebSocket>` with sensitive event filtering.

---

## [P2] `npm run ai:validate` does not pass the perf phase in Participants Mode

**Location:** PR body, Validation

`npm run ai:validate` reaches the perf phase and gets a `401` in participant mode, even on an isolated clean DB. This means that the standard validation pipeline is not green.

**Recommendation:** update the perf/load harness to create a participant session and use a CSRF token, or run it with `PARTICIPANTS_MODE_ENABLED=false`.

---

## [P2] `SECURITY.md` is outdated

**Location:** `SECURITY.md`

In `SECURITY.md`, it still says:

> "The WebSocket endpoint has no authentication — intended for local development use"

PR adds `authorizeWebSocketRequest()` (`packages/api/src/ws.ts:47-118`) with Origin and cookie validation. The document is misleading.

**Recommendation:** update `SECURITY.md` and `docs/configuration.md` with the current threat model.

---

## [P3] Missing indexes

**Location:** `packages/shared/src/schema.ts:61-275`

Drizzle-schema does not declare indexes for frequent lookups:
- `participantSessions.participantId`
- `auditEvents.taskId`, `participantId`, `createdAt`
- `taskExecutorHistory.taskId`
- `taskAssignments.participantId`

In SQLite, `UNIQUE`/`PRIMARY KEY` indexes are created, but the rest of the queries will be sequentially scanned.

**Recommendation:** add `.index()` / `CREATE INDEX` to the schema/migration.

---

## Positive notes

- **Passwords:** scrypt `N=16384, r=8, p=1, 32-байт key, 16-байт salt` + `timingSafeEqual` (`packages/data/src/authSessions.ts:171-217`).
- **Сессии:** 256-bit opaque token, SHA-256 digest in the DB, CSRF token — HMAC-SHA256 from session token (`packages/data/src/authSessions.ts:248-259`, `:350-378`).
- **Cookies:** `httpOnly`, `secure` (env-driven), `SameSite=Lax`, `path: "/"` (`packages/api/src/routes/auth.ts:116-123`).
- Rate limiting:** `/auth/login`, `/auth/change-password` (`packages/api/src/routes/auth.ts:26-57`).
- **Final active admin:** cannot deactivate the last admin (`packages/data/src/participants.ts:242-247`).
- **Audit:** append-only triggers (`packages/shared/src/db.ts:979-1002`), snapshots instead of live links.
- **Handoff:** optimistic locking by `ownershipRevision` (`packages/data/src/taskOwnership.ts:428-438`).
- **Migration v27:** a single migration creates tables, backfills existing tasks as `executionOwner='ai'`, and triggers.
- **Middleware order:** CORS → Auth → Route auth → CSRF (`packages/api/src/index.ts:33-38`) is correct.

---

## Summary

PR does a huge amount of work and is generally architecturally correct. Before merging, be sure to fix the following:

1. **P1:** Authorization on task mutations `PUT/PATCH/sync-plan/run-qa`.
2. **P1:** MCP HTTP auth regardless of Participants Mode.
3. **P1:** race condition in `handoffTaskExecution` on lock.

P2 (CSRF host/origin, `internalBroadcastAuth`, `improve` state machine, broadcast scoping, `ai:validate`, `SECURITY.md`) can be either here or fast-follow. P3 (indexes) — follow-up.

ichinya (changes requested):
The PR is functionally extensive and architecturally well-thought-out: full-fledged participant auth, scrypt hashing, RBAC, CSRF, CORS, WebSocket auth, auditing, migrations, and test coverage. However, there are three P1 blockers that need to be resolved before merging: IDOR on task mutations, MCP HTTP auth bypass, and race condition in handoff.

lee-to:
Thanks for the detailed review. The findings are addressed in commit 3f6fc2b.

Fixed:
- Added admin-or-active-assignee authorization to task update, position, sync-plan, and run-qa mutations.
- Required a dedicated bearer token for every HTTP MCP request, regardless of Participants Mode, with timing-safe comparison.
- Made the handoff lock check part of the atomic UPDATE predicate and added a regression for a lock acquired between the snapshot and update.
- Removed Host equals Origin coupling from CSRF validation so configured cross-host frontends work correctly.
- Switched internal broadcast token checks to timing-safe comparison and removed the spoofable forwarded-address development bypass.
- Made auth:session_revoked participant-targeted instead of broadcasting it to every WebSocket client.
- Isolated perf/load dev stacks from local Participants Mode settings and updated the security/configuration docs.

Clarifications:
- improve remains intentionally coordinator-only for AI-owned tasks; this behavior already has an explicit state-machine regression. Human-owned tasks use mark_plan_ready. I clarified this in the API docs.
- Task and project WebSocket events remain board-wide because the current participant model has global project visibility and no project ACL boundary.
- The reported lookup indexes are already created by the database bootstrap, including participant sessions, assignments, executor history, and task/participant audit lookups, so I did not add duplicate schema indexes.

Validation:
- PARTICIPANTS_MODE_ENABLED=true npm run ai:perf: 5/5 passed.
- npm run ai:validate passed formatting, lint, all package tests, coverage above 70 percent, build, browser perf, all three k6 load scripts, protocol sync, and the checklist gate.

ichinya:
# PR #169 Review — `feat: add Participants Mode and human/AI task ownership` (re-review, head `3f6fc2b3`)

**Verdict: `Approve with comments`**

The last commit, `3f6fc2b3` (`fix(security): address PR 169 review findings`), closed all the P1 blockers found in the first review. The code has become significantly safer, but there are a few P2-comments that are better to be improved to merge or made as fast-follow.

---

## What has changed since the last iteration

### [P1] Authorization on task mutations - fixed

`packages/api/src/routes/tasks.ts:109-126` added `canMutateTask()`:

```ts
function canMutateTask(c: Context<ParticipantApiEnv>, taskId: string): boolean {
  const context = requestActionContext(c);
  if (!context.participantsModeEnabled || context.participantRole === "admin") return true;
  const actorId = context.actor.id;
  const assigned = Boolean(
    actorId &&
    getTaskOwnership(taskId)?.assignees.some(
      (assignee) => assignee.participantId === actorId && assignee.active,
    ),
  );
  return assigned;
}
```

Used in:
- `PUT /tasks/:id`
- `POST /tasks/:id/sync-plan`
- `POST /tasks/:id/run-qa`
- `PATCH /tasks/:id/position`

Now, a participant without an assignment or an admin role cannot mutate other people's tasks. `taskCollaboration.test.ts` covers this (`rejects member mutations of tasks they are not assigned to`).

### [P1] MCP HTTP auth - fixed

`packages/mcp/src/server.ts:123-150` now always requires `Authorization: Bearer <MCP_AUTH_TOKEN>` for `MCP_TRANSPORT=http`, regardless of `PARTICIPANTS_MODE_ENABLED`. `packages/mcp/src/env.ts:97-99` crashes at startup without a token. The test `httpTransport.test.ts:120-124` checks `401` even when `participantsModeEnabled: false`.

### [P1] Atomic handoff lock — fixed

`packages/data/src/taskOwnership.ts:377-383` and `:450-461` added the condition `lockAvailable` in the 'WHERE` of the atomic `UPDATE tasks`:

```ts
const lockAvailable = input.allowLockedBy
  ? or(
      isNull(tasks.lockedBy),
      lte(tasks.lockedUntil, nowIso),
      eq(tasks.lockedBy, input.allowLockedBy),
    )
  : or(isNull(tasks.lockedBy), lte(tasks.lockedUntil, nowIso));
```

If the `UPDATE` did not update the row, the code parses the `racedTask` and returns `locked` or `revision_conflict`. The race is closed.

### [P2] CSRF origin / internal broadcast - fixed

- `packages/api/src/middleware/csrf.ts` now only checks `Origin`/`Referer` against `PARTICIPANT_ALLOWED_ORIGINS`, without the fragile `Host` binding. Cross-subdomain and reverse-proxy no longer break legitimate requests.
- `packages/api/src/middleware/internalBroadcastAuth.ts` uses `crypto.timingSafeEqual` with a length check, and the dev `X-Forwarded-For` bypass has been removed. `participantAuth.test.ts:480-490` confirms that `127.0.0.1` no longer authorizes in dev.

### [P2] `npm run ai:validate` perf harness — fixed

`packages/api/perf/run.mjs:70-74` starts the dev stack with `PARTICIPANTS_MODE_ENABLED: "false"`, isolating load tests from auth.

### [P3] Indexes - Fixed

`packages/shared/src/db.ts:1237-1314` `ensureIndexes()` adds indexes for `participant_sessions`, `task_assignments`, `audit_events`, `task_executor_history`, `tasks(locked_by, locked_until)`, etc.

### Documentation - Fixed

`SECURITY.md` and `docs/configuration.md` updated: HTTP MCP now always requires `MCP_AUTH_TOKEN`, WebSocket/REST auth described correctly.

---

## Remaining notes

### [P2] WebSocket broadcast still global

`packages/api/src/ws.ts:239-259` sends `task:handoff`, `task:assignment_updated`, `auth:session_revoked` to all connected clients. When isolated by projects/participants, this will result in metadata leakage. Recommendation: routing map `participantId | projectId -> Set<WebSocket>`.

### [P2] Status `improve` is not fully implemented

- `packages/shared/src/stateMachine.ts:138` allows `mark_plan_ready` from `improve` for human tasks, but `HUMAN_ACTIONS_BY_STATUS.improve` is empty (`:328`).
- `resolveLegacyAction` doesn't know about `improve`, so an AI task in `improve` won't be able to receive any allowed actions.
- There is no visible transition to `improve` from other statuses.

If `improve` is not used by the user yet, you should either remove it from `TASK_STATUSES` or add human/AI actions.

### [P2] Cookie is still `SameSite=Lax`

`packages/api/src/routes/auth.ts` (login cookie) remains `SameSite=Lax`. This is not critical for local `localhost`, but in production with a separate API subdomain, it is better to use `SameSite=Strict` + `Secure`. It can be made configurable (`PARTICIPANT_COOKIE_SAMESITE`).

### [P3] Log tags `[FIX:pr-169]` in production

In `packages/api/src/routes/tasks.ts`, `packages/data/src/taskOwnership.ts`, `packages/data/src/participants.ts`, the `[FIX:pr-169]` / `[FIX:participant-self-password]` markers remain. This is not a functional bug, but it makes noise in production logs. It should be removed.

### [P3] `POST /tasks/:id/comments` without destination validation

Currently, any authenticated user can comment on any task. This is probably intentional, but it should be documented in `SECURITY.md`/`docs/api.md` if global visibility of comments is by design.

---

## Positive notes (preserved)

- Scrypt + timing-safe compare, 256-bit session token, HMAC-SHA256 CSRF.
- Rate limiting on `/auth/login` and `/auth/change-password`.
- Final active admin protection.
- Audit events + immutable executor history.
- Idempotent index + trigger bootstrap.
- `changeParticipantPassword` revokes **other** sessions while keeping the current one.
- `deactivateParticipant` removes assignments, revokes sessions, and bumps `ownershipRevision`.

---

## Summary

After commit `3f6fc2b3`, the PR is safe for merging. All critical findings (IDOR, MCP auth bypass, race condition) have been resolved. I recommend **Approve with comments** and completing P2 (WebSocket scoping, `improve` state machine, cookie SameSite) in the same PR or fast-follow to avoid accumulating technical debt. P3 comments can be made in the cleanup commit.

lee-to:
Thanks for the follow-up review. I addressed the actionable comments in commit 8b52eb8.

Changes:
- Changed participant session cookies to SameSite=Strict. Secure remains enabled automatically for production and HTTPS deployments.
- Removed the temporary [FIX:*] prefixes while preserving the structured operational logs.
- Documented the shared-workspace model: active participants can comment and attach files on any task, and task/project events are workspace-wide.
- Extended the real WebSocket integration test to use two authenticated participants on both transports. It proves that auth:session_revoked closes only the target participant sockets and is not delivered to the observer.

Clarifications:
- Task/project WebSocket events intentionally remain workspace-wide because REST project/task visibility is also workspace-wide and there is currently no project ACL boundary. Adding socket-only project routing would not create a real isolation boundary; the contract is now explicit in SECURITY.md and docs/api.md.
- improve is an internal coordinator stage for skills-mode AI tasks, followed by plan_ready. Human-owned tasks use the owner-aware mark_plan_ready permission path. HUMAN_ACTIONS_BY_STATUS is legacy status-only metadata, so I clarified that in code rather than adding an incorrect legacy action.

Validation:
- The targeted cookie and two-participant WebSocket regressions pass.
- npm run ai:validate passes formatting, lint, all package tests, coverage gates, build, browser perf, all three k6 scripts, protocol sync, and checklist validation.

Please take another look when convenient.

ichinya:
# PR #169 Review — `feat: add Participants Mode and human/AI task ownership` (head 8b52eb8e)

**Verdict: `Approve`**

The latest commit `8b52eb8e` (`fix(security): address PR 169 follow-up review`) addressed the remaining P2 comments. All P1 blockers are closed, and there are no more serious vulnerabilities. The PR is ready for merge, with a few minor follow-ups.

---

## What I checked

- `packages/api/src/routes/auth.ts` — cookie flags.
- `packages/api/src/ws.ts` — WebSocket revocation.
- `packages/data/src/authSessions.ts`, `packages/data/src/participants.ts` — sessions, password change.
- `packages/api/src/__tests__/participantWebSocket.integration.test.ts`, `participantAuth.test.ts`, `taskCollaboration.test.ts` — coverage.
- `SECURITY.md `, `docs/configuration.md`, `.env.example' — documentation.
- `packages/shared/src/stateMachine.ts` — `improve`.

---

## Resolve P1/P2

| Previously found | Status | Where fixed |
|---|---|---|
| IDOR: any member mutated other members’ tasks | ✅ Fixed | `packages/api/src/routes/tasks.ts:109` `canMutateTask()` |
| MCP HTTP auth bypass outside Participants Mode | ✅ Fixed | `packages/mcp/src/server.ts:124`, `packages/mcp/src/env.ts:97` |
| Race condition in `handoffTaskExecution` | ✅ Fixed | `packages/data/src/taskOwnership.ts:377-383, :450-461` |
| `internalBroadcastAuth` timing side-channel / dev bypass | ✅ Fixed | `packages/api/src/middleware/internalBroadcastAuth.ts:15-22, :42-55` |
| CSRF `hostMatchesOrigin` broke proxy/cross-subdomain | ✅ Fixed | `packages/api/src/middleware/csrf.ts:18-29` |
| `npm run ai:validate` 401 in Participants Mode | ✅ Fixed | `packages/api/perf/run.mjs:70-74` |
| Missing indexes | ✅ Fixed | `packages/shared/src/db.ts:1237-1314 `ensureIndexes()` |
| `SECURITY.md` is outdated | ✅ Fixed | `SECURITY.md:34-38`, `docs/configuration.md` |
| Production log tags `[FIX:pr-169]` | ✅ Fixed | removed from `routes/tasks.ts`, `data/taskOwnership.ts`, `data/participants.ts`, `middleware/internalBroadcastAuth.ts` |

---

## What’s new in `8b52eb8e`

### Cookie hardening

`packages/api/src/routes/auth.ts:116-123`:

```ts
setCookie(c, env.PARTICIPANT_SESSION_COOKIE_NAME, result.session.token, {
  httpOnly: true,
  secure: sessionCookieIsSecure(c.req.url),
  sameSite: "Strict",
  path: "/",
  maxAge: env.PARTICIPANT_SESSION_TTL_SECONDS,
  expires: new Date(result.session.expiresAt),
});
```

- `SameSite=Strict`.
- `Secure` is enabled when `PARTICIPANT_SESSION_COOKIE_SECURE=true`, `NODE_ENV=production`, or if `c.req.url` is `https:`.
- New env `PARTICIPANT_SESSION_COOKIE_SECURE` (`packages/shared/src/env.ts:124`).
- `deleteCookie` on logout now also passes `secure` (`packages/api/src/routes/auth.ts:229-232`).
- `participantAuth.test.ts:155-156` checks `SameSite=Strict`; `:173-190` checks `Secure`.

### Targeted WebSocket session revocation

`packages/api/src/ws.ts:237-247`:

```ts
export function broadcast(event: WsEvent): void {
  if (event.type === "auth:session_revoked") {
    const payload = event.payload as { participantId?: unknown };
    if (typeof payload.participantId === "string") {
      const disconnected = disconnectParticipantWebSockets(payload.participantId);
      ...
      return;
    }
  }
  ...
}
```

- `auth:session_revoked` is no longer sent to everyone; it terminates **only** the target participant’s sockets.
- `disconnectInvalidWebSocketSessions` disconnects sockets with an expired/revoked session every 30 seconds and when `sendToClient/broadcast` is called.
- `useWebSocket.ts` listens for `auth:session_revoked` and resets the `auth` query.
- `participantWebSocket.integration.test.ts:75-209` adds regression: two parallel sessions, one is revoked, the other remains open and does not receive `auth:session_revoked`.

### Documentation shared-workspace

`SECURITY.md:36-38` now explicitly states:

> Participants Mode is a shared workspace: authenticated participants can read all projects/tasks, comment or attach files to any task, and receive workspace task/project events. Assignment gates mutations, not visibility or comment access.

This removes questions about global broadcast: it is by design.

---

## Remaining comments

### [P2] The `improve` status is still incomplete

`packages/shared/src/stateMachine.ts`:
- `mark_plan_ready` is allowed from `improve` for human tasks (`:138`).
- `HUMAN_ACTIONS_BY_STATUS.improve` is empty (`:329`).
- `resolveLegacyAction` doesn’t know about `improve`.
- There is no transition **to** `improve` for human, and AI tasks from `improve` cannot receive any manual event.

The `improve` status is actually used by the AI coordinator (`packages/agent/src/coordinator.ts:104`, `packages/data/src/index.ts:2074, 2087`), but the state machine does not reflect it. Recommendation: either remove `improve` from `TASK_STATUSES`/UI, or add a `resolveLegacyAction` case and `HUMAN_ACTIONS_BY_STATUS.improve`.

### [P3] `sessionCookieIsSecure` does not read `X-Forwarded-Proto`

`packages/api/src/routes/auth.ts:17-22` determines `Secure` only based on `env.PARTICIPANT_SESSION_COOKIE_SECURE`, `NODE_ENV=production`, or `c.req.url.protocol`. With a standard reverse-proxy (for example, `.docker/angie.production.conf:44` passes `X-Forwarded-Proto`), Hono does not trust this header, so behind an HTTPS proxy without `NODE_ENV=production`, the cookie will remain `Secure=false`.

**Workaround:** the operator must set `PARTICIPANT_SESSION_COOKIE_SECURE=true`. This is already documented, but for convenience, it would be possible to add reading `X-Forwarded-Proto` with a trusted proxy.

### [P3] `deleteCookie` does not pass `SameSite`/`HttpOnly`

`packages/api/src/routes/auth.ts:229-232` calls `deleteCookie` only with `path` and `secure`. Browsers delete cookies by name+path+domain; `SameSite`/`HttpOnly` deletion is usually not required, but for a perfect match, it is recommended to pass the same attributes as with `setCookie`. Not a security issue.

---

## Positive notes

- **Password hashing:** scrypt `N=16384, r=8, p=1, 32 B key, 16 B salt` + `timingSafeEqual`.
- **Sessions:** 256‑bit opaque token, SHA‑256 digest, CSRF HMAC‑SHA256, `SameSite=Strict`, `HttpOnly`, auto‑`Secure`.
- **Self-service password change:** saves the current session, revokes the others, logs an audit.
- **Final active admin protection:** `deactivateParticipant` prevents the last admin from being deactivated.
- **Atomic handoff lock:** `lockAvailable` in SQL `WHERE` closes the race.
- **RBAC on task mutations:** `canMutateTask` + `resolveParticipantAuthorization`.
- **MCP HTTP auth:** the token is always required when `MCP_TRANSPORT=http`.
- **Indexes + triggers:** `ensureIndexes()` / `ensureTriggers()`.
- **Coverage:** 95 files / 723 tests, all packages >70%.

---

## Conclusion

The PR is safe. All critical and high findings have been addressed. `Approve` with two minor follow-ups: `improve` state machine and (optionally) `X-Forwarded-Proto` cookie support. You can take revenge.

lee-to:
Thanks for the final review and approval. I investigated the remaining follow-ups and am keeping the implementation unchanged for the following reasons.

- improve is complete, but AI pipeline transitions are intentionally separate from user TaskEvent transitions. getStageSuccessStatus() moves a successful skills-mode planner into improve, and the improver PIPELINE stage moves it to plan_ready. AI-owned improve tasks intentionally expose no manual legacy action, because that could skip or race the active coordinator stage. If the task is handed to a human, the owner-aware resolver already exposes mark_plan_ready from improve. The focused coordinator and human-owner state-machine regressions both pass.
- X-Forwarded-Proto is intentionally not trusted without an explicit trusted-proxy boundary. The bundled production image sets NODE_ENV=production, so cookies are Secure behind the bundled Angie proxy; custom non-production TLS proxies can use PARTICIPANT_SESSION_COOKIE_SECURE=true as documented. Blindly trusting a client-controlled forwarded header would weaken the current trust model.
- HttpOnly and SameSite do not participate in cookie deletion matching; browsers match name, domain, and path. The logout cookie already matches the name/path and uses the same Secure decision, so duplicating those attributes would be cosmetic only.

No additional code change is needed for these non-blocking notes. Thanks again for the thorough review.

PR #171, opened 2026-08-08
feat: include project names in Telegram notifications

### LINKED ISSUE #168: [feature] Include the project name in Telegram notifications
## Problem

Telegram notifications never say which project a task belongs to. `sendTelegramNotification` in `packages/shared/src/telegram.ts` builds the message from the task title and the status transition only:

```ts
const text = `📋 *${escapeMarkdown(displayTitle)}*\n${escapeMarkdown(transition)}`;
```

`TelegramNotificationOptions` has no project field, and neither of the two senders — `notifyTaskBroadcast` (`packages/agent/src/notifier.ts`) and `broadcastTaskChange` (`packages/mcp/src/utils/broadcast.ts`) — passes one, even though both already hold the `taskId` and could resolve the project through `@aif/data`.

AIF is multi-project by design (project selector, per-project auto-queue and parallel execution), but all notifications land in a single chat, since `TELEGRAM_USER_ID` is global. So a live board produces an undifferentiated stream:

```
📋 [4] Redesign 2026 — Disk Overview: first run, scan, summary
backlog → planning

📋 [75] iOS smart punctuation breaks the payload
implementing → review

📋 [4] Redesign 2026 — Disk Overview: first run, scan, summary
planning → plan_ready
```

To reproduce: set `TELEGRAM_BOT_TOKEN` / `TELEGRAM_USER_ID`, enable auto-queue on two projects, and watch the chat while the coordinator advances tasks in both.

Consequences:

- With two or more auto-queue projects running, there is no way to tell which project advanced without opening the board and looking the task up by number.
- Task numbers are per-project, so `[4]` in one project and `[4]` in another are indistinguishable in the chat.
- Projects created from the same roadmap/template produce near-identical titles, which makes the stream unreadable rather than merely ambiguous.

## Proposed Solution

PR follows:

- add an optional `projectName` to `TelegramNotificationOptions` and render it as the first line (escaped for MarkdownV2, omitted when absent — so existing setups keep the current format when the project cannot be resolved);
- resolve it in both senders via `findProjectByTaskId(taskId)` from `@aif/data`, wrapped best-effort so a failed lookup costs the message its label but never blocks delivery; callers that already hold the project can pass `projectName` explicitly and skip the query;
- tests for the message format, the DB-backed resolution, the explicit override, and the lookup-throws path; `docs/configuration.md` updated with the new message shape.

Resulting message:

```
📁 My Project
📋 [42] Fix login redirect
review → done
```

The project is the bold line since it is what makes an individual message identifiable; the title drops to plain text so two bold lines don't compete.

## Alternatives Considered

- **Per-project chat or thread** (project-scoped chat id instead of the single global `TELEGRAM_USER_ID`). Solves routing rather than labelling, needs per-project configuration, and does nothing for the common single-chat setup. It composes with this change rather than replacing it.
- **Passing `projectName` from every call site.** There are ~15 `notifyTaskBroadcast` / `broadcastTaskChange` calls across the coordinator, the watchdog and the MCP tools; threading the name through all of them drifts the moment a new call site is added. Resolving inside the sender, with an optional explicit override, keeps one path.
- **Prefixing the title** (`My Project / [42] Fix login redirect`). Cheaper, but long project and task names collide on one line, which is exactly where the message is already hard to scan.

## Additional Context

- `main` @ 5060210, Node 22, macOS.
- Touches `packages/shared/src/telegram.ts`, `packages/agent/src/notifier.ts`, `packages/mcp/src/utils/broadcast.ts` and their tests, plus `docs/configuration.md`.

lee-to:
## Summary

- include the project name in Telegram task-status notifications
- keep the existing two-line notification format when the project cannot be resolved
- preserve best-effort delivery when project lookup or Telegram delivery fails

Closes #168.

## Changes

- added optional `projectName` support to `TelegramNotificationOptions`
- render the escaped project name as the first bold MarkdownV2 line and the task title as plain text
- resolve task projects through `findProjectByTaskId` in both agent and MCP broadcasters
- allow callers to provide an explicit project name and skip the database lookup
- log lookup failures at debug level without blocking notification delivery
- added coverage for formatting, Markdown escaping, database-backed resolution, explicit overrides, and lookup failures
- documented the project-aware Telegram message format and fallback behavior

## AI Model

- [x] GPT (latest)
- [ ] Claude Fable (latest)
- [ ] Claude Opus (latest)
- [ ] Claude Sonnet (latest)
- [ ] Other
- [ ] No AI used

## Test Plan

- [x] `@aif/shared`: 232 tests passed; 80.56% statement coverage
- [x] `@aif/agent`: 383 tests passed; 87.63% statement coverage
- [x] `@aif/mcp`: 104 tests passed; 93.61% statement coverage
- [x] Lint passed for all touched packages
- [x] TypeScript builds passed for all touched packages
- [x] Documentation updated
- [x] `git diff --check` passed
- [ ] `npm run ai:validate` is fully green — the repository-wide run is currently blocked by the installed Agent SDK bundling a Claude Code version below the minimum required by `claudeVersion.test.ts`; the failure reproduces in isolation and is unrelated to these changes
- [ ] Manually verified against a live Telegram bot


lee-to:
## Code review

This PR adds project-aware Telegram task status notifications across the agent and MCP broadcasters, with formatting/tests/docs for the new three-line message. The implementation is small and CI is green, but it changes an existing external notification contract and adds a task-project DB lookup to existing task-move broadcast paths, so it needs the required off-by-default rollout gate before merge.

### Must fix

1. **Gate the new Telegram project-name behavior** — `packages/agent/src/notifier.ts:151`, `packages/mcp/src/utils/broadcast.ts:65` — these call `resolveTelegramProjectName(...)` directly in the existing `task:moved` notification path. That means every opted-in Telegram installation gets the new message shape immediately, and even installations without Telegram configured still do the project lookup before `sendTelegramNotification()` can return early. Per the risky-change gate, this external-contract change should be behind an env flag that defaults to off. Suggested shape: add `AIF_NOTIFICATIONS_PROJECT_NAMES_ENABLED=false` to `packages/shared/src/env.ts`, `.env.example`, and `docs/configuration.md`; read it once through the central env path; and when false, pass no `projectName` and perform no DB lookup so the old two-line behavior remains exact.

### Should fix

1. **Remove duplicated project lookup plumbing** — `packages/agent/src/notifier.ts:18` and `packages/mcp/src/utils/broadcast.ts:13` contain the same `resolveTelegramProjectName` helper. This is already duplicated across two packages, including identical logging and exception handling. After adding the flag, extract the shared behavior to one appropriate owner, or make the broadcaster call a small shared notification/project-name helper so future changes to lookup fallback semantics do not have to be patched twice. This maps to the project DRY rule in `.ai-factory/RULES.md:8`.

### Nits

None.

### Context gates

- Architecture: pass — agent and MCP DB access goes through `@aif/data`, matching `packages/agent/CHECKLIST.md:10` and `packages/mcp/CHECKLIST.md:6`.
- Rules: error — the rollout gate above blocks merge; DRY also needs cleanup per `.ai-factory/RULES.md:8`.
- Roadmap: pass — notification polish does not contradict the current roadmap.
- CHECKLIST compliance (touched packages: `agent`, `mcp`, `shared`): warn — relevant package checks are mostly covered by tests and DB-boundary usage, but the root checklist's package-review requirement applies (`CHECKLIST.md:1`).
- Docs sync: pass for current behavior docs — `docs/configuration.md` was updated. If the feature flag is added, `.env.example` and configuration docs need the new flag entry too.
- PR size (245 changed lines / 7 files / one concern): pass — reviewable size.
- Risk gating (feature flag): error — new external notification behavior is not protected by an off-by-default `AIF_*_ENABLED` flag.

Verdict: `REQUEST_CHANGES` until the project-name notification path is flag-gated and the disabled path preserves the current lookup-free two-line behavior.

lee-to:
## Follow-up review

Reviewed `0f0a826` against the previous findings.

### Response pass

- ✅ Added the off-by-default `AIF_NOTIFICATIONS_PROJECT_NAMES_ENABLED` rollout flag; disabled mode preserves the legacy message and performs no project lookup.
- ✅ Centralized the lazy, best-effort project-name resolution in `@aif/shared`; the duplicated agent/MCP helpers were removed.
- ✅ Updated configuration docs and corrected the Telegram example to match the actual formatter.

### Gates

- Architecture, project rules, roadmap scope, package checklists, and documentation: pass.
- `npm run ai:validate`: pass.
- GitHub Actions: Build, ESLint, Tests, MCP Unit, and npm audit all pass.

### Verdict

LGTM — the blocking and should-fix findings are addressed.

PR #183, opened 2026-09-13
fix(agent): admit eligible parallel-project work mid-cycle

### LINKED ISSUE #182: Backlog admission latency is bounded by the longest in-flight lane, not by the task's own project capacity
# Backlog admission latency is bounded by the longest in-flight lane, not by the task's own project capacity

## Symptom

A task dispatched to the backlog of a **parallel-enabled project with all slots free** can wait
hours before processing starts. The wait is not caused by the task's own project — it is caused
by _another_ project's lane still being mid-pass inside the same coordinator poll cycle. Queue
admission (backlog → `planning`, and pipeline pickup) happens only at poll-cycle boundaries, and a
cycle stays open until **every** selected project lane has drained. Measured on a live instance:
a task waited **90.6 minutes** after creation while its own project sat completely idle, because
an unrelated project's review stage held the cycle open; historical p90 reaches **16.3 hours**.

## Mechanism

All anchors verified against `main` @ `d82e466` (`packages/agent/src/coordinator.ts` unless noted):

1. **Admission runs once per cycle, at cycle start.**
   `runPollCycle()` (`coordinator.ts:1048` — `async function runPollCycle(): Promise<void> {`)
   calls `processAutoQueueAdvance()` (`:1062` — `processAutoQueueAdvance();`) as part of its
   opening prelude. `processAutoQueueAdvance()` (`:907` —
   `export function processAutoQueueAdvance(): number {`) lifts backlog tasks into `planning`
   with a fill loop (`:977` — `while (active < limit) {`) guarded by the atomic
   `claimBacklogTaskForAdvance` CAS claim (`:995` —
   `if (!claimBacklogTaskForAdvance(next.id, autoQueueCommit)) {`;
   `packages/data/src/index.ts:2305`). The next backlog candidate is picked by
   `nextBacklogTaskByPosition` (`packages/data/src/index.ts:2625`). Nothing re-runs this lift
   until the _next_ cycle starts.

2. **A cycle stays open until every lane drains.**
   `runPollCycle` selects up to `COORDINATOR_MAX_CONCURRENT_PROJECTS` project lanes and awaits
   them all (`Promise.allSettled(projectIds.map(processProjectLane))`). Each lane
   (`coordinator.ts:1115` — `async function processProjectLane(projectId: string): Promise<void> {`)
   walks the pipeline stages sequentially and, per stage, `await`s its started tasks
   (`await Promise.allSettled(spawned)`). So one long implementer/review stage anywhere keeps
   the whole cycle — and therefore the next admission — open for its full remaining runtime.

3. **Triggers during an active cycle only queue a follow-up.**
   The single-flight guard in `pollAndProcess()` (`coordinator.ts:1295–1298`):
   `if (activePollPromise) { followUpPollRequested = true; … return activePollPromise; }`.
   The scheduler tick _and_ the wake channel (`task:created` / `task:moved` / `agent:wake`
   events → `triggerWake` → `pollAndProcess`) both hit this guard while a cycle is active —
   the event-driven wake path exists but cannot admit anything mid-cycle.

Consequence: a task created (or made eligible) at time _T_ is admitted only when the cycle that
was active at _T_ fully drains — i.e. its admission latency equals the **remaining runtime of the
longest in-flight lane**, independent of its own project's free capacity. A manual `start_ai` on
the waiting task changes its status to `planning` but starts nothing, because candidate pickup
also happens only inside a cycle's lanes.

## Empirical impact

**Incident timeline (UTC, 2026-09-11):**

- 12:22:48.753 — a task reaches done; the advance pass lifts 3 tasks across 3 projects. For a
  now-emptied project the pass logs `active:0, limit:5, "no more backlog ready"` — it had nothing
  queued.
- 12:27:40–47 — four tasks are created on that emptied project.
- 12:24–12:45 — every coordinator tick logs `Poll cycle already active; queued one follow-up
cycle` (16×). The four tasks wait.
- 12:45 — manual `start_ai` on the four tasks flips status to `planning` but starts nothing.
- 13:58:14.033 — the last foreign task passes review (`Auto review gate accepted review, moving
to done` → `Poll cycle complete` → `Starting poll cycle`). Same second, the advance lifts 2
  more backlog tasks, and at 13:58:14.829 the four waiting tasks are claimed via the
  `[FIX:149] Revalidating task candidate after coordinator permit` lane path; worktrees are
  created 13:58:16–19.
- **Admission latency 90.6–90.7 min = exactly the foreign cycle's remaining runtime.**

**Historical distribution** (200 tasks, 2026-07-26 → 2026-09-11; latency = first activity-log
timestamp − `created_at`):

| Day    | Median | Max                     |
| ------ | ------ | ----------------------- |
| Aug 01 | 27.7m  | —                       |
| Aug 12 | —      | 136.1m                  |
| Aug 17 | 58.9m  | —                       |
| Aug 18 | —      | p90 **978.2m (16.3 h)** |
| Aug 19 | 45.8m  | —                       |
| Sep 03 | 57.3m  | 151.2m                  |
| Sep 06 | 66.6m  | 174.4m                  |
| Sep 08 | 65.4m  | **225.6m**              |

Honest caveat: **instant admission (<2 min) accounts for ≈ half of all tasks** — dispatches that
landed in an idle window were admitted immediately. The design is not broken for idle factories;
the cost lands specifically on factories that are busy when the task arrives, and it grows with
the runtime of whatever is in flight. Batch signatures are visible (five tasks created together
all waiting exactly 65.0–65.4m — one admission window lifted them together), consistent with the
cycle-boundary mechanism rather than per-task noise.

### A. Mid-cycle admission pass on busy-cycle triggers (small, reuses all machinery)

When `pollAndProcess()` hits the active-cycle guard, additionally fire a single-flight
**admission pass**: run the same CAS-protected `processAutoQueueAdvance()` lift, then start
project lanes for **parallel-capable projects only**, reusing the exact cycle lane function.

- Pros: minimal diff; lanes already enforce correctness under concurrency — the shared
  `StageSemaphore` (per-project-stage key max + global ceiling), the `[FIX:149]`
  post-permit revalidation, and the atomic `claimCoordinatorTaskIfEligible` CAS make concurrent
  cycle/pass lanes safe by construction. Sequential projects can be excluded from the pass
  entirely, preserving their cycle-boundary semantics with zero new windows.
- Cons: per-tick query cost while a cycle is busy (a few indexed queries per project — the same
  work a cycle start does); another moving part in the coordinator's lifecycle; the pass must be
  detached (must not extend the returned cycle promise) or admission serializes again behind the
  very drain it is meant to bypass.

### B. Per-project long-lived lanes (architectural)

Replace per-cycle lane batches with one long-lived lane per project that reacts to capacity
changes (task exits) and arrivals.

- Pros: conceptually cleanest; admission latency collapses to event-handling time; no cycle
  concept to reason about.
- Cons: a rewrite of the lane lifecycle — lifecycle ownership, backpressure when
  `COORDINATOR_MAX_CONCURRENT_PROJECTS` projects are active, interaction with
  `failedInCycle`-style cycle scoping, watchdog/recovery placement all need redesign. High
  regression risk relative to the size of the bug.

### Risks either design must address

- **Double-processing** — solved today by task-row CAS claims (`claimCoordinatorTaskIfEligible`)
  plus lock TTLs; any concurrent-lane design must route through them unchanged.
- **Cap bypass** — per-project and global caps are enforced by the shared semaphore, not by the
  cycle boundary; a design that admits outside a cycle must keep using that semaphore.
- **Sequential-project regression** — non-parallel projects rely on the
  `hasActiveLockedTaskForProject` cross-cycle guard + key-max-1 semaphore; a mid-cycle path must
  not open the brief lock gap between a sequential task's stages.
- **Review-gate reentrancy** — the auto-review gate runs inside `processOneTask` per claimed
  task; CAS claiming keeps it non-reentrant; no change needed, but worth an explicit test.

## Concrete proposal (implemented and linked as a PR)

Design **A**, plus a rollout flag:

1. Extract the lane function out of `runPollCycle`'s closure so cycles and passes share one
   implementation (`runProjectLane(projectId, failedSet, concurrencyCache)`).
2. In `pollAndProcess`'s busy branch, fire a detached, single-flight admission pass:
   `processAutoQueueAdvance()` + `Promise.allSettled` of lanes for
   `listCoordinatorActionableProjectIds(COORDINATOR_MAX_CONCURRENT_PROJECTS)` filtered to
   parallel-capable projects (sequential projects keep exact cycle-boundary semantics).
3. `AGENT_MID_CYCLE_ADMISSION_ENABLED` (default `true`) restores pure cycle-boundary admission
   for rollout protection.
4. The pass deliberately does **not** run cycle-start recovery jobs (stale-claim release,
   watchdog passes, GitHub sync) — those keep their per-cycle frequency contract.

What the patch deliberately does not do: no DB schema changes, no per-project long-lived lanes,
no change to sequential-project scheduling, no change to any concurrency cap semantics, and no
new required configuration (the flag defaults to the new behavior).

Known follow-up (out of scope for the patch): `listCoordinatorActionableProjectIds` orders by
`min(createdAt)` with the `COORDINATOR_MAX_CONCURRENT_PROJECTS` limit, so in a factory with more
than that many busy projects a freshly-created project's tasks can still wait behind older
actionable projects. The patch applies the same bound to the pass to preserve the knob's
contract; making lane-batch ordering arrival-aware is a separate, compatible improvement.

artyhoo:
Fixes #182.

## What

Admits eligible backlog work mid-poll-cycle instead of waiting for the cycle boundary. At `main` @ `d82e466`, `processAutoQueueAdvance()` runs only inside `runPollCycle`'s opening prelude (`coordinator.ts:1062`), and a cycle stays open until every selected project lane drains — so a task dispatched to a parallel-enabled project with all slots free can sit in backlog for the full remaining runtime of an unrelated project's lane (measured live: 90.6 min; historical p90 16.3 h; details in the issue).

## How

- Extracts the lane function out of `runPollCycle`'s closure (`runProjectLane`) so cycle lanes and the new pass share one implementation.
- In `pollAndProcess`'s busy branch, fires a detached, single-flight mid-cycle admission pass: `processAutoQueueAdvance()` first, then `Promise.allSettled` over parallel-capable project lanes only. Sequential projects keep exact cycle-boundary semantics (cross-cycle DB locks preserved); worktree-sharing projects stay forced-serial.
- The pass deliberately does not extend the cycle promise and does not run cycle-start recovery jobs (stale-claim release, watchdog, GitHub sync) — those keep their per-cycle frequency contract.
- Rollout flag `AGENT_MID_CYCLE_ADMISSION_ENABLED` (default `true`); `false` restores pure cycle-boundary admission (test-verified).

No DB schema changes, no concurrency-cap semantics changes, no new required configuration.

## Testing

- New regression suite `packages/agent/src/__tests__/midCycleAdmission.test.ts` (3 tests): **3/3 pass on this branch; 3/3 fail on the base `d82e466`** (assertions show the defect itself — a parallel project's backlog task never admitted mid-cycle).
- Full agent suite differential: base 386 tests / 1 pre-existing failure (`autoReviewHandler`) vs branch 389 / same 1 — **zero new failures**, +3 new tests green.
- `packages/shared`: 235/235 green (env flag touched).
- Live-fire on a production instance: a wake arriving during an active cycle ran the pass and moved another project's backlog task to `planning` ~2.5 min after deploy (log-evidenced in `A1-REPORT.md` on the branch).

## Docs on the branch

`ISSUE-ADMISSION-LATENCY.md` (the research write-up behind #182) and `A1-REPORT.md` (worker report with the falsifiers list).


artyhoo:
Author's own follow-up after a cold verification pass on this branch. Three things the reviewer should have before spending a round on it, including one that argues against reviewing it in its current shape.

## 1. Overlap with #165 — please read that one first

I opened this without noticing #165 (`fix(agent): stop the poll cycle from stalling behind long stages`), which is already approved and only waiting on conflict resolution. It addresses the same root cause and is broader. Concretely:

- Its **Change 2** is a precondition for this PR to work at all on the cron path. `startPollScheduler` returns early on `isRunning` *before* calling the callback (`packages/agent/src/pollScheduler.ts`), and the callback awaits `pollAndProcess()`, which settles only when the whole cycle drains. Measured on this branch's base: a pending callback receives **1 invocation across ~2 interval boundaries**. So `triggerMidCycleAdmission()` in this PR is reachable from the wake channel only — with `AGENT_WAKE_ENABLED=false`, or whenever the WS drops to polling-only mode, this PR contributes nothing.
- Its **Change 1** (bounded overlapping cycles) already covers the main case this PR was written for: admitting work for *other* projects while a long stage runs.
- What remains unique here is narrow: a second task of the **same** parallel project in a different stage — precisely the case #165 excludes on purpose via `hasActiveProject()` to preserve project-local stage order from #149.

Given that, I would rather not consume a review round now. Suggestion: land #165, then I re-measure what delta, if any, is left and either rework this branch on top of it or close it.

## 2. Rollout default — this PR does what #165 was asked not to do

`AGENT_MID_CYCLE_ADMISSION_ENABLED` defaults to `true` here, i.e. the coordinator hot path changes unconditionally on upgrade. That is the same shape as must-fix #1 in the first review of #165 ("gate the overlapping-cycle rollout off by default"). If this branch survives the step above, the default flips to `false` before any review round.

## 3. The two red checks are not from this change

- **npm audit** — also red on `lee-to/aif-handoff@main` (current head `af9b2ee`). Pre-existing.
- **MCP Unit** — the audit itself passed: `Total Score: 100/100`, `Finding Counts: total=0`, `outcome=success`. The job failed on the following step, `github/codeql-action/upload-sarif`, which cannot write code-scanning results from a fork PR. Infrastructure, not the change. Build, Tests and ESLint are green.

## 4. One correction already pushed

Commit `3a1172f` (already in this PR's head) fixes an inaccurate claim in my own CHANGELOG entry: it said "all concurrency caps unchanged". Measured against the base — a cycle lane and an admission lane can serve one parallel project at once, holding permits under different per-project-*stage* keys, so its in-flight task count reaches 2 with `COORDINATOR_MAX_CONCURRENT_TASKS_PER_PROJECT=1` (base: 1, because a single lane walks stages sequentially). The knob's formal semantics are unchanged (`docs/configuration.md` has always called it a per-stage cap) and the global cap still bounds the total, but the "caps unchanged" phrasing was wrong and is now replaced by a description of the actual effect. The same commit records that a busy cycle sees at most one admission wave, since the pass is single-flight and lives until the work it admitted drains.

lee-to:
## Code review

This PR tackles a real admission-latency problem in the agent coordinator, but it changes the coordinator hot path and concurrent lane semantics. The current shape is too risky to merge as-is: the rollout flag is enabled by default, the documented periodic-poll trigger is not actually reachable through the scheduler, and the PR exceeds the review size bar while bundling implementation, tests, operational docs, and large research artifacts.

### Must fix

1. **Make the risky coordinator change dark-launchable** — `packages/shared/src/env.ts:170`, `.env.example:142`, `docs/configuration.md:58` — the new behavior changes request/agent hot-path scheduling and can increase same-project in-flight work across stages, but `AGENT_MID_CYCLE_ADMISSION_ENABLED` defaults to `true`. The project review rule for risky changes requires an environment-driven feature flag that defaults to `false`, and it also requires the `AIF_<AREA>_<FEATURE>_ENABLED` naming shape. Please rename this to something like `AIF_AGENT_MID_CYCLE_ADMISSION_ENABLED`, default it to `false`, and update the docs/tests/examples accordingly. The existing off path in `packages/agent/src/coordinator.ts:1397` is the right place for the gate; the problem is the default and name.

2. **Wire the poll-tick trigger or narrow the contract to wake-only** — `packages/agent/src/pollScheduler.ts:23` and `packages/agent/src/coordinator.ts:1414` — the PR says a “poll tick or wake event” that lands during an active cycle fires mid-cycle admission, but periodic ticks do not call `pollAndProcess()` while the previous scheduler callback is still awaiting. `startPollScheduler()` returns at `if (isRunning) return`, and `packages/agent/src/index.ts:44-47` awaits the whole `pollAndProcess()` promise, so a long active cycle suppresses interval callbacks before they can reach the active-cycle branch at `coordinator.ts:1415-1418`. The new tests directly call `pollAndProcess()` (`midCycleAdmission.test.ts:169-170`) and therefore cover wake/direct calls, not the scheduler path. Please either change the scheduler/agent wiring so overlapping ticks invoke a lightweight trigger path, or update the code/docs/tests to state this feature is wake-only and does not help when `AGENT_WAKE_ENABLED=false` or the WS is down.

3. **Decomposition required** — PR size is 1,288 changed lines across 13 files (`+1121/-167`). That exceeds the hard >800-line review threshold. The largest issue is that the functional coordinator refactor/change is bundled with a 272-line regression suite plus two large root-level research/report artifacts (`A1-REPORT.md`, `ISSUE-ADMISSION-LATENCY.md`). Please split this into reviewable boundaries, for example: implementation + focused tests (`packages/agent/src/coordinator.ts`, agent/shared env tests), rollout/docs (`.env.example`, `docs/configuration.md`, `docs/architecture.md`, `CHANGELOG.md`, checklist), and research artifacts in their own docs PR or issue attachment. If the implementation stays dependent on #165, rebase after #165 and keep only the residual delta.

### Should fix

1. **Avoid shipping temporary task reports at repo root** — `A1-REPORT.md` and `ISSUE-ADMISSION-LATENCY.md` look like AI Factory handoff/research artifacts rather than product docs. They add most of the review surface and duplicate material already in the PR body/issue. Move durable content into `docs/` or the GitHub issue; otherwise keep them out of the product repository.

2. **Reconcile with PR #165 before another review round** — the author note says #165 already covers the broader “long stage blocks unrelated project admission” case, and that this PR’s remaining unique behavior is narrower: same parallel project, different stage. That overlap makes the current design hard to reason about independently. Rebase on the accepted scheduler/cycle model from #165, then re-measure whether mid-cycle admission is still needed.

### Nits

- No debug leftovers, TODO/FIXME leaks, mojibake, or direct DB-boundary violations found in the changed hunks.

### Context gates

- Architecture: pass — changed agent code still uses the existing data/runtime boundaries; I did not see direct DB access from `agent` outside `@aif/data`.
- Rules: error — `.ai-factory/RULES.md` requires SOLID/DRY, checklist use, docs sync, and safe rollout discipline; the risky feature flag defaults on instead of off.
- Roadmap: warn — this is a fix PR and references issue #182, but the overlap with #165 should be resolved before merge.
- CHECKLIST compliance (touched packages: `packages/agent`, `packages/shared`): warn — agent/shared checklist items were considered, but the scheduler checklist now claims periodic ticks can fire the mid-cycle pass even though `startPollScheduler()` suppresses those callbacks.
- Docs sync: error — docs were updated, but they document a poll-tick trigger that the current scheduler wiring does not provide.
- PR size (1,288 lines / 13 files / implementation + tests + rollout docs + research artifacts): error — split as described under Must fix.
- Risk gating (feature flag): error — use `AIF_AGENT_MID_CYCLE_ADMISSION_ENABLED=false` by default, with the current behavior reachable only when explicitly enabled.

Verdict: REQUEST_CHANGES.

